PR #40, opened 2022-06-15
refactor: rename sub-crate `udf` to `df_operator`

### LINKED ISSUE #18: Rename `udf` sub-crate
**Description**
We have a sub-crate `udf` that contains some operator implementation for DataFusion. The name is some kinds of ambiguous, those operators are UDF from DF's perspective, but are the same as other DF build-in from the user. 

**Proposal**
Rename it, like `df-operator` or `df-function`. Or other appropriate names.

**Additional context**
 There is a `query_engine` sub-crate that also extends DF operators. Merge these two crates should also be an option.

GoGim1:

Closes #18

# Rationale for this change
 
We have a sub-crate udf that contains some operator implementation for DataFusion. The name is some kinds of ambiguous, those operators are UDF from DF's perspective, but are the same as other DF build-in from the user.

* rename sub-crate `udf` to `df_operator`

no

# How does this change test
no


waynexia (approved):
Brilliant! 🚀 Appreciate it @GoGim1

PR #47, opened 2022-06-16
refactor: Reduce lock exclusivity for read operation

VirrageS:
# Rationale for this change
 

Since we are not mutating the values there is no need to take a `write` lock. Taking a `read` lock is sufficient.

No

# How does this change test

Run limiter tests.

### DIFF
--- a/server/src/limiter.rs
+++ b/server/src/limiter.rs
@@ -67,11 +67,11 @@ impl Limiter {
     }
 
     pub fn get_write_reject_list(&self) -> HashSet<String> {
-        self.write_reject_list.write().unwrap().clone()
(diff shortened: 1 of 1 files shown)


chunshao90 (approved):
LGTM

waynexia (approved):
Nice catch 🚀 Thank you @VirrageS 

A little more request: it would be awesome if the caller of these two methods can be simplified as well. But this PR is good enough, I can merge it after CI.

https://github.com/CeresDB/ceresdb/blob/2fda373f49b17269464afe41b275fcf0ca2cc497/server/src/handlers/admin.rs#L60-L69

PR #48, opened 2022-06-16
refactor: Simplify and optimize `should_limit` method in Limiter

VirrageS:
# Rationale for this change

This MR simplifies the `Plan::Query` case to improve the performance and readability.

There are 2 changes:
1. Remove inefficient `clone()` which can be expensive and simply iterate over values holding the lock.
2. Refactor `for` loop into `.iter().any(...)` which can be potentially faster than `for` loop as compiler could optimize it.

No

# How does this change test

Run limiter tests.

### DIFF
--- a/server/src/limiter.rs
+++ b/server/src/limiter.rs
@@ -22,19 +22,12 @@ impl Default for Limiter {
 impl Limiter {
     pub fn should_limit(&self, plan: &Plan) -> bool {
         match plan {
-            Plan::Query(query) => {
-                let read_reject_list = self.read_reject_list.read().unwrap().clone();
-                for table in read_reject_list {
-                    if query
-                        .tables
-                        .get(TableReference::from(table.as_str()))
-                        .is_some()
-                    {
-                        return true;
-                    }
-                }
-                false
-            }
+            Plan::Query(query) => self.read_reject_list.read().unwrap().iter().any(|table| {
+                query
+                    .tables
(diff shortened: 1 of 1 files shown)


jiacai2050 (approved):
LGTM

server/src/limiter.rs
```diff
@@ -22,19 +22,15 @@ impl Default for Limiter {
 impl Limiter {
     pub fn should_limit(&self, plan: &Plan) -> bool {
         match plan {
-            Plan::Query(query) => {
-                let read_reject_list = self.read_reject_list.read().unwrap().clone();
```
jiacai2050, line 26: Clone is not useless here, it will release read_reject_list's lock.

VirrageS: Yeah, I think you are right.

But in sense of whole function its useless as it's much faster to just iterate and check values instead of cloning them and then checking.

VirrageS: Or maybe the better word is "inefficient" instead of "useless".

VirrageS: I've updated the MR description.

waynexia: I feel OK with this refactor. Readers would not block each other and acquiring a write lock of `read_reject_list` is a fairly unfrequent operation. So I prefer to check values inside the read lock rather than clone them out.

waynexia (approved):
Thanks @VirrageS ❤️

PR #56, opened 2022-06-19
feat: add mysql server support

### LINKED ISSUE #43: add mysql server support
**Description**

ceresdb doesn't seem to support mysql client yet

**Proposal**

we can use this crate solves that problem by acting as a MySQL server, and delegating operations such as querying and query execution to user-defined logic.

**Additional context**

https://github.com/jonhoo/msql-srv

dust1:

Closes #43 

# Rationale for this change
Convenient for mysql client to use

- [x] add mysql server structure
- [x] implement sql execute

there are user-facing changes that needs to update the documentation or configuration.

a new configuration items have been added to the `server::config.rs` file:
1. mysql_port: the mysql server listening port, default 3307

# How does this change test

todo

# Note
what I tried at first was opensrv, but I am not very proficient in `tokio` and related concurrent operations. 
since ceresDB needs to pass in `Instance<C, Q>` when executing sql, when I try to use `AsyncMysqlIntermediary` to run the service, I am troubled by the asynchrony of these two generic types. It requires me to declare the specific types of C and Q, such as: `impl<C: Catalog, Q: Query>`, but in this case, the `server.rs` needs to be changed to this:
```rust
pub struct Server<C: Catalog, Q: Query> {
    http_service: Service<C, Q>,
    rpc_services: RpcServices,
    mysql_service: mysql::MysqlHandler<C, Q>,
}
```
I think this modification is not very good, so I finally use `msql-srv`

# How to use it?
First, we can start ceresdb as usual. We can see that there is an extra entry in the startup log:
```
2022-06-22 21:16:10.980 INFO [server/src/mysql/handler.rs:54] Mysql service listens on 127.0.0.1:3307
```
Mysql service starts on port 3307 by default, you can customize it by modifying the ·mysql_port· option in the configuration file

Next, you should install a `mysql-client` and connect to mysql by typing:
```
mysql -P3307 -h127.0.0.1
```
We will enter the mysql console.
image

Finally, Use ceresdb like mysql.

### Create table
```
CREATE TABLE `hello_mysql` (`name` string TAG, `value` double NOT NULL, `t` timestamp NOT NULL, TIMESTAMP KEY(t)) ENGINE=Analytic with (enable_ttl='false');
```
image

### Write data
```
INSERT INTO hello_mysql(t, name, value) VALUES(1651737067000, 'ceresdb', 100);
```
image

### Read data
```
select * from hello_mysql;
```
image

### Show create table
```
show create table hello_mysql;
```
image

### Drop table
```
DROP TABLE hello_mysql;
```
image


server/src/server.rs
```diff
@@ -59,12 +70,14 @@ define_result!(Error);
 pub struct Server<C, Q> {
     http_service: Service<C, Q>,
     rpc_services: RpcServices,
+    mysql_service: mysql::MysqlHandler<C, Q>,
```
waynexia, line 73: >I think this modification is not very good, so I finally use msql-srv

I'm ok with this. Add trait constraints is acceptable if we need to.

waynexia (commented):
Thanks a lot! This is really cool :heart: 

>but I am not very proficient in tokio and related concurrent operations.

Don't worry :wink: feel free to reach out to me if anything is unclear. I personally prefer tokio for that in this case we can consolidate the runtime with other parts in CeresDB. i.e.: serve MySQL connection with `EngineRuntimes` like HTTP service does.

server/src/server.rs
```diff
@@ -158,6 +171,21 @@ impl<C: CatalogManager + 'static, Q: QueryExecutor + 'static> Builder<C, Q> {
...
+            port: self.config.mysql_port,
+            thread_num: self.config.mysql_thread_num,
+        };
+
+        let mysql_service = mysql::Builder::new(mysql_config)
+            .runtimes(runtimes.clone())
+            .instance(instance.clone())
+            .build()
+            .context(BuildMysqlService)?;
+        if let Err(err) = mysql_service.start() {
```
waynexia, line 185: `MysqlHandler::start()` serves the TCP stream in the current thread, so this `start()` won't return until TCP is finished.

server/src/server.rs
dust1: Using `tokio` is indeed more better than `threadpool`.😂 I will change him to `tokio` form.

server/src/server.rs
dust1: After adding the trait constraints, maybe ceresDB can start the mysql service directly in the `build` function, I will try it later🧐

server/Cargo.toml
```diff
@@ -39,6 +39,8 @@ tokio = { version = "1.0", features = ["full"] }
 twox-hash = "1.6"
 df_operator = { path = "../df_operator" }
 warp = "0.3"
+msql-srv = "0.9"
```
chunshao90, line 42: dep order

dust1: ok

dust1:
@waynexia hi!, I think I have done the mysql service startup part😀. If nothing else, I'll implement the mysql_worder part next.

waynexia:
> @waynexia hi!, I think I have done the mysql service startup part😀. If nothing else, I'll implement the mysql_worker part next.

Great! I'll review it the next day 👍 . And since you have done an excellent separation of the work and code, we are able to deliver this feature in serval PRs with each focus on one part. For example, this one constructs the MySQL server framework and implements worker in the following PR. But it's absolutely ok to implement worker in this PR, up to you 🤗

server/src/mysql/mysql_writer.rs
```diff
@@ -0,0 +1,124 @@
...
+                Datum::Int64(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int32(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int16(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int8(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Boolean(_) => ColumnType::MYSQL_TYPE_SHORT,
+                Datum::Null => ColumnType::MYSQL_TYPE_VARCHAR,
+            }
+        }
+
+        fn make_column_by_filed(k: &str, v: &Datum) -> Column {
```
ShiKaiWi, line 68:
```suggestion
        fn make_column_by_field(k: &str, v: &Datum) -> Column {
```

server/src/mysql/mysql_writer.rs
```diff
@@ -0,0 +1,124 @@
...
+        writer: QueryResultWriter<'a, W>,
+        rows: Vec<HashMap<String, Datum>>,
+    ) -> std::io::Result<()> {
+        let default_response = OkResponse::default();
+        if rows.is_empty() {
+            writer.completed(default_response)?;
+            return Ok(());
+        }
+
+        fn convert_field_type(field: &Datum) -> ColumnType {
```
ShiKaiWi, line 48: Maybe we should define the `convert_field_type` and `make_column_by_field` outside the `write_rows`. Otherwise, it will be a little hard to write unit test for them.

server/src/mysql/mysql_writer.rs
```diff
@@ -0,0 +1,124 @@
...
+            match field {
+                Datum::Timestamp(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Double(_) => ColumnType::MYSQL_TYPE_DOUBLE,
+                Datum::Float(_) => ColumnType::MYSQL_TYPE_FLOAT,
+                Datum::Varbinary(_) => ColumnType::MYSQL_TYPE_VARCHAR,
```
ShiKaiWi, line 53: I guess MYSQL_TYPE_LONG_BLOB may be a better choice.

server/src/mysql/mysql_writer.rs
```diff
@@ -0,0 +1,124 @@
...
+                Datum::UInt64(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::UInt32(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::UInt16(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::UInt8(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int64(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int32(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int16(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Int8(_) => ColumnType::MYSQL_TYPE_LONG,
+                Datum::Boolean(_) => ColumnType::MYSQL_TYPE_SHORT,
+                Datum::Null => ColumnType::MYSQL_TYPE_VARCHAR,
```
ShiKaiWi, line 64: MYSQL_TYPE_NULL?

server/src/mysql/mysql_writer.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        let columns = &rows[0]
+            .iter()
+            .map(|(k, v)| make_column_by_filed(k, v))
+            .collect::<Vec<_>>();
+        let columns_keys = columns.iter().map(|c| c.column.clone()).collect::<Vec<_>>();
+        let mut row_writer = writer.start(columns)?;
+        for row in &rows {
+            for key in &columns_keys {
```
ShiKaiWi, line 85: Is it better to loop over the `columns`(no need to generate `column_keys` at L82 and clone `columns` at L83)?

ShiKaiWi (commented):
I guess there is no need ad mysql prefix to the files in the mysql mod, e.g. `mysql_builder` -> `builder`.

dust1:
thanks @waynexia & @ShiKaiWi . I will fix these later.

ShiKaiWi:
> thanks @waynexia & @ShiKaiWi . I will fix these later.

👌. We are not in a hurry so just do it as your wish.

server/src/mysql/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+mod mysql_builder;
+mod mysql_handler;
+mod mysql_worker;
+mod mysql_writer;
```
chunshao90, line 6:
```
mod builder;
mod handler;
mod worker;
mod writer;
```

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,141 @@
...
+        _param: opensrv_mysql::ParamParser<'a>,
+        writer: QueryResultWriter<'a, W>,
+    ) -> io::Result<()> {
+        writer.error(
+            ErrorKind::ER_UNKNOWN_ERROR,
+            "Execute is not supported in CeresDB".as_bytes(),
+        )
+    }
+
+    async fn on_close(&mut self, _id: u32) {}
```
waynexia, line 79: This is in conjunction with `on_prepare` and `on_execute`, and should return an error as well

server/src/mysql/builder.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+#[derive(Debug)]
+pub struct Config {
+    pub ip: String,
+    pub port: u16,
+}
+
+#[derive(Debug, Snafu)]
+pub enum Error {
```
waynexia, line 25: These `Error`s have reduced a lot. I'm wondering can we move all of them (those defined under `mysql` sub module) to `mysql/mod.rs` for a more concrete error type? E.g., it only needs to export `mysql::Error as MySQLError`, rather than export builder's error, handler's error etc separately.

waynexia (commented):
This implementation generally looks good to me 👍 . There are a few things to do before  merging it:
- Add unit test. For this version I think covering `writer` is enough.
- Add an example. I suppose you have tried this via some MySQL client. You can update the process to PR description. Like how to config and start CeresDB server, how to connect to it, what query have you tried and what is the result.

server/src/config.rs
```diff
@@ -67,6 +68,7 @@ impl Default for Config {
         Self {
             bind_addr: String::from("127.0.0.1"),
             http_port: 5000,
+            mysql_port: 3307,
```
jiacai2050, line 71: 3306?

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,129 @@
...
+        writer: QueryResultWriter<'a, W>,
+    ) -> io::Result<()> {
+        writer.error(
+            ErrorKind::ER_UNKNOWN_ERROR,
+            "Execute is not supported in CeresDB".as_bytes(),
+        )
+    }
+
+    async fn on_close(&mut self, _id: u32) {
+        info!("mysql client id{} closes", _id);
```
jiacai2050, line 66: id

server/src/mysql/handler.rs
```diff
@@ -0,0 +1,105 @@
...
+use query_engine::executor::Executor as QueryExecutor;
+use snafu::ResultExt;
+use table_engine::engine::EngineRuntimes;
+use tokio::sync::oneshot::{self, Receiver, Sender};
+
+use crate::{
+    instance::{Instance, InstanceRef},
+    mysql::{error::*, worker::MysqlWorker},
+};
+pub struct MysqlHandler<C, Q> {
```
jiacai2050, line 18: How about `MysqlService`? it's consistent with others.

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,129 @@
...
+            runtimes,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl<W: std::io::Write + Send + Sync, C: CatalogManager + 'static, Q: QueryExecutor + 'static>
+    AsyncMysqlShim<W> for MysqlWorker<W, C, Q>
+{
+    type Error = std::io::Error;
```
ShiKaiWi, line 40: I find it the `Error` here requires `From<io::Error>` and I guess it will be more elegant to just `impl From<io::Error> for Error` so that all the returned results is our `Result`.

Ref: https://docs.rs/opensrv-mysql/0.1.0/opensrv_mysql/trait.AsyncMysqlShim.html

dust1:
About the `on_close` method, I found that it is not called after the mysql client exits, so I modified its description to the original comment of this method.🤔
```
    /// Called when the client wishes to deallocate resources associated with a previously prepared
    /// statement.
    async fn on_close<'a>(&'a mut self, stmt: u32)
    where
        W: 'async_trait;
```
https://docs.rs/opensrv-mysql/0.1.0/src/opensrv_mysql/lib.rs.html#173

server/src/mysql/handler.rs
```diff
@@ -0,0 +1,105 @@
...
+use opensrv_mysql::AsyncMysqlIntermediary;
+use query_engine::executor::Executor as QueryExecutor;
+use snafu::ResultExt;
+use table_engine::engine::EngineRuntimes;
+use tokio::sync::oneshot::{self, Receiver, Sender};
+
+use crate::{
+    instance::{Instance, InstanceRef},
+    mysql::{error::*, worker::MysqlWorker},
+};
```
chunshao90, line 17: newline

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,132 @@
...
+        Self {
+            generic_hold: PhantomData::default(),
+            instance,
+            runtimes,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl<W: std::io::Write + Send + Sync, C: CatalogManager + 'static, Q: QueryExecutor + 'static>
```
chunshao90, line 38:
```
where
    W: std::io::Write + Sync + Send,
    C: CatalogManager + 'static,
    Q: QueryExecutor + 'static,
```

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,132 @@
...
+            }
+            Err(err) => return Self::err(results, err),
+        }
+    }
+}
+
+impl<W: std::io::Write + Sync + Send, C: CatalogManager + 'static, Q: QueryExecutor + 'static>
+    MysqlWorker<W, C, Q>
+{
+    fn err(writer: QueryResultWriter<'_, W>, error: Error) -> Result<()> {
```
ShiKaiWi, line 105: It seems this method is useless now because the `Error` type is already `crate::mysql::error::Error` and we can just use `?` to return the error directly.

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,132 @@
...
+    async fn on_close(&mut self, id: u32) {
+        info!("client(id={}) wishes to deallocate resources associated with a previously prepared statement.", id)
+    }
+
+    async fn on_query<'a>(
+        &'a mut self,
+        sql: &'a str,
+        results: QueryResultWriter<'a, W>,
+    ) -> Result<()> {
+        let ctx = match self.create_ctx() {
```
ShiKaiWi, line 77: how about: `let ctx = self.create_ctx()?;`

server/src/mysql/worker.rs
ShiKaiWi: For example, the codes (L77-L80):
```rust
let ctx = match self.create_ctx() {
    Ok(res) => res,
    Err(error) => return Self::err(results, error),
};
```

can written as:
```rust
let ctx = self.create_ctx()?;
```

ShiKaiWi: I'm sorry that I misunderstand the usage of the `QueryResultWriter`. 

So the error must be returned by `writer.error` method? 

If it does, I guess we can define an method called `do_query( &'a mut self, sql: &'a str)` which just returns the query result and then in `on_query` use the `QueryResultWriter` to process the query result from the `do_query`. After that, I guess the error handling will be more concise and elegant.

server/src/mysql/writer.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+use std::collections::HashMap;
+
+use common_types::datum::Datum;
+use opensrv_mysql::{Column, ColumnFlags, ColumnType, OkResponse, QueryResultWriter};
+
+use crate::{handlers::sql::Response, mysql::error::*};
```
chunshao90, line 8: we prefer specific deps to *

chunshao90:
LGTM

ShiKaiWi:
@dust1  Thanks a lot for your contribution. The latest changeset is already good enough for merging except only a few issues to address. Apologize for requiring your so much work on addressing our CR. :smiling_face_with_three_hearts:

server/src/mysql/handler.rs
```diff
@@ -0,0 +1,105 @@
...
+pub struct MysqlService<C, Q> {
```
ShiKaiWi, line 18: - One newline above this line.
- Maybe the filename should be `service.rs` if `MysqlService` is used here.

dust1:
> @dust1 Thanks a lot for your contribution. The latest changeset is already good enough for merge except only a few issues to address. Apologize for requiring your so much work on addressing our CR. 🥰

It's also an improvement for me, and I'm very happy to be officially involved in this project.😆 thanks every reviewer.

ShiKaiWi (commented):
LGTM

waynexia (dismissed):
🚀

waynexia (approved):

waynexia:
Such an EPIC PR 🤩 thanks a lot @dust1 !

server/src/mysql/worker.rs
```diff
@@ -0,0 +1,145 @@
...
+{
+    type Error = crate::mysql::error::Error;
+
+    async fn on_prepare<'a>(
+        &'a mut self,
+        _query: &'a str,
+        info: StatementMetaWriter<'a, W>,
+    ) -> Result<()> {
+        info.error(
+            ErrorKind::ER_UNKNOWN_ERROR,
```
jiacai2050, line 61: Just came across this when review recent PR. 

Isn't [ER_NOT_SUPPORTED_YET](https://docs.rs/opensrv-mysql/0.1.0/opensrv_mysql/enum.ErrorKind.html#variant.ER_NOT_SUPPORTED_YET) more suitable here?

PR #69, opened 2022-06-28
chore: fix storage config

jiacai2050:

Closes #

# Rationale for this change
 

`data_path` config is changed in #20, but `example.toml` still use old format.

No
# How does this change test

Test manually.


ShiKaiWi (commented):
LGTM

PR #73, opened 2022-06-29
feat: new flush policy - purge flush

### LINKED ISSUE #68: Implement follower flush
**Description**

`flush` dumps memtable to sst file.

CeresDB current doesn't distinguish behaviors between leader and follower. In follower it only needs to purge its memtables and skip dumping.

**Proposal**

Implement the follower flush behavior described above. Serval steps I come up with:
- [ ] Receive and track the max sequence number flushed by leader
- [ ] Flush (drop) memtable based on the max seq
- [ ] (considering) Flush behavior after becoming the new leader.

**Additional context**

This issue may only cover the implementation part. The trigger will come with a follow up PR, which depends on WAL replication
and/or instance role.

waynexia:

Closes #68 

# Rationale for this change
 

Implement `Purge` flush policy. Detailed in #68.

- new flush option `TableFlushPolicy`
- new field in `TableFlushRequest`: `max_sequence` and `policy`
- change `MemTableView::pick_memtables_to_flush` to pick memtable based on sequence number rather than memtable id

no

# How does this change test

By existing UT

### DIFF
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -174,6 +174,7 @@ pub struct TableFlushRequest {
 }
 
 /// Policy of how to perform flush operation.
+#[derive(Debug)]
 pub enum TableFlushPolicy {
     /// Dump memtable to sst file.
     Dump,
@@ -279,12 +280,16 @@ where
     ) -> Result<TableFlushRequest> {
         let current_version = table_data.current_version();
         let last_sequence = table_data.last_sequence();
-        // Switch all mutable memtables
+        // Switch (freeze) all mutable memtables. And update segment duration if
+        // suggestion is returned.
         if let Some(suggest_segment_duration) =
             current_version.switch_memtables_or_suggest_duration(worker_local)
         {
-            info!("Switch memtable and suggest segment duration, table:{}, table_id:{}, segment_duration:{:?}", table_data.name, table_data.id, suggest_segment_duration);
-            assert!(suggest_segment_duration.as_millis() > 0);
+            info!(
+                "Update segment duration, table:{}, table_id:{}, segment_duration:{:?}",
+                table_data.name, table_data.id, suggest_segment_duration
+            );
+            assert!(!suggest_segment_duration.is_zero());
 
             let mut new_table_opts = (*table_data.table_options()).clone();
             new_table_opts.segment_duration = Some(ReadableDuration(suggest_segment_duration));
@@ -335,7 +340,7 @@ where
         let table = table_data.name.clone();
 
         let instance = self.clone();
-        let flush_job = async move { instance.flush_memtables_to_outputs(&flush_req).await };
+        let flush_job = async move { instance.flush_memtables(&flush_req).await };
 
         let compact_req = TableCompactionRequest::no_waiter(
             table_data.clone(),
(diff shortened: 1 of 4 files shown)


waynexia:
Two thoughts:
- deprecate `max_table_id` as `sequence_number` is the thing makes decisions
- consider reclaiming memtables as soon as they get dumped. They are removed together after all the memtables are dumped now.

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -167,6 +167,19 @@ pub struct TableFlushRequest {
...
+    pub policy: TableFlushPolicy,
+}
+
+/// Policy of how to perform flush operation.
+#[derive(Debug)]
+pub enum TableFlushPolicy {
+    /// Dump memtable to sst file.
+    Dump,
+    /// Drop memtables.
+    Purge,
```
ShiKaiWi, line 182: I guess we need one policy called `Block` which handles that the follower's memtables can't be flushed until the flushed sequence number from its leader is greater than its memtable to flush.

waynexia: This is covered in `pick_memtables_to_flush`. It only picks memtables whose `last_sequence` ≤ sequence number in the flush request.

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -167,6 +167,21 @@ pub struct TableFlushRequest {
...
     /// Max id of memtable to flush (inclusive).
     pub max_memtable_id: MemTableId,
+    /// Max sequence number to flush (inclusive).
+    pub max_sequence: SequenceNumber,
+    /// Flush policy.
+    pub policy: TableFlushPolicy,
+}
+
+/// Policy of how to perform flush operation.
+#[derive(Debug)]
```
ShiKaiWi, line 177: Shoult it derive Copy?

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -167,6 +167,21 @@ pub struct TableFlushRequest {
...
+pub enum TableFlushPolicy {
```
ShiKaiWi, line 178: Maybe this should be called `TableFlushAction` and the `Policy` may be reserved to for the trait to decide the `TableFlushRequest`, e.g.:
```rust
pub trait TableFlushPolicy {
    fn decide(&self, ...) -> Result<TableFlushRequest>;
}
```

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -350,40 +371,116 @@ where
...
-    /// Caller should guarantee flush of single table is sequential
-    async fn flush_memtables_to_outputs(&self, flush_req: &TableFlushRequest) -> Result<()> {
-        // TODO(yingwen): Record memtables num to flush as statistics
+    /// Each table can only have one running flush job.
+    async fn flush_memtables(&self, flush_req: &TableFlushRequest) -> Result<()> {
         let TableFlushRequest {
             table_data,
-            max_memtable_id,
+            // TODO: consider deprecate this field,
+            max_memtable_id: _,
```
ShiKaiWi, line 379: Can it be deprecated just in this pr?

ShiKaiWi (approved):

PR #74, opened 2022-06-29
feat: flush memtable in time when replay wal

ShiKaiWi:

Closes #

# Rationale for this change
 The memtable may cause oom if it can not be flushed in time when replaying wal.

Flush memtable when replaying wal.

None.

# How does this change test
Test manually.

### DIFF
--- a/analytic_engine/src/instance/open.rs
+++ b/analytic_engine/src/instance/open.rs
@@ -23,8 +23,8 @@ use crate::{
     context::OpenContext,
     instance::{
         engine::{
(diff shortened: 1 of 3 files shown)


waynexia (approved):

analytic_engine/src/instance/open.rs
```diff
@@ -326,7 +328,6 @@ where
             table_data.name, table_data.id, last_sequence
         );
 
-        // TODO(yingwen): Maybe we need to trigger flush if memtable is full during
         // recovery Replay entries
```
waynexia, line 331: one line remains

waynexia (approved):

PR #86, opened 2022-07-06
feat: integration test framework

### LINKED ISSUE #83: Integration test framework
**Description**

Provides a test framework to make us easy to write integration tests at SQL level.

**Proposal**

I plan to write this in python. It should be able to start server with special (non-default) configs, execute SQL and compare the result. Configuration, input data, SQL and expected output can be put in other text files.

**Additional context**

Other systems' framework:
- [MariaDB](https://github.com/MariaDB/server/blob/10.10/mysql-test/README)
- [Databend](https://github.com/datafuselabs/databend/blob/main/tests/README.md)
- [ClickHouse](https://github.com/ClickHouse/ClickHouse/blob/master/tests/clickhouse-test)
- [InfluxDB IOx](https://github.com/influxdata/influxdb_iox/blob/main/query_tests/README.md)

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

Closes #83

# Rationale for this change
 

This framework can make us easy to write integration tests at SQL level.

A new component that can run SQL integration tests.

This is implemented in rust. It can read SQL from a file, execute it to the server and compare the result with the expected. A case is made up of a ".sql" file and a ".result" file. 

It runs like:
```
Server from "/home/ruihang/repo/CeresDB/target/debug/ceresdb-server" is starting ...
Takes 49.020203ms. Diff: false. Test case "/home/ruihang/repo/CeresDB/tests/cases/example.sql" finished.
Run 1 finished. 0 cases are different.
```
An example case is included in this pr.

To use this harness, one should first build it under its crate root.
It now takes three env variable as parameters:
- `BINARY_PATH_ENV` defines the binary to test
- `SERVER_ENDPOINT_ENV` defines where server is set
- `CASE_ROOT_PATH_ENV` defines the cases dir

# How does this change test

By it self


server/src/grpc/query.rs
```diff
@@ -168,7 +168,10 @@ pub async fn fetch_query_output<C: CatalogManager + 'static, Q: QueryExecutor +
 fn convert_output(output: &Output) -> Result<QueryResponse> {
     match output {
         Output::Records(records) => convert_records(records),
-        _ => unreachable!(),
+        Output::AffectedRows(rows) => Err(ServerError::ErrNoCause {
```
jiacai2050, line 171: It seems a little strange to return this as error

tests/harness/src/runner.rs
```diff
@@ -0,0 +1,136 @@
...
+use crate::case::TestCase;
+
+/// Input SQL query cases
+const TEST_CASE_EXTENSION: &'static str = "sql";
+/// Output generated by running queries
+const OUTPUT_FILE_EXTENSION: &'static str = "out";
+/// Expected result provides with queries
+const RESULT_FILE_EXTENSION: &'static str = "result";
+/// Tools to compare file
+const DIFF_BINARY: &'static str = "diff";
```
jiacai2050, line 24: There is a prettydiff, it has color support, which is useful to review differences.
- https://github.com/romankoblov/prettydiff

tests/harness/src/runner.rs
```diff
@@ -0,0 +1,136 @@
...
+                let timer = Instant::now();
+                case.execute(&self.client, &mut output_file).await?;
+                let elapsed = timer.elapsed();
+
+                output_file.flush().await?;
+                let is_different = self.compare(&path);
+                if !is_different {
+                    remove_file(output_path).await?;
+                } else {
+                    diff_cases.push(path.as_os_str().to_str().unwrap().to_owned());
```
jiacai2050, line 58: Only collect path, no diff content?

tests/cases/example.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE demo (name string TAG, value double NOT NULL, t timestamp NOT NULL, TIMESTAMP KEY(t)) ENGINE=Analytic with (enable_ttl='false');
```
jiacai2050, line 1: If I have a new test case, whether should I create a new file or modify existing one?

Is there any suggested guideline on how to group test cases?

tests/cases/example.result
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE demo (name string TAG, value double NOT NULL, t timestamp NOT NULL, TIMESTAMP KEY(t)) ENGINE=Analytic with (enable_ttl='false');
```
jiacai2050, line 1: It seems a little inconvenient to repeat the input SQL.

Maybe we can only use one file(same with this result format), and let this framework to extract SQL out automatically. Any thoughts?

jiacai2050:
```test
# Simple sum.
eval instant at 50m SUM BY (group) (http_requests{job="api-server"})
  {group="canary"} 700
  {group="production"} 300

eval instant at 50m SUM BY (group) (((http_requests{job="api-server"})))
  {group="canary"} 700
  {group="production"} 300

# Test alternative "by"-clause order.
eval instant at 50m sum by (group) (http_requests{job="api-server"})
  {group="canary"} 700
  {group="production"} 300
```
This is what PromQL test cases look like.
- https://github.com/ceresdb/prometheus/blob/16bcbdaac510f1e5be7e532f7060a8a6f95e05d5/promql/testdata/aggregators.test#L11

tests/harness/src/runner.rs
waynexia: I don't intend to keep the diff content. If the result and expected are the same it will automatically remove the output. Otherwise the output will be kept and reported. So one only needs to check whether there is any new ".out" file generated after a run. 

If we wants to inspect which lines are diff we can use whatever diff tools we want. Another reason that not reporting the diff content directly is to keep the ".out" and ".result" files have the same format.

tests/harness/src/runner.rs
waynexia: Yep, explained above.

One more thing, the whole folder is tracked by git. So we won't miss any ".out" file if there is

tests/cases/example.sql
waynexia: I'll write a readme about how to use

tests/cases/example.result
waynexia: The ".result" file is not intended to be written out. Instead it's generated (usually it's renamed from a ".out" file). Thus it's ok to be redundant here. Repeat SQL can improve the readability of this file.

server/src/grpc/query.rs
waynexia: Yes. There are two reasons.
- The execution error is also a kind of execution result, which needs to be tracked in the test case.
- Our protocol currently doesn't support this field. This is a workaround and I'll file an issue for this.

This is definitely weird and should be fixed soon.

waynexia: Filed https://github.com/CeresDB/ceresdb/issues/89 for this. To my understanding the issue doesn't block this PR? cc @ShiKaiWi may have the same question

tests/cases/example.result
jiacai2050: Using one file have the same effect with what you said, and without any redundancies.

```org
select * from xx;

--result begin---
xxx
--result end---

select * from yy;

--result begin---
yyy
--result end---
```


When write a new test case, users can just write SQL, leaving the result part empty, then run the tests to get the results(in `out` file), then copy & paste that to `result` part to get job done.

tests/harness/src/case.rs
```diff
@@ -0,0 +1,153 @@
...
+                self.write_result(writer, format!("Failed to execute query, err: {:?}", e))
+                    .await?
+            }
+        }
+
+        Ok(())
+    }
+
+    // TODO: implement query interceptor.
+    fn intercept(&self, _result: QueriedRows) -> QueriedRows {
```
jiacai2050, line 127: This intercept is used for post process? If yes, in what situation would users need this(instead of using SQL function to archive same result)?

waynexia: E.g:
- require the runner to sort the result
- replace some non-constant, like timestamp
- display something out to the output

Some SQL's results won't always be the same and it isn't required to be the same. Like query without sort, or gets the current timestamp. This interceptor performs some pre- or post-process that makes us can still compare the result

tests/cases/example.result
waynexia: This is feasible in implementation. But the query result can be a lot. Consider an example: [SQL file](https://github.com/MariaDB/server/blob/10.10/mysql-test/main/intersect.test) and [result file](https://github.com/MariaDB/server/blob/10.10/mysql-test/main/intersect.result). I think it is hard to even find out a SQL you want in the result file. And the input SQLs in one file are usually causally related, putting results between them may make it harder to read.

jiacai2050: > And the input SQLs in one file are usually causally related, putting results between them may make it harder to read

This make sense to me. Let's wait see what others opinion is.

tests/harness/src/case.rs
jiacai2050: > require the runner to sort the result
> display something out to the output
> gets the current timestamp

Those can be done in SQL itself.

> Like query without sort

In this situation, what should I do in post-process in order to make it comparable?

FYI, I just think an easy-to-use post-process framework is not easy to design, if we have no clear requirements, we can think about it later.

tests/cases/example.sql
waynexia: Added

tests/harness/src/runner.rs
jiacai2050: > So one only needs to check whether there is any new ".out" file generated after a run.

When multiple testcases failed, How one can check all bad diff? `cat` output file one by one is kinds of inconvenient.

tests/harness/src/case.rs
jiacai2050: How about this syntax? 
```
-- postprocess
```
IMO It's more clear than before.

jiacai2050 (changes requested):

tests/harness/src/case.rs
waynexia: Preprocess like changing settings is also allowed.

tests/harness/src/case.rs
```diff
@@ -0,0 +1,153 @@
...
+impl Query {
+    fn push_interceptor(&mut self, post_process: String) {
+        self.interceptors.push(post_process);
+    }
+
+    fn append_query_line(&mut self, line: &str) {
+        self.query_lines.push(line.to_string());
+    }
+
+    async fn execute<W>(&self, client: &Client, writer: &mut W) -> Result<()>
```
jiacai2050, line 99: This execute only test via gRPC, should we also support MySQL?

tests/harness/src/runner.rs
waynexia: When multiple cases failed one may also want to fix them one by one. Mix up all the different together is unnecessary as the cases are organized by scenario. Failures in one file have a clear context, those across multiple files don't have.

tests/harness/src/case.rs
waynexia: One is enough I think. The difference is limited to the protocol level. Both protocols are using (and should use) the same underlying interfaces.

jiacai2050: > Both protocols are using (and should use) the same underlying interfaces.

Yes, we only test the frontend layer of MySQL, if we recommend MySQL protocol in production, then it's best this test cover that.

tests/harness/src/setup.rs
```diff
@@ -0,0 +1,55 @@
+use std::{
+    env,
+    process::{Child, Command, Stdio},
+};
+
+use ceresdb_client_rs::{client::Client, Builder};
+
+const BINARY_PATH_ENV: &'static str = "CERESDB_TEST_BINARY_PATH";
```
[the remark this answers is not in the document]
waynexia: done

tests/harness/src/case.rs
jiacai2050: Then like this?
```
-- interceptor
```

waynexia:
image

The diff is printed to stdout.

tests/harness/src/case.rs
waynexia: this might be too general. The preserved "keyword" should be more special?

jiacai2050: I'm fine this `keyword` if we haven't get a better one...

tests/harness/src/case.rs
waynexia: The protocol frontend is only used for communicating with server but not the target of those test cases. Protocols should have their own suits to guarantee the behavior.

jiacai2050 (approved):
LGTM

PR #93, opened 2022-07-07
fix: affected rows for create/drop statements default 0

### LINKED ISSUE #87: Affected rows reports wrong number for create/drop statements
**Describe this problem**

```plaintext
Failed to execute query, err: Server(ServerError { code: 500, msg: "Failed to convert output, query: DROP TABLE demo;. Caused by: Rpc error, code:500, message:Failed to convert output - affected rows: 1" })
```
A drop table tells one row is affected

**Steps to reproduce**

```
drop table demo;
```

**Expected behavior**

```
mysql> create table xx like meta_2 ;
Query OK, 0 rows affected (0.11 sec)
```

**Additional Information**

These SQLs don't actually count rows, 1 is the default value. Change the default value should fix this.

dust1:

Closes #87 

# Rationale for this change
 
bug fix

# How does this change test

no test

### DIFF
--- a/analytic_engine/src/table/mod.rs
+++ b/analytic_engine/src/table/mod.rs
@@ -219,7 +219,7 @@ impl<
             .await
             .map_err(|e| Box::new(e) as _)
             .context(AlterSchema { table: self.name() })?;
-        Ok(1)
+        Ok(0)
     }
 
     async fn alter_options(&self, options: HashMap<String, String>) -> Result<usize> {
@@ -228,7 +228,7 @@ impl<
             .await
             .map_err(|e| Box::new(e) as _)
             .context(AlterOptions { table: self.name() })?;
-        Ok(1)
+        Ok(0)
     }
 
     async fn flush(&self, request: FlushRequest) -> Result<()> {
--- a/interpreters/src/drop.rs
+++ b/interpreters/src/drop.rs
@@ -4,7 +4,7 @@
 
 use async_trait::async_trait;
 use catalog::{manager::Manager, schema::DropOptions};
-use log::info;
+use log::warn;
 use snafu::{Backtrace, OptionExt, ResultExt, Snafu};
 use sql::plan::DropTablePlan;
 use table_engine::engine::{DropTableRequest, TableEngineRef};
@@ -109,12 +109,12 @@ impl<C: Manager> DropInterpreter<C> {
             table_engine: self.table_engine,
         };
 
-        if !schema
+        if schema
             .drop_table(request, opts)
             .await
             .context(SchemaDropTable { table: &table })?
         {
-            info!("Failed to drop table {}", &table);
+            warn!("Table {} has been dropped already", &table);
         }
 
         Ok(Output::AffectedRows(0))
--- a/interpreters/src/drop.rs
+++ b/interpreters/src/drop.rs
@@ -4,6 +4,7 @@
 
 use async_trait::async_trait;
 use catalog::{manager::Manager, schema::DropOptions};
+use log::info;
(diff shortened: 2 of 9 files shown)


analytic_engine/src/table/mod.rs
```diff
@@ -219,7 +219,7 @@ impl<
             .await
             .map_err(|e| Box::new(e) as _)
             .context(AlterSchema { table: self.name() })?;
-        Ok(1)
+        Ok(0)
```
jiacai2050, line 222: If those alter statements always return 0 when success, isn't it better to just return `Result<()>`, and let the upper layer to return expected `affected_row`?

IMO, `affected_row` have no meaning in `Table` layer, only useful in upper layer.

dust1: yes! I'll change it later

jiacai2050: @dust1 Don't hurry, Let wait and see what others view is.

waynexia: We have some (intend to be informal) discussions in [slack]( https://ceresdbcommunity.slack.com/archives/C03KG4E4J3Z/p1657251293081759), if this link doesn't work you can try to join with [this one](https://join.slack.com/t/ceresdbcommunity/shared_invite/zt-1au1ihbdy-5huC9J9s2462yBMIWmerTw).

And the question we are discussing is how to define `affected_rows`, should it be "engine independent" or not.

waynexia: Considering CeresDB is designed to be multi-engine, we'd better not have hypotheses about how an engine will implement an operation. In this case `analytic_engine` won't change any row in an `ALTER TABLE`, but maybe an engine in the future would. So the `Table` trait should keep the return value.

interpreters/src/drop.rs
```diff
@@ -108,12 +109,15 @@ impl<C: Manager> DropInterpreter<C> {
...
         };
 
-        let dropped = schema
+        if !schema
             .drop_table(request, opts)
             .await
-            .context(SchemaDropTable { table: &table })?;
+            .context(SchemaDropTable { table: &table })?
+        {
+            info!("Failed to drop table {}", &table);
```
ShiKaiWi, line 117: The returned false indicates real drop does not happen (maybe the table is dropped by another requests) so this log seems not accurate.
```suggestion
            warn!("Table {} has been dropped already", &table);
```

sql/src/planner.rs
```diff
@@ -1289,7 +1289,7 @@ mod tests {
 
     #[test]
     fn test_show_tables_statement_to_plan() {
-        let sql = "SHOW TABLES;";
+        let sql = "show tables;";
```
ShiKaiWi, line 1292: Seems no need to change this?

interpreters/src/drop.rs
ShiKaiWi, line 117: The returned false indicates real drop does not happen (usually the table is dropped by another requests) so the log message may read:
```suggestion
            warn!("Table {} has been dropped already", &table);
```

interpreters/src/drop.rs
```diff
@@ -108,12 +109,15 @@ impl<C: Manager> DropInterpreter<C> {
             table_engine: self.table_engine,
         };
 
-        let dropped = schema
+        if !schema
```
ShiKaiWi, line 112: Maybe this `!` should be removed?

sql/src/planner.rs
dust1: this is to restart the workflow.🤣That bug caused me to modify it back and forth many times

ShiKaiWi (approved):
LGTM

PR #97, opened 2022-07-13
fix: Query fields are sorted by creation

### LINKED ISSUE #60: Query fields are sorted by creation
**Description**
ceresDB doesn't seem to require strict ordering of query result fields.

**Proposal**

Use the order in which fields are created.

**Additional context**

when I create table use: 
```
CREATE TABLE `hello_mysql` (`name` string TAG, `value` double NOT NULL, `t` timestamp NOT NULL, TIMESTAMP KEY(t)) ENGINE=Analytic with (enable_ttl='false');
```
and select it. his field order seems to be random? I found that the query result is wrapped in HashMap.
image

This doesn't seem to have any effect on normal usage, so I don't know if this situation needs improvement.

dust1:

Closes #60 

# Rationale for this change
 

This PR modifies the Response data structure in ceresdb/server/src/handlers/sql.rs. 
* ResponseRow: one record with response
* ResponseRows: type of Vec<ResponseRow>
* ResponseColumn: column struct, for keep column type and name
* Row: Helper type that is convenient for http serialize, this is one map
* mod http_format: ResponseRow in custom format by Serializer
* mod row_format: Row in custom format by Serializer

not

# How does this change test

### DIFF
--- a/server/src/handlers/sql.rs
+++ b/server/src/handlers/sql.rs
@@ -10,7 +10,10 @@ use common_types::{
 use interpreters::{context::Context as InterpreterContext, factory::Factory, interpreter::Output};
 use log::info;
 use query_engine::executor::RecordBatchVec;
-use serde::Serialize;
+use serde::{
+    ser::{SerializeMap, SerializeSeq},
+    Serialize,
+};
 use snafu::ensure;
 use sql::{
     frontend::{Context as SqlContext, Frontend},
@@ -22,8 +25,6 @@ use crate::handlers::{
     prelude::*,
 };
 
-type ResponseRows = Vec<ResponseRow>;
-
 #[derive(Debug, Deserialize)]
 pub struct Request {
     query: String,
@@ -34,11 +35,10 @@ pub struct Request {
 #[serde(rename_all = "snake_case")]
 pub enum Response {
     AffectedRows(usize),
-    #[serde(with = "http_format")]
     Rows(ResponseRows),
 }
 
-pub struct ResponseRow {
+pub struct ResponseRows {
     pub column_names: Vec<ResponseColumn>,
     pub data: Vec<Vec<Datum>>,
 }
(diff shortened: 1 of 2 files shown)


server/src/handlers/sql.rs
```diff
@@ -21,23 +22,39 @@ use crate::handlers::{
...
-    }
-}
-
 // TODO(yingwen): Improve serialize performance
 #[derive(Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum Response {
     AffectedRows(usize),
-    Rows(Vec<HashMap<String, Datum>>),
+    #[serde(with = "http_format")]
```
waynexia, line 37: We can `impl Serialize for ResponseRow` and let `#[derive(Serialize)]` cover this. Then the sub-mod `http_format` and `row_format` can be removed. So do the util struct `Row`.

server/src/handlers/sql.rs
```diff
@@ -128,26 +145,100 @@ fn convert_output(output: Output) -> ArrowResult<Response> {
...
+            let column_names = &record.column_names;
+            // let mut map = seq.serialize_map(Some(column_names.len()))?;
+
+            for rows in &record.data {
+                let data = rows
+                    .iter()
+                    .enumerate()
+                    .map(|(col_idx, datum)| {
+                        let column_name = &column_names[col_idx].name.clone();
+                        (column_name.clone(), datum.clone())
```
waynexia, line 215: Those clones might be avoidable. Maybe with some lifetime annotations on `Row` 🧐

waynexia (commented):
Thanks @dust1 for fixing this! The implementation generally looks good to me. I left some style comments.

And I'm wondering if we can add some assertions on the serialized JSON format in test (if `SerializeMap` can preserve the ordering, not sure).

waynexia (approved):
👍 This looks good to me. Thanks @dust1. I'm going to merge it this afternoon.

PR #99, opened 2022-07-16
feat: Support more expressions in predicate pushdown

### LINKED ISSUE #94: Pushdown `IN` predicate to TableScan
**Description**
Now the predicates whose type are only primitive binary expression are able to be pushed down to table scan. Actually `IN` predicate is also able to be pushed down and used to do filtering duration scanning stage.

**Proposal**
Only the [predicate module](https://github.com/CeresDB/ceresdb/blob/main/table_engine/src/predicate/mod.rs) in table engine needs changes.
- Make `IN` predicate able to be pushed down;
- Support the filtering RowGroups by `IN` expression;

**Additional context**
Shall we make more kinds of predicates able to be pushed down.

waynexia:

Closes #94

# Rationale for this change
 

Pushing more predicates down to data source (parquet) can improve performance. E.g.: `IN` expr

Remove the unnecessary pre-check that may prevent some expressions from being pushed down. Now we support to push down all expressions that are supported in data fusion https://github.com/apache/arrow-datafusion/blob/c528986d8c9df1340431644cd8b3f2f9e725abc8/datafusion/core/src/physical_optimizer/pruning.rs#L660

No

# How does this change test

new added test cases in `predicate` mod


ShiKaiWi (approved):
LGTM

Issue #72, opened 2022-06-29
Refactor EngineBuilder to remove duplicate code


chunshao90:
**Description**
Refactor EngineBuilder to remove duplicate code

**Proposal**
code: 
https://github.com/CeresDB/ceresdb/blob/890fecf10848a3d09b485af52ff516deccf771fc/analytic_engine/src/setup.rs#L92
https://github.com/CeresDB/ceresdb/blob/890fecf10848a3d09b485af52ff516deccf771fc/analytic_engine/src/setup.rs#L120
https://github.com/CeresDB/ceresdb/blob/890fecf10848a3d09b485af52ff516deccf771fc/analytic_engine/src/setup.rs#L155

**Additional context**

ygf11:
hi, chunshao90.

It seems we can make `Wal` as an `associate type` of EngineBuilder, then the build logic can be reused, like:
```rust
#[async_trait]
pub trait EngineBuilder: Default + Send + Sync {
    type Wal: WalManager + Send + Sync + 'static;

    /// Build the analytic engine from `config` and `engine_runtimes`.
    async fn build(
        &self,
        config: Config,
        engine_runtimes: Arc<EngineRuntimes>,
    ) -> Result<TableEngineRef> {
        assert!(!config.obkv_wal.enable);

        match config.storage {
            crate::storage_options::StorageOptions::Local(ref opts) => {
                let store = open_storage_local(opts.clone()).await?;
                let instance = self
                    .open_instance::<LocalFileSystem>(config.clone(), engine_runtimes, store)
                    .await?;
                Ok(Arc::new(TableEngineImpl::new(instance)))
            }
            crate::storage_options::StorageOptions::Aliyun(ref opts) => {
                let store = open_storage_aliyun(opts.clone()).await?;
                let instance = self
                    .open_instance::<AliyunOSS>(config.clone(), engine_runtimes, store)
                    .await?;
                Ok(Arc::new(TableEngineImpl::new(instance)))
            }
        }
    }

    async fn open_instance<Store>(
        &self,
        config: Config,
        engine_runtimes: Arc<EngineRuntimes>,
        store: Store,
    ) -> Result<InstanceRef<Self::Wal, ManifestImpl<Self::Wal>, Store, FactoryImpl>>
    where
        Store: ObjectStore;
}
```

Is this the expected to this task?

ShiKaiWi:
[...]

Issue #13, opened 2022-06-01
Tracking issue for OSS integration


waynexia:
## Target
- Support Aliyun OSS as storage backend.
- Able to extend OSS to other cloud services.

## Steps
- [x] #15
- [x] #19
- [x] #28
- [x] Local file system as test env #101

PR #101, opened 2022-07-17
refactor: remove `Store` type param in `Instance`

waynexia:

Related to #72 and #13

# Rationale for this change
 

We support various `ObjectStore` implements to be assembled in a recursive way. Like `CachedStore` -> `LocalStore` + `RemoteStore`. The existing type param way that requires fully specifying a type would be hard to implement. For simplicity and feasibility, I change this type param to a trait object. A related discussion can be found at https://github.com/CeresDB/ceresdb/issues/72.

The `CachedStore` has been implemented but the corresponding config is not added yet. This PR also provides a way to config a storage cache layer. Example config has been updated to `docs/example.toml`

This PR contains two changes:

- refactor: remove `Store` type param in `Instance`
- feat: add config for `CachedStore`

Maybe I need to separate it into two PRs

The config file has changed. 

# How does this change test

The refactor is tested by the compiler. New config is tested manually.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3283,6 +3283,8 @@ dependencies = [
  "lru",
  "object_store 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "oss-rust-sdk",
+ "serde",
+ "serde_derive",
  "snafu 0.6.10",
  "tempfile",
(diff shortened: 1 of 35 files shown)


ShiKaiWi (approved):
LGTM

PR #102, opened 2022-07-18
refactor: remove Fa type param in Instance

ygf11:

Related #72

# Rationale for this change
 
To support various `SstFactory` in Instance easily, we need make it as an inner `Trait Object`.

Remove `Fa` in `Instance`, and `various` place change to `SstFactoryRef`.

No

# How does this change test


waynexia (approved):
Wonderful! Thanks for your contribution ❤️

PR #108, opened 2022-07-19
refactor: feature gate `arrow_deps` in `common_types`

### LINKED ISSUE #103: Conditional import deps for `arrow-deps` sub-crate
**Description**

`arrow-deps` is used to delegate all arrow-related deps (arrow, parquet, datafusion etc.) to facilitate common deps' management. But for other crates that don't need all of them it brings lots of unnecessary deps - a relative discussion in python SDK: https://github.com/CeresDB/ceresdb-client-py/pull/2#issuecomment-1186925905

**Proposal**
- Remove this `arrow-deps` and let each crate import itself.
- Keep `arrow-deps` for sub-crates in this project. Other projects import necessary deps manually.
- Add feature gate to conditional import those deps.

**Additional context**

messense:

Closes #103

# Rationale for this change
 
Feature-gate heavy dependencies in `common_types` crate, allows downstream `ceresdb-client-py` to disable them.

Adds an on-by-default `arrow_deps` feature to make `arrow_deps` optional.

`arrow_deps` is a on-by-default feature, so it's not a breaking change.

# How does this change test

```bash
cd common_types
# build with default features
cargo build
# build without default features
cargo build --no-default-features
```

### DIFF
--- a/common_types/src/datum.rs
+++ b/common_types/src/datum.rs
@@ -4,6 +4,7 @@
 
 use std::{convert::TryFrom, fmt, str};
 
+#[cfg(feature = "arrow_deps")]
 use arrow_deps::{
     arrow::datatypes::{DataType, TimeUnit},
     datafusion::scalar::ScalarValue,
@@ -156,6 +157,7 @@ impl DatumKind {
 
     /// Create DatumKind from [arrow_deps::arrow::datatypes::DataType], if the
     /// type is not supported, returns None
+    #[cfg(feature = "arrow_deps")]
     pub fn from_data_type(data_type: &DataType) -> Option<Self> {
         match data_type {
             DataType::Null => Some(Self::Null),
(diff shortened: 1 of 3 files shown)


common_types/src/datum.rs
```diff
@@ -4,6 +4,7 @@
 
 use std::{convert::TryFrom, fmt, str};
 
+#[cfg(feature = "arrow_deps")]
```
ShiKaiWi, line 7: Now the cfg statements are distributed across too many places in this mod and maybe we can declare a new sub-mod called `arrow_convert` in this file into which all the methods and imports relavant to `arrow_deps` can be grouped. After that we can use only one `#[cfg(feature = "arrow_deps")]` to control the mod `arrow_convert`.

And I guess this parttern can be applied to other mods, such as `column`, `column_schema` and so on.

ShiKaiWi: Exactly!

ShiKaiWi (approved):
LGTM

PR #116, opened 2022-07-20
fix: table name with quote

### LINKED ISSUE #113: table name with back quote has strange behavior in SDK
**Describe this problem**

run log:
```
DROP TABLE IF EXISTS `t0`;

affected_rows: 0

CREATE TABLE `t0`(a int, t timestamp NOT NULL, TIMESTAMP KEY(t)) ENGINE = Analytic;

affected_rows: 0

INSERT INTO TABLE `t0`(a, t) values(1,1);

affected_rows: 1

SELECT * FROM `t0`;

Failed to execute query, err: Server(ServerError { code: 500, msg: "Failed to create plan, query: SELECT * FROM `t0`;. Caused by: Failed to create plan, err:DataFusion Failed to plan, err:Execution error: MetaProvider not found" })
```

**Steps to reproduce**

add the following to a integration test
```sql
DROP TABLE IF EXISTS `t0`;

CREATE TABLE `t0`(a int, t timestamp NOT NULL, TIMESTAMP KEY(t)) ENGINE = Analytic;

INSERT INTO TABLE `t0`(a, t) values(1,1);

SELECT * FROM `t0`;
```

and run it

**Expected behavior**

the last SQL
```sql
SELECT * FROM `t0`;
```

should return data instead of reporting an error.

**Additional Information**

the client version is 
```
https://github.com/CeresDB/ceresdb-client-rs.git?rev=a935a0ebacd1304d69d25e3e8611afa7d64611d6
```

jiacai2050:

Closes #113 

# Rationale for this change
 

Table Name (or other identifiers) with back quotes should be treated as literal token. But our interpreter will use the token with back quotes which will lead to some inconsistent behaviors like reported in #113 

Replace `ObjectName` with `TableName` in parsed SQL statement to override the `ToString` implement, which will remove quotes and join every token with `.`

No

# How does this change test

Unit test

### DIFF
--- a/sql/src/ast.rs
+++ b/sql/src/ast.rs
@@ -26,6 +26,32 @@ pub enum Statement {
     Exists(ExistsTable),
 }
 
+#[derive(Debug, PartialEq)]
+pub struct TableName(ObjectName);
+
+impl TableName {
+    pub fn is_empty(&self) -> bool {
+        self.0 .0.is_empty()
+    }
+}
+
+impl ToString for TableName {
+    fn to_string(&self) -> String {
+        self.0
+             .0
+            .iter()
+            .map(|ident| ident.value.as_str())
+            .collect::<Vec<_>>()
+            .join(".")
+    }
+}
+
+impl From<ObjectName> for TableName {
+    fn from(object_name: ObjectName) -> Self {
+        Self(object_name)
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum ShowCreateObject {
@@ -37,7 +63,7 @@ pub struct CreateTable {
     /// Create if not exists
     pub if_not_exists: bool,
     /// Table name
-    pub name: ObjectName,
+    pub table_name: TableName,
     pub columns: Vec<ColumnDef>,
     pub engine: String,
     pub constraints: Vec<TableConstraint>,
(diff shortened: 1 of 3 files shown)


sql/src/ast.rs
```diff
@@ -26,6 +26,32 @@ pub enum Statement {
     Exists(ExistsTable),
 }
 
+#[derive(Debug, PartialEq)]
+pub struct TableName(ObjectName);
+
+impl TableName {
+    pub fn is_empty(&self) -> bool {
+        self.0 .0.is_empty()
```
waynexia, line 34:
```suggestion
        self.0.0.is_empty()
```

sql/src/ast.rs
```diff
@@ -26,6 +26,32 @@ pub enum Statement {
...
+}
+
+impl ToString for TableName {
+    fn to_string(&self) -> String {
+        self.0
+             .0
+            .iter()
+            .map(|ident| ident.value.as_str())
+            .collect::<Vec<_>>()
+            .join(".")
```
waynexia, line 45:
```suggestion
    fn to_string(&self) -> String {
        self.0
            .0
            .iter()
            .map(|ident| ident.value.as_str())
            .collect::<Vec<_>>()
            .join(".")
```

waynexia (approved):
Good job :+1: But looks like the codes are not formatted, could you please check them?

waynexia:
Also wondering why ci can pass 🧐

sql/src/ast.rs
jiacai2050: same with before.

sql/src/ast.rs
waynexia: okay, TIL

sql/src/ast.rs
waynexia: I can understand the first one but this is pretty strange to me. anyway it goes out of the scope of this PR.

PR #119, opened 2022-07-21
refactor: make read_batch object safe in WalManager

ygf11:

Part of #72

# Rationale for this change
 
Make `WalManager` object safe, so we can use `Arc<dyn WalManager>` in other place.

1. Remove `LogReader` trait.
2. The item of `BlockingLogIterator` is `Vec<u8>` now.
3. Real value decoding will do in `BatchLogIteratorAdapter`.
4. Move LogKeyEncoder/LogValueEncoder etc from `wal::rocks_impl::encoding` to `wal::kv_encoder`, which are both used by `rocks_impl` and `table_kv_impl`. 

No

# How does this change test


ygf11:
@ShiKaiWi @waynexia PTAL.

components/table_kv/src/obkv/tests.rs
```diff
@@ -71,6 +71,7 @@ impl ObkvTester {
         let mut iter = self.obkv.scan(ctx, table_name, scan_req).unwrap();
 
         let mut pairs = Vec::new();
+        println!("scan");
```
waynexia, line 74: A remaining debug print

ygf11: Addressed.

waynexia (commented):
Thanks @ygf11. This patch generally looks good to me. I left a comment about `BlockingLogIterator`'s item. This is a great step towards a neat `WalManager` 👍

wal/src/manager.rs
```diff
@@ -216,14 +219,11 @@ pub struct ReadRequest {
...
-pub trait BlockingLogIterator {
+pub trait BlockingLogIterator: Send + fmt::Debug {
     /// Fetch next log entry from the iterator.
     ///
     /// NOTE that this operation may **BLOCK** caller thread now.
-    fn next_log_entry<D: PayloadDecoder>(
-        &mut self,
-        decoder: &D,
-    ) -> Result<Option<LogEntry<D::Target>>>;
+    fn next_log_entry(&mut self) -> Result<Option<LogEntry<Vec<u8>>>>;
```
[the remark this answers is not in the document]
ygf11: > Clone the bytes and keep it in TableLogIterator to decouple the lifetime.

Sorry I misunderstanding the idea where you mentioned in #120. 

Let me add a `Vec<u8>` to `TableLogIterator` as the reference source.

I think it is what you want :D.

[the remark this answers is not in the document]
ygf11: Addrressed

wal/src/table_kv_impl/region.rs
```diff
@@ -691,20 +691,26 @@ impl<T: TableKv> BlockingLogIterator for TableLogIterator<T> {
...
             .context(manager::Encoding)?;
-        let log_entry = LogEntry {
-            sequence: self.current_log_key.1,
-            payload,
-        };
+
+        // To unblock pr#119, we use the following to simple resolve borrow-check error.
+        // detail info: https://github.com/CeresDB/ceresdb/issues/120
+        let value_start = self.values.len();
+        self.values.extend_from_slice(payload);
```
waynexia, line 701: Is it necessary to keep previous bytes rather than drop them? How about
```rust
self.values = payload.to_owned();
// ...
&self.values;
```

ygf11: Make sense.

waynexia (approved):
Thanks again :heart:

PR #140, opened 2022-07-25
feat: Simplify SQL syntax that specifies timestamp key column

### LINKED ISSUE #115: Simplify SQL syntax that specifies timestamp key column
**Description**

`TIMESTAMP KEY` is a special column in analytic engine. It need to be set to a column with `Timestamp` type and with `NOT NULL` constrain.
```sql
CREATE TABLE demo (
    name string TAG,
    value double NOT NULL,
    t timestamp NOT NULL,
    TIMESTAMP KEY(t)
) ENGINE=Analytic;
```

It must have one and only one `TIMESTAMP KEY` column in a table.

**Proposal**
there are a few ways to simplify it.

---------

- Specify `TIMESTAMP KEY` without create a column
Considering the constraints, user is actually giving a name to the `TIMESTAMP KEY` column. 
```sql
CREATE TABLE demo (
    name string TAG,
    value double NOT NULL,
    TIMESTAMP KEY(t)
) ENGINE=Analytic;
```

------------

- Move `TIMESTAMP KEY` to column definition.
This makes `TIMESTAMP KEY` act like primary key.
```sql
CREATE TABLE demo (
    name string TAG,
    value double NOT NULL,
    t timestamp NOT NULL TIMESTAMP KEY
) ENGINE=Analytic;
```

-----------
- Do not need to define it and let CeresDB infers the `TIMESTAMP KEY`
Allow user to not specify `TIMESTAMP KEY` in create table SQL. CeresDB will try to find a column with `Timestamp` type and `NOT NULL`. Report error if zero or multiple columns are found.
```sql
CREATE TABLE demo (
    name string TAG,
    value double NOT NULL,
    t timestamp NOT NULL
) ENGINE=Analytic;
```

**Additional context**

dust1:

Closes #115 
ref #154 

# Rationale for this change
 

I added 3 structs just to transmit the `is_timestamp` tag.
in `ast.rs`

now, user can use this sql to create table.
```
CREATE TABLE demo (
    name string TAG,
    value double NOT NULL,
    t timestamp NOT NULL TIMESTAMP KEY
) ENGINE=Analytic;
```

# How does this change test

# note
image


sql/src/ast.rs
```diff
@@ -107,6 +111,141 @@ pub struct ExistsTable {
     pub table_name: TableName,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct CeresColumnOptionDef {
```
ShiKaiWi, line 116: The current implementation is a little bit complex and maybe a better and more simple way is to use `ColumnOption::DialectSpecific` provided by `sql-parser`.

And I guess you can refer to the implementation for `comment`:
https://github.com/CeresDB/ceresdb/blob/037735eb80a27f2cd476a4e1d443747fbee098fc/sql/src/parser.rs#L474

ShiKaiWi: We can record the `TimestampKey` information as `ColumnOption::DialectSpecific` in `ColumnDef.options` and afterwards we read that during generating `ColumnSchema`:
https://github.com/CeresDB/ceresdb/blob/037735eb80a27f2cd476a4e1d443747fbee098fc/sql/src/planner.rs#L668

dust1: ok🧐

waynexia:
We now have [integration test](https://github.com/CeresDB/ceresdb/blob/main/tests/README.md#add-a-test) suite for this kind of changes. Would you like to add this new syntax to it? Maybe some new SQLs in https://github.com/CeresDB/ceresdb/blob/main/tests/cases/local/05_ddl/create_tables.sql 🤗

dust1:
> We now have [integration test](https://github.com/CeresDB/ceresdb/blob/main/tests/README.md#add-a-test) suite for this kind of changes. Would you like to add this new syntax to it? Maybe some new SQLs in https://github.com/CeresDB/ceresdb/blob/main/tests/cases/local/05_ddl/create_tables.sql 🤗

sure! not adding test cases is my mistake😂

waynexia:
> > We now have [integration test](https://github.com/CeresDB/ceresdb/blob/main/tests/README.md#add-a-test) suite for this kind of changes. Would you like to add this new syntax to it? Maybe some new SQLs in https://github.com/CeresDB/ceresdb/blob/main/tests/cases/local/05_ddl/create_tables.sql 🤗
> 
> sure! not adding test cases is my mistake😂

Not your fault! This is newly added. I'll add it to our dev guide

sql/src/parser.rs
```diff
@@ -352,8 +352,39 @@ impl<'a> Parser<'a> {
                 );
             }
         }
+        let mut cols: Vec<ColumnDef> = vec![];
```
ShiKaiWi, line 355: I guess there is no need to reconstruct the `ColumnDef`s because there is no influence to keep the extra TimestampKey option in the `column.options`.

So I guess what we need here is just to build `TableConstraint` if possible.

sql/src/parser.rs
```diff
@@ -352,8 +352,39 @@ impl<'a> Parser<'a> {
...
+                                });
+                            }
+                            t => warn!("Unsupported Keyword in column option: {}", t),
+                        }
+                    }
+                    _ => new_opts.push(option),
+                }
+            }
+            column.options = new_opts;
+            cols.push(column);
```
ShiKaiWi, line 384: I think we can put this procedure as a method called `try_build_timestamp_key_constraint` for the simplicity of the `parse_columns`.

The method may looks like:
```rust
fn try_build_tskey_constraint(cols: &[ColumnDef]) -> Option<TableConstraint> {
...
}
```

sql/src/parser.rs
```diff
@@ -352,8 +352,39 @@ impl<'a> Parser<'a> {
...
+            let mut new_opts = vec![];
+            for option in options {
+                match &option.option {
+                    ColumnOption::DialectSpecific(ts) => {
+                        if ts.is_empty() || ts.len() != 1 {
+                            continue;
+                        }
+                        let token = &ts[0];
+                        match token {
+                            Token::Word(w) if w.value.eq(TS_KEY) => {
```
ShiKaiWi, line 367: I guess we put it more simpler:
```rust
if let [Token::Word(keyword)] = &tokens[..] {
    if keyword.value == TS_KEY {
       ...
    }
}
```

sql/src/parser.rs
```diff
@@ -352,8 +352,39 @@ impl<'a> Parser<'a> {
                 );
             }
         }
+        let mut cols: Vec<ColumnDef> = vec![];
+        for mut column in columns {
+            let options = column.options;
+            let mut new_opts = vec![];
+            for option in options {
+                match &option.option {
+                    ColumnOption::DialectSpecific(ts) => {
```
ShiKaiWi, line 361: `ts` is confusing because `TS_KEY` is already introduced and maybe we can call it `tokens`.

ShiKaiWi:
@dust1 Thanks. It looks much better but I guess it can be made more simple.

sql/src/parser.rs
```diff
@@ -492,9 +503,42 @@ impl<'a> Parser<'a> {
     }
 }
 
+fn try_build_tskey_constraint(col_defs: &[ColumnDef]) -> Option<TableConstraint> {
```
ShiKaiWi, line 506: One more thing we may need to take considerations is that it is possible to define multiple tskey constraints and that is not allowed. So we may need to make the result type is: `Result<Option<TableConstraint>>`.

ShiKaiWi (commented):
@dust1 Now it looks clean and pretty. And now maybe we should pay some attention to user's wrong use case.

ShiKaiWi (approved):
Thanks！

PR #149, opened 2022-07-26
docs: Update user guide summary

jiacai2050:

Closes #

# Rationale for this change

- There is some duplication in README and user guide
-  About page of user guide is empty now

Remove duplicated content in README and add a short description in about page

No

# How does this change test
No need

### DIFF
--- a/docs/user-guide/src/about.md
+++ b/docs/user-guide/src/about.md
@@ -1,5 +1,4 @@
-
-CeresDB is a timeseries database. However, CeresDB's goal is to handle both timeseries and analytic workloads compared with the traditional ones, which usually have a poor performance in handling analytic workloads.
+CeresDB is a timeseries database that aims to handle both timeseries and analytic workloads efficiently.
 
 # Motivation
 In the traditional timeseries database, the `Tag` columns (InfluxDB calls them `Tag` and Prometheus calls them `Label`) are normally indexed by generating an inverted index. However, it is found that the cardinality of `Tag` varies in different scenarios. And in some scenarios the cardinality of `Tag` is very high, and it takes a very high cost to store and retrieve the inverted index. On the other hand, it is observed that scanning+pruning often used by the analytical databases can do a good job to handle such these scenarios.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,6 +4,7 @@
 [![CI](https://github.com/CeresDB/ceresdb/actions/workflows/ci.yml/badge.svg)](https://github.com/CeresDB/ceresdb/actions/workflows/ci.yml)
 [![OpenIssue](https://img.shields.io/github/issues/CeresDB/ceresdb)](https://github.com/CeresDB/ceresdb/issues)
 [![Slack](https://badgen.net/badge/Slack/Join%20CeresDB/0abd59?icon=slack)](https://join.slack.com/t/ceresdbcommunity/shared_invite/zt-1dcbv8yq8-Fv8aVUb6ODTL7kxbzs9fnA)
+[![Docker](https://img.shields.io/docker/v/ceresdb/ceresdb-server?logo=docker)](https://hub.docker.com/r/ceresdb/ceresdb-server)
 
 [中文](./README-CN.md)
```
waynexia, line 9: Those change should be taken in CN version as well

docs/user-guide/src/about.md
```diff
@@ -0,0 +1,12 @@
+
```
waynexia, line 1: empty line

docs/user-guide/src/about.md
```diff
@@ -0,0 +1,12 @@
...
+CeresDB is a timeseries database. However, CeresDB's goal is to handle both timeseries and analytic workloads compared with the traditional ones, which usually have a poor performance in handling analytic workloads.
```
waynexia, line 2: I write a simplified one:

```suggestion
CeresDB is a timeseries database that aims to handle both timeseries and analytic workloads efficiently.
```

Motivation:
- the _However_ is a bit abrupt
- to my understanding, traditional is bad at spreading tags, not every analytic scenario. So I think only to state we can handle both is enough as a short introduction.

README.md
waynexia: Good 👍

waynexia (approved):

PR #159, opened 2022-07-28
refactor: make write method object safe in WalManager

ygf11:

Part of #72 

# Rationale for this change
 
Make WalManager object safe, so we can use Arc<dyn WalManager> in other place.

1. Remove LogWriter, and move `write` to `WalManager`.
2. Remove generic and associate params in `Payload`.

No

# How does this change test


components/bytes/src/lib.rs
```diff
@@ -245,6 +251,20 @@ impl<'a, B: MemBufMut> Write for Writer<'a, B> {
...
+            Error::WouldOverflow { .. } => io::Error::new(io::ErrorKind::WriteZero, e),
+        })?;
+        Ok(src.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
```
ygf11, line 267: `&mut dyn MemBufMut` can not be wrapped to `Writer`, because the size of `dyn MemBufMut` can not know at compile time.
So I directly implement `Write` for `&mut dyn MemBufMut`.

components/bytes/src/lib.rs
```diff
@@ -245,6 +251,20 @@ impl<'a, B: MemBufMut> Write for Writer<'a, B> {
     }
 }
 
+impl Write for &mut dyn MemBufMut {
```
ygf11, line 254: &mut dyn MemBufMut can not be wrapped to Writer, because the size of dyn MemBufMut can not know at compile time.
So I directly implement Write for &mut dyn MemBufMut.

ygf11:
@waynexia PTAL.

waynexia:
Fancinating! I plan to review this later this day 🤩

waynexia (approved):
Thanks for submitting this PR. It looks good to me 👍 

For overhead introduced here (`Payload` becomes a trait object) we can find some way to alleviate it like changing the `write` interface to directly accept encoded bytes. But this is beyond the scope of the refactor this time. And we are lacking of benchmark so don't know how much regression is brought actually.

I'm going to merge tomorrow if there is no other review.

ygf11:
Thanks for your suggestions.

> For overhead introduced here (Payload becomes a trait object) we can find some way to alleviate it like changing the write interface to directly accept encoded bytes. 

I feel the same. we can decouple encoding from write of `WalManager`. I plan to create a issue later about this, we can talk deeply there :D

PR #164, opened 2022-08-03
refactor: avoid validation for table constraints during parse

ShiKaiWi:

Closes #

# Rationale for this change
Disallow for mutiple definitions of timestamp key is checked and should be checked by `planner`: https://github.com/CeresDB/ceresdb/blob/75e9b7ea549cde7cd906f008a3879ebed6a9db57/sql/src/planner.rs#L294

- Avoid validation for exact only one timestamp key definition during parsing procedure.
- Add test for defining multiple timestamp key constraints.

No.

# How does this change test
Integration test.


ShiKaiWi:
@dust1 I find that the disallow for multiple timestamp keys definition has been checked in planner module so there is no need to do that in the parser module. Sorry for misleading you and PTAL at this PR.

dust1:
that's okay.😃
do I need to submit a new PR to remove the verification in the parser module? Or is it up to you to remove it in this PR?

tests/cases/.gitignore
```diff
@@ -0,0 +1 @@
+*.out
```
waynexia, line 1: don't ignore the `.out` file. It should be tracked and we need to guarantee no `.out` will be added.

sql/src/parser.rs
```diff
@@ -504,63 +501,30 @@ impl<'a> Parser<'a> {
...
-    }
-    Err(ParserError::ParserError(
-        "Failed to parser, not allowed define mutiple tskey constraints.".to_string(),
-    ))
-}
-
-fn try_build_tskey_constraint(col_defs: &[ColumnDef]) -> Result<Option<TableConstraint>> {
-    let mut constraint = None;
+// Build the tskey constraint from the column definitions if any.
+fn try_build_timestamp_key_constraint(
```
waynexia, line 505: the `try_` prefix in fn name is not appropriate since this function won't fail

waynexia (commented):
👍

ShiKaiWi:
> that's okay.😃 do I need to submit a new PR to remove the verification in the parser module? Or is it up to you to remove it in this PR?

Let this pr remove it. Thanks.

tests/cases/.gitignore
ShiKaiWi: Should it be guaranteed that no `.out` is added if we ignore it?

ShiKaiWi: I see. We can find the test failure by checking the existence of these `.out` files.

waynexia (approved):
Good job 👍

PR #166, opened 2022-08-04
feat: persist `AlterSchema` and `AlterOptions` to data WAL

### LINKED ISSUE #82: Move `AlterSchema` and `AlterOption` to data WAL
**Description**

`AlterSchema` and `AlterOption` are now persisted in the manifest. It would be better to keep manifest simple (only contains SST file information, sequence number, etc.). And schema modification is related to data, putting them in two separate places may bring some problems.

**Proposal**

Make manifest only to contain create, drop table, and version edit. The new procedure of altering schema and option is:
- trigger flush  
- persist new schema/option to WAL  
- update OSS metadata (next PR)  
- update manifest to include the DDL record  
- update memory status  

A new file that contains "present schema" and "present option" will be persisted to OSS as table's metadata.

**Additional context**

waynexia:

Closes #82

# Rationale for this change
 
quote from #82:
>`AlterSchema` and `AlterOption` are now persisted in the manifest. It would be better to keep manifest simple (only contains SST file information, sequence number, etc.). And schema modification is related to data, putting them in two separate places may bring some problems.

This PR focus on the second target -- put DDL (altering schema and options) together with data WAL.

A more detailed reason is that in distributed mode, we have a writer node and serval reader nodes. Where the writer serves all the related write requests, including writing data and altering schema/options. And readers sync new data from their writer. They communicate with data WAL. However, the data WAL used only contains the data logs. This is problematic when for example, the schema is changing. Readers don't know at which point the schema has changed, and therefore cannot handle their memory status correctly.

After this patch readers can know when to update their schema and/or options. Those DDL records are associated with data records.

Besides this, those DDL records still exist in the Manifest. I.e. this PR doesn't touch the Manifest. I haven't found a better place to put those metadata. This kind of data changes frequently thus OSS is not a good choice.

Duplicate `AlterSchema` and `AlterOptions` to data WAL

No, all changes are internal.

# How does this change test

Unit test

### DIFF
--- a/analytic_engine/src/instance/alter.rs
+++ b/analytic_engine/src/instance/alter.rs
@@ -1,6 +1,6 @@
 // Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
 
-//! Alter schema logic of instance
+//! Alter [Schema] and [TableOptions] logic of instance.
 
 use std::{collections::HashMap, sync::Arc};
 
--- a/analytic_engine/src/instance/alter.rs
+++ b/analytic_engine/src/instance/alter.rs
@@ -8,13 +8,16 @@ use log::info;
 use snafu::{ensure, ResultExt};
 use table_engine::table::AlterSchemaRequest;
 use tokio::sync::oneshot;
-use wal::manager::WalManager;
+use wal::{
+    log_batch::{LogWriteBatch, LogWriteEntry},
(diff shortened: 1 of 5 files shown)


ShiKaiWi (approved):
Well done!

waynexia:
I've merged main to resolve conflicts with #163, PTAL again @ShiKaiWi

PR #174, opened 2022-08-07
fix:select system.public.tables will return error

### LINKED ISSUE #129: select system.public.tables will return error
**Describe this problem**
```
SELECT
    `timestamp`,
    `catalog`,
    `schema`,
    `table_name`,
    `engine`
FROM
    system.public.tables
WHERE
    table_name = '01_system_table1';

```
Currently, execute SQL above will cause an error

```

Failed to execute query, err: Server(ServerError { code: 500, msg: "Failed to execute interpreter, query: SELECT     `timestamp`,     `catalog`,     `schema`,     `table_name`,     `engine` FROM     system.public.tables WHERE     table_name = '01_system_table1';. Caused by: Failed to execute select, err:Failed to execute logical plan, err:Failed to collect record batch stream, err:Stream error, msg:Convert from arrow record batch, err:External error: Execution error: Failed to read table, partition:3, err:Failed to scan table, table:tables, err:Failed to append datum, err:Data type conflict, expect:UInt64, given:String." })

```

**Steps to reproduce**

Execute SQL above

**Expected behavior**

No error

**Additional Information**

https://github.com/CeresDB/ceresdb/blob/334bc6ce13aac2e72a0771419b4aefe2164255e9/tests/cases/01_system/system_tables.result#L11

ref #154

dust1:

Closes #129 

# Rationale for this change
 
select system.public.tables will return error

I modified the creation method of record_schema_with_key. The original creation method cannot create a RecordBatch of the corresponding capacity according to the request parameters in the request, resulting in an error when filling the data

https://github.com/CeresDB/ceresdb/blob/main/system_catalog/src/tables.rs#141 

However, I found that when I executed the following statement, the returned `RecordBatchVec` was empty,I will find the reason for this problem later
```
SELECT
    `timestamp`,
    `catalog`
FROM
    system.public.tables
WHERE
    table_name = '01_system_table1';
```

none

# How does this change test

none

### DIFF
--- a/system_catalog/src/tables.rs
+++ b/system_catalog/src/tables.rs
@@ -138,8 +138,8 @@ impl<M: Manager> SystemTable for Tables<M> {
             .all_catalogs()
             .map_err(|e| Box::new(e) as _)
             .context(table_engine::table::Scan { table: self.name() })?;
-        let mut builder =
-            RecordBatchWithKeyBuilder::new(self.schema.clone().to_record_schema_with_key());
+        let record_schema_key = request.projected_schema.to_record_schema_with_key();
+        let mut builder = RecordBatchWithKeyBuilder::new(record_schema_key);
 
         let projector = request
             .projected_schema
--- a/system_catalog/src/tables.rs
+++ b/system_catalog/src/tables.rs
@@ -138,8 +138,8 @@ impl<M: Manager> SystemTable for Tables<M> {
             .all_catalogs()
             .map_err(|e| Box::new(e) as _)
             .context(table_engine::table::Scan { table: self.name() })?;
-        let record_schema_key = request.projected_schema.to_record_schema_with_key();
-        let mut builder = RecordBatchWithKeyBuilder::new(record_schema_key);
+        let projected_record_schema = request.projected_schema.to_record_schema_with_key();
+        let mut builder = RecordBatchWithKeyBuilder::new(projected_record_schema);
 
         let projector = request
             .projected_schema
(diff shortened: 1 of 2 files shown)


system_catalog/src/tables.rs
```diff
@@ -138,8 +138,8 @@ impl<M: Manager> SystemTable for Tables<M> {
             .all_catalogs()
             .map_err(|e| Box::new(e) as _)
             .context(table_engine::table::Scan { table: self.name() })?;
-        let mut builder =
-            RecordBatchWithKeyBuilder::new(self.schema.clone().to_record_schema_with_key());
+        let record_schema_key = request.projected_schema.to_record_schema_with_key();
```
Rachelint, line 141: `projected_record_schema` may be better ?

dust1: I'll try it later.

Rachelint: Everything else seems OK to me.

Rachelint: I think of something, you should also update the related test in `tests/cases/00_dummy` to demonstrate it .

dust1: ah yes i forgot about it. thanks for reminding me

dust1: hello @Rachelint , I don't quite understand what `projected_record_schema` means, I tried to modify the `Request` last night but there is no `projected_record_schema` parameter in it. Can you be a little more detailed?😥

Rachelint: > hello @Rachelint , I don't quite understand what `projected_record_schema` means, I tried to modify the `Request` last night but there is no `projected_record_schema` parameter in it. Can you be a little more detailed?😥

Oh, I just think the name `record_schema_key` is  a bit strange ... and suggest to renaming it.

Rachelint (approved):

Rachelint:
Thanks @dust1 .

Issue #175, opened 2022-08-08
Implement table roles


waynexia:
**Description**

In distributed mode a table might be one of three roles: `Writer`, `Reader` and `Querier` based on the cluster's status. We need to implement these roles. The server only contains logic when a table is `Writer`, which can write data, append WAL, modify table options or perform compaction. Others two roles are
- `Reader` which only replicates WAL from the corresponding `Writer`, and provides the same query ability as `Writer`
- `Querier` which won't communicate with WAL. It is designed to only provides query ability on top of data in OSS. But we can let `Reader` provide those un-flushed fresh data in the future.

**Proposal**

This has serval parts:
- How __Role__ is represented
  Introduce a trait `RoleTable` and three impls `WriterTable`, `ReaderTable` and `QuerierTable`. The different logics of the same interface are encapsulated inside. This obstruction should only be used to group operation logic. The underlying `TableData` is the same one for different roles (for one table of course).
- Changes to the existing mental level
  The current hierarchy is `Instance` -> `Spaces` -> `Tables`, I propose to insert `RoleTable` between `Space` and `Table`. A `RoleTable` (and its three implementations) holds one `TableData`, and provides top-level interfaces to operate on the data like write or read. We are operating those data directly in `Instance`.
- How to manage the state of table's role
  `RoleTable` provides two mechanisms for syncing roles.
  - A atomic flag to represent current status.
    In most cases we are using `RoleTable` with type like `Arc<dyn RoleTable>`. That `Arc` handler cannot be changed quickly, thus an atomic flag is used to show what the actual status is. This mechanism is (designed) for rejecting calls to incorrect role. E.g., A table changed from `Writer` to `Reader` but the `WriterTable` handler is still kept by some tasks. We can set the flag and fail those tasks.
  - A notifier that sends a message on the last reference is dropped.
    Role changing is a long procedure. We might want to do something after the previous role's handles are fully dropped. This notifier is for that.
- How does WAL replicate (#179)
  WAL is an important and complex part in this "role system". `Writer` wants to write to it and `Reader` should keep reading from it. The proposed workflow looks like the following, I plan to use one replicator per `Instance` to replicate all the tables:
```
          register IDs
         need replicate
     ┌─────────────────────┐
     │                     │
     │              ┌──────▼───────┐
┌────┴─────┐        │  background  │
│Role Table│        │WAL Replicator│
└────▲─────┘        └──────┬───────┘
     │                     │
     └─────────────────────┘
          replicate log
[...]

PR #179, opened 2022-08-08
feat: `WalReplicator` skeleton implementation

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

ref #175

# Rationale for this change
 

Follow the context from #175. `ReaderTable` needs to get the newest data from the corresponding `WriterTable` via WAL. This implements a replicator that keeps polling data from WAL.

This replicator works as following:
```
          register IDs
         need replicate
     ┌─────────────────────┐
     │                     │
     │              ┌──────▼───────┐
┌────┴─────┐        │  background  │
│Role Table│        │WAL Replicator│
└────▲─────┘        └──────┬───────┘
     │                     │
     └─────────────────────┘
          replicate log
            to table
```
A background replicate task is starting with `Instance`. `Instance` can register table to it with `register_table()`. The replicator will loop on the registered table to poll logs from WAL. Invalid table (those tables with incorrect states) will be removed from the queue.

There are two problems I can think of.
- A table may be kicked out in accident without notification.
- The entire replicate procedure inside a replicator is serialized. 

Implement the replicator described above.

No

# How does this change test

New added unit test. The add/evict table logics need a `RoleTable` trait definition, I'm going to cover this part after we got one.


analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
...
+                .await
+                .is_ok()
+            {
+                info!("WAL Replicator stopped");
+                break;
+            }
+        }
+    }
+
+    async fn consume_iter(
```
ShiKaiWi, line 230: Is it better if we call this method `consume_logs`?

analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
...
+
+        // update sequence number in state
+        replicate_state
+            .last_synced_seq
+            .fetch_max(max_seq, Ordering::Relaxed);
+
+        Ok(())
+    }
+
+    async fn replay_log(
```
ShiKaiWi, line 267: `replay_logs` seems better.

analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
...
+        &self,
+        log: &mut VecDeque<LogEntry<ReadPayload>>,
```
ShiKaiWi, line 269:
```suggestion
        logs: &mut VecDeque<LogEntry<ReadPayload>>,
```

analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
...
+        let join_handle = runtime.spawn(
+            self.inner
+                .clone()
+                .start_replicate(self.stop_receiver.take().unwrap()),
+        );
+        *self.join_handle.lock().await = Some(join_handle);
+    }
+}
+
+pub struct WalReplicatorInner {
```
ShiKaiWi, line 149: Shall we expose this struct? Maybe we call it `Inner` if it is private.

ShiKaiWi (approved):

analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+//! WAL Replicator implementation.
```
chunshao90, line 3: Rename `Replicator` to `Consumer`.

analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
...
+
+                // double check state before writing to table. Error due to
+                // state changed after this check will be treat as normal error.
+                if !state.check_state() {
+                    invalid_regions.push(state.region_id);
+                    continue;
+                }
+
+                // read logs from iterator
+                if let Err(e) = self.consume_logs(&mut iter, state).await {
```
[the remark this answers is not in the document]
waynexia: https://github.com/CeresDB/ceresdb/pull/179#discussion_r943123353

tired of changing naming again and again😣  leave it be if one name is style compatible and isn't ambiguous

analytic_engine/src/wal_replicator.rs
waynexia: Could you please describe why `Consumer` is better than `Replicator`? My reason for using `Replicator` is stated in #175. TL;DR: A replicator replicates logs from WAL to tables, it keeps the data between `Leader` table and its replicas in sync.

analytic_engine/src/wal_replicator.rs
```diff
@@ -0,0 +1,380 @@
...
+        let mut max_seq = SequenceNumber::MIN;
+
+        while should_continue {
+            // fetch entries
+            buf = iter
+                .next_log_entries(WalDecoder::default(), buf)
+                .await
+                .context(ReadWal)?;
+            if buf.len() <= self.config.batch_size {
+                should_continue = false;
```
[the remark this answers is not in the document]
waynexia: `buf` is and needs to be consumed at the end of each loop. I need a flag to record whether the loop can keep going based on the length of `buf` I polled. `should_continue` is that flag.

analytic_engine/src/wal_replicator.rs
chunshao90: `WAL` is divided into `log replicator` and `log consumer`. Currently, the log replicator is implemented by wal mod such as `WAL based on OBKV`, and followers only need to consume logs. In `start_replicate` func, actually calls `consume_log` not `replicate_log`.

waynexia: Or I can interpret `Consumer` as the one that eats logs and generate new data, which is `Table` in `Instance`. Your explanation might be right but I don't think it's better. And if so then we can keep it.

Furthermore in our model, distributed WAL component like `OBKV` is a place to store the write ahead logs. It doesn't "replicate" logs to other nodes neither by definition nor in practice. It's those followers that actually copy logs from storage to itself. IMHO calling this copy procedure "replicate" has no problem.

chunshao90: @ShiKaiWi  what you think?

ShiKaiWi: Here are my some thoughts about this naming issue:
- `Consumer` is actually more suitable for a more common case (only read logs without knowing how to apply these log contents to table).
- `Replicator` basically denotes makes a redundant copy of the data without changes about the original data, but here the data has been changed into `Table`'s data from the log entries.

So I propose a new name `Synchronizer`: synchronize logs from WAL to table.

waynexia: `Synchronizer` looks good to me. How about @chunshao90

ShiKaiWi (approved):
LGTM

Issue #77, opened 2022-06-30
[PoC] Hybrid storage format


jiacai2050:
**Description**

For now, data by default is ordered by `timestamp` column within one SST file(currently in Parquet format), each tag/field being a column. This design is good for OLAP queries, as it will only scan relevant columns, and CeresDB can take advantage of this ordering to filter unnecessary file, reducing IO further.

But for time-series user case like IoT or DevOps, this maybe not the best format. Those queries will typically first group its result by series id(or device-id), then by timestamp. This ordering isn't match with SST, so many random IOs will be incurred.

A general approach is to duplicate data twice: one ordered by timestamp first, and the other ordered by series id first.

Apparently this isn't very cost-effective, and will require some replication algorithm to synchronize data, which is very error-prone. It's best we could solve this `ordering` issue in one format.

**Proposal**

This issue propose one potential hybrid format (OLAP and time-series):

| Device ID | Timestamp           | Status Code | Tag 1 | Tag 2 |
|--------- |------------------- |----------- |----- |----- |
| A         | [12:01,12:02,12:03] | [0,0,0]     | v1    | v1    |
| B         | [12:01,12:02,12:03] | [0,1,0]     | v2    | v2    |
| &#x2026;  |                     |             |       |       |

In the above schema, instead of store timestamp row by row, we put timestamp within a device id in one array, and the corresponding values are also in array type, so we can easily map between them. The table is ordered by device ID.

In this way, we can avoid random IO when query one specific device, since its data are stored together. And this format isn't very bad for OLAP queries since we can maintain min/max time in row group's some metadata to help reader to filter unnecessary chunks.

**Additional context**

Some references
- [Building columnar compression in a row-oriented database](https://www.timescale.com/blog/building-columnar-compression-in-a-row-oriented-database/)

jiacai2050:
I have done a benchmark in my local env, This hybrid format is better than the old one.

Table below summarize read cost in each format(each is read ten times).

## Hybrid

| cost  | row nums |
|----- |-------- |
| 615ms | 10367743 |
| 576ms | 10367743 |
| 585ms | 10367743 |
| 511ms | 10367743 |
| 558ms | 10367743 |
| 569ms | 10367743 |
| 568ms | 10367743 |
| 555ms | 10367743 |
| 557ms | 10367743 |
| 584ms | 10367743 |


## Old

| cost   | row nums |
|------ |-------- |
| 1304ms | 10367743 |
| 1283ms | 10367743 |
[...]

PR #185, opened 2022-08-10
feat: write hybrid storage format

jiacai2050:

part of  #77 

# Rationale for this change
 

See issue above

- And a new table option `stroage_format`, read from env now
- Write record as hybrid format, read will be supported in future PR.

# How does this change test

Unit tests

### DIFF
--- a/analytic_engine/src/sst/builder.rs
+++ b/analytic_engine/src/sst/builder.rs
@@ -10,7 +10,7 @@ use crate::sst::file::SstMetaData;
 
 pub mod error {
     use common_util::define_result;
-    use snafu::Snafu;
+    use snafu::{Backtrace, Snafu};
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub))]
@@ -25,9 +25,14 @@ pub mod error {
             source: Box<dyn std::error::Error + Send + Sync>,
         },
 
-        #[snafu(display("Failed to encode record batch into sst, err:{}.", source,))]
+        #[snafu(display(
+            "Failed to encode record batch into sst, err:{}.\nBacktrace:\n{}",
+            source,
+            backtrace
+        ))]
         EncodeRecordBatch {
             source: Box<dyn std::error::Error + Send + Sync>,
+            backtrace: Backtrace,
         },
 
         #[snafu(display("Failed to poll record batch, err:{}", source))]
--- a/analytic_engine/src/sst/builder.rs
+++ b/analytic_engine/src/sst/builder.rs
@@ -35,6 +35,11 @@ pub mod error {
             backtrace: Backtrace,
         },
 
+        #[snafu(display("Invalid schema, err:{}", source))]
+        Schema {
+            source: Box<dyn std::error::Error + Send + Sync>,
+        },
+
         #[snafu(display("Failed to poll record batch, err:{}", source))]
         PollRecordBatch {
             source: Box<dyn std::error::Error + Send + Sync>,
--- a/analytic_engine/src/sst/builder.rs
+++ b/analytic_engine/src/sst/builder.rs
@@ -35,6 +35,26 @@ pub mod error {
             backtrace: Backtrace,
         },
 
+        #[snafu(display(
+            "Hybrid format don't support variable length type, type:{}.\nBacktrace:\n{}",
+            type_str,
+            backtrace
+        ))]
+        VariableLengthType {
+            type_str: String,
+            backtrace: Backtrace,
+        },
+
+        #[snafu(display(
+            "Not suppored arrow type. type:{}.\nBacktrace:\n{}",
+            type_str,
+            backtrace
+        ))]
+        NotSupportedArrowType {
+            type_str: String,
+            backtrace: Backtrace,
+        },
+
         #[snafu(display("Failed to poll record batch, err:{}", source))]
         PollRecordBatch {
             source: Box<dyn std::error::Error + Send + Sync>,
--- a/analytic_engine/src/sst/builder.rs
+++ b/analytic_engine/src/sst/builder.rs
@@ -45,16 +45,6 @@ pub mod error {
             backtrace: Backtrace,
         },
 
-        #[snafu(display(
-            "Not suppored arrow type. type:{}.\nBacktrace:\n{}",
-            type_str,
-            backtrace
-        ))]
-        NotSupportedArrowType {
-            type_str: String,
-            backtrace: Backtrace,
-        },
-
         #[snafu(display("Failed to poll record batch, err:{}", source))]
         PollRecordBatch {
             source: Box<dyn std::error::Error + Send + Sync>,
--- a/analytic_engine/src/sst/builder.rs
+++ b/analytic_engine/src/sst/builder.rs
@@ -35,16 +35,6 @@ pub mod error {
             backtrace: Backtrace,
         },
 
-        #[snafu(display(
-            "Hybrid format don't support variable length type, type:{}.\nBacktrace:\n{}",
-            type_str,
-            backtrace
-        ))]
-        VariableLengthType {
-            type_str: String,
-            backtrace: Backtrace,
-        },
-
         #[snafu(display("Failed to poll record batch, err:{}", source))]
         PollRecordBatch {
             source: Box<dyn std::error::Error + Send + Sync>,
--- a/analytic_engine/src/sst/builder.rs
+++ b/analytic_engine/src/sst/builder.rs
@@ -10,7 +10,7 @@ use crate::sst::file::SstMetaData;
 
 pub mod error {
     use common_util::define_result;
-    use snafu::{Backtrace, Snafu};
+    use snafu::Snafu;
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub))]
@@ -25,18 +25,8 @@ pub mod error {
             source: Box<dyn std::error::Error + Send + Sync>,
         },
 
-        #[snafu(display(
-            "Failed to encode record batch into sst, err:{}.\nBacktrace:\n{}",
-            source,
-            backtrace
-        ))]
+        #[snafu(display("Failed to encode record batch into sst, err:{}.", source,))]
         EncodeRecordBatch {
-            source: Box<dyn std::error::Error + Send + Sync>,
-            backtrace: Backtrace,
-        },
-
-        #[snafu(display("Invalid schema, err:{}", source))]
-        Schema {
             source: Box<dyn std::error::Error + Send + Sync>,
         },
 
--- a/analytic_engine/src/sst/parquet/encoding.rs
+++ b/analytic_engine/src/sst/parquet/encoding.rs
@@ -186,10 +186,10 @@ trait RecordEncoder {
     /// Encode vector of arrow batch, return encoded row number
     fn encode(&mut self, arrow_record_batch_vec: Vec<ArrowRecordBatch>) -> Result<usize>;
 
-    /// Return already encoded bytes
+    /// Return encoded bytes
     /// Note: trait method cannot receive `self`, so take a &mut self here to
     /// indicate this encoder is already consumed
-    fn into_bytes(&mut self) -> Result<Vec<u8>>;
+    fn close(&mut self) -> Result<Vec<u8>>;
 }
 
 /// EncodingWriter implements `Write` trait, useful when Writer need shared
@@ -263,7 +263,7 @@ impl RecordEncoder for ColumnarRecordEncoder {
         Ok(record_batch.num_rows())
     }
 
-    fn into_bytes(&mut self) -> Result<Vec<u8>> {
+    fn close(&mut self) -> Result<Vec<u8>> {
         assert!(self.arrow_writer.is_some());
 
         let arrow_writer = self.arrow_writer.take().unwrap();
@@ -324,7 +324,7 @@ impl HybridRecordEncoder {
             }
         }
 
-        let arrow_schema = hybrid::build_hybrid_arrow_schema(tsid_idx, &schema);
+        let arrow_schema = hybrid::build_hybrid_arrow_schema(tsid_idx, schema);
 
         let buf = EncodingWriter(Arc::new(Mutex::new(Vec::new())));
         let arrow_writer =
@@ -368,7 +368,7 @@ impl RecordEncoder for HybridRecordEncoder {
         Ok(record_batch.num_rows())
     }
 
-    fn into_bytes(&mut self) -> Result<Vec<u8>> {
+    fn close(&mut self) -> Result<Vec<u8>> {
         assert!(self.arrow_writer.is_some());
 
         let arrow_writer = self.arrow_writer.take().unwrap();
@@ -429,7 +429,7 @@ impl ParquetEncoder {
         self.record_encoder.encode(arrow_record_batch_vec)
     }
 
-    pub fn into_bytes(mut self) -> Result<Vec<u8>> {
-        self.record_encoder.into_bytes()
+    pub fn close(mut self) -> Result<Vec<u8>> {
+        self.record_encoder.close()
     }
 }
--- a/analytic_engine/src/sst/parquet/encoding.rs
+++ b/analytic_engine/src/sst/parquet/encoding.rs
@@ -16,6 +16,7 @@ use arrow_deps::{
 };
 use common_types::{
     bytes::{BytesMut, MemBufMut, Writer},
+    datum::DatumKind,
     schema::{ArrowSchemaRef, Schema, StorageFormat},
 };
 use common_util::define_result;
@@ -120,6 +121,16 @@ pub enum Error {
 
     #[snafu(display("Tsid is required for hybrid format.\nBacktrace:\n{}", backtrace))]
     TsidRequired { backtrace: Backtrace },
+
+    #[snafu(display(
+        "Key column must be string type. type:{}\nBacktrace:\n{}",
+        type_str,
+        backtrace
+    ))]
+    StringKeyColumnRequired {
+        type_str: String,
+        backtrace: Backtrace,
+    },
 }
 
 define_result!(Error);
@@ -283,8 +294,9 @@ struct HybridRecordEncoder {
     arrow_schema: ArrowSchemaRef,
     tsid_type: IndexedType,
     timestamp_type: IndexedType,
-    tag_types: Vec<IndexedType>,
-    field_types: Vec<IndexedType>,
+    key_types: Vec<IndexedType>,
+    // columns that can be collpased into list
+    non_key_types: Vec<IndexedType>,
 }
 
 impl HybridRecordEncoder {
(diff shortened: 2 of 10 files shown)


analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +173,93 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+pub struct ParquetEncoder<W: Write> {
+    writer: ArrowWriter<W>,
+    format: StorageFormat,
+    origin_arrow_schema: ArrowSchemaRef,
+    hybrid_arrow_schema: Option<ArrowSchemaRef>,
+    schema: Schema,
+}
+
+impl<W: Write> ParquetEncoder<W> {
+    pub fn try_new(
```
waynexia, line 186: should we check if the format is compatible with the table? E.g., some var-len types are not supported and we can check in advance here.

common_types/src/schema.rs
```diff
@@ -745,6 +750,21 @@ impl Schema {
     pub fn string_buffer_offset(&self) -> usize {
         self.column_schemas.string_buffer_offset
     }
+
+    /// Data format in storage
+    pub fn storage_format(&self) -> StorageFormat {
+        // TODO: parse it from table options
+        match std::env::var("CERESDB_TABLE_FORMAT") {
```
waynexia, line 757: parse from table option is more elegant. If we choose to use env we should document it.

common_types/src/schema.rs
```diff
@@ -445,6 +445,11 @@ pub fn compare_row<LR: RowView, RR: RowView>(
     Ordering::Equal
 }
 
+pub enum StorageFormat {
+    Columnar,
+    Hybrid,
+}
```
waynexia, line 451: I think we need a clear and concrete document about what these formats are

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,435 @@
...
+        hybrid_arrow_schema,
+        tsid_type,
+        timestamp_type,
+        tag_types,
+        field_types,
+        batch_by_tsid,
+    )
+}
+
+#[cfg(test)]
```
waynexia, line 377: The coverage looks not enough to me. Can we add more tests for this? Like round-trip (encode and decode) all supported types or Arrow arrays, and parqute-related tests

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,435 @@
...
+
+    let mut tag_types = Vec::new();
+    let mut field_types = Vec::new();
+    for (idx, col) in schema.columns().iter().enumerate() {
+        if col.is_tag {
+            tag_types.push(IndexedKind {
+                idx,
+                kind: schema.column(idx).data_type,
+            });
+        } else if idx != timestamp_type.idx && idx != tsid_type.idx {
```
chunshao90, line 294: We can impl is_field() in `ColumnSchema`. So we can use as
```
else if col.is_field() {
            field_types.push(IndexedKind {
                idx,
                kind: schema.column(idx).data_type,
            });
        }
```

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,435 @@
...
+            field_types.push(IndexedKind {
+                idx,
+                kind: schema.column(idx).data_type,
+            });
+        }
+    }
+    // TODO: should keep tsid ordering here?
```
chunshao90, line 301: Add newline above.

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,435 @@
...
+            }
+        })
+        .collect::<Vec<_>>();
+    Arc::new(ArrowSchema::new_with_metadata(
+        new_fields,
+        arrow_schema.metadata().clone(),
+    ))
+}
+
+fn merge_array_vec_to_list_array(
```
chunshao90, line 136: Split into some small fn. And add more comment to make  code more readable.

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,435 @@
...
+            }
+            length_so_far += *length as i32;
+            values.extend_from_slice(
+                &shared_buffer[offset * data_type_size..(offset + length) * data_type_size],
+            );
+        }
+        offsets.push(length_so_far);
+    }
+    debug!(
+        "merge_array_vec_to_list offsets:{:?},values:{:?}",
```
chunshao90, line 179: `"merge_array_vec_to_list offsets:{:?}, values:{:?}"`

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,435 @@
...
+        .collect::<BTreeMap<_, _>>()
+        .into_values()
+        .collect::<Vec<_>>();
+
+    ArrowRecordBatch::try_new(arrow_schema, all_columns)
+        .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send + Sync>)
+        .context(EncodeRecordBatch)
+}
+
+pub fn convert_to_hybrid(
```
chunshao90, line 271: Split into some small fn. And add more comment to make code more readable.

common_types/src/schema.rs
waynexia: Temporary is acceptable, but we need to doc this env someplace like PR's desc.

analytic_engine/src/sst/builder.rs
```diff
@@ -35,6 +35,16 @@ pub mod error {
             backtrace: Backtrace,
         },
 
+        #[snafu(display(
+            "Hybrid format don't support variable length type, type:{}.\nBacktrace:\n{}",
```
ShiKaiWi, line 39:
```suggestion
            "Hybrid format doesn't support variable length type, type:{}.\nBacktrace:\n{}",
```

analytic_engine/src/sst/builder.rs
```diff
@@ -35,6 +35,16 @@ pub mod error {
...
+            type_str,
+            backtrace
+        ))]
+        VariableLengthType {
+            type_str: String,
```
ShiKaiWi, line 44: How about `type_name`? The `str` suffix is verbose because we all know its type.

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -1,15 +1,31 @@
...
+    datum::DatumKind,
+    schema::{ArrowSchemaRef, Schema, StorageFormat},
+};
 use common_util::define_result;
 use proto::sst::SstMetaData as SstMetaDataPb;
 use protobuf::Message;
 use snafu::{ensure, Backtrace, OptionExt, ResultExt, Snafu};
 
-use crate::sst::file::SstMetaData;
+use super::hybrid::IndexedType;
```
ShiKaiWi, line 27: In favor of absolute import path.

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
 
     SstMetaData::try_from(meta_data_pb).context(ConvertSstMetaData)
 }
+
+/// RecordEncoder is used for encoding ArrowBatch
```
ShiKaiWi, line 193: Maybe we should make every statement end with `.`.

common_types/src/datum.rs
```diff
@@ -176,6 +176,28 @@ impl DatumKind {
     pub fn into_u8(self) -> u8 {
         self as u8
     }
+
+    /// Return None for variable-length type
+    pub fn size(&self) -> Option<usize> {
+        let size = match self {
+            DatumKind::Null => 1,
```
waynexia, line 183: https://arrow.apache.org/docs/format/Columnar.html#null-layout

Not sure but according to the spec NULL type's size is 0?

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let mut inner = self.0.lock().unwrap();
+        inner.flush()
+    }
+}
+
+struct ColumnarRecordEncoder {
+    buf: EncodingWriter,
+    // wrap in Option so ownership can be take out behind `&mut self`
```
ShiKaiWi, line 236:
```suggestion
    // wrap in Option so ownership can be taken out behind `&mut self`
```

common_types/src/datum.rs
```diff
@@ -755,6 +777,26 @@ pub mod arrow_convert {
                 | DataType::Map(_, _) => None,
             }
         }
+
+        pub fn to_arrow_data_type(&self) -> DataType {
+            match self {
+                DatumKind::Null => DataType::Null,
+                DatumKind::Timestamp => DataType::Timestamp(TimeUnit::Millisecond, None),
+                DatumKind::Double => DataType::Float64,
+                DatumKind::Float => DataType::Float32,
```
waynexia, line 786: Do we have `DataType::Float32`? I see the size of `Float` is also defined to 8 above

common_types/src/schema.rs
```diff
@@ -445,6 +445,41 @@ pub fn compare_row<LR: RowView, RR: RowView>(
...
+    /// Note: minTime/maxTime is optional and not implemented yet, mainly used
+    /// for time-range pushdown filter
+    ///
+    ///```plaintext
+    /// | Device ID | Timestamp           | Status Code | Tag 1 | Tag 2 | minTime | maxTime |
+    /// |-----------|---------------------|-------------|-------|-------|---------|---------|
+    /// | A         | [12:01,12:02,12:03] | [0,0,0]     | v1    | v1    | 12:01   | 12:03   |
+    /// | B         | [12:01,12:02,12:03] | [0,1,0]     | v2    | v2    | 12:01   | 12:03   |
+    /// | ...       |                     |             |       |       |         |         |
+    /// ```
```
waynexia, line 479: Nice illustration 👍

common_types/src/schema.rs
```diff
@@ -595,6 +630,16 @@ impl Schema {
         self.timestamp_index
     }
 
+    /// Whether i-nth column is key column
+    /// TODO: should tag column be key columns?
+    pub fn non_key_column(&self, i: usize) -> bool {
```
waynexia, line 635: Maybe `is_key_column` is more suitable for this kind of predictor.

And maybe we need to also explain what "key column" is in the comment

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+struct HybridRecordEncoder {
+    buf: EncodingWriter,
+    // wrap in Option so ownership can be take out behind `&mut self`
+    arrow_writer: Option<ArrowWriter<EncodingWriter>>,
+    arrow_schema: ArrowSchemaRef,
+    tsid_type: IndexedType,
+    timestamp_type: IndexedType,
+    key_types: Vec<IndexedType>,
+    // columns that can be collpased into list
+    non_key_types: Vec<IndexedType>,
```
ShiKaiWi, line 299: Maybe the counterpart to `key_types` is `normal_types` which is also used in the schema module: `key_columns` and `normal_columns`.

analytic_engine/src/sst/parquet/encoding.rs
waynexia, line 27: No relative import except test mod

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+impl Write for EncodingWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let mut inner = self.0.lock().unwrap();
+        inner.write(buf)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let mut inner = self.0.lock().unwrap();
+        inner.flush()
+    }
```
waynexia, line 231: `Vec` needn't flush. An empty function is enough

common_types/src/schema.rs
ShiKaiWi, line 635: Maybe what we need is a just a method called `is_key_column` and that works too by reverting the returned boolean when check whether one column is a non-key column.

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+        let tsid_type = IndexedType {
+            idx: tsid_idx,
+            data_type: schema.column(tsid_idx).data_type,
+        };
+
+        let mut key_types = Vec::new();
+        let mut non_key_types = Vec::new();
+        for (idx, col) in schema.columns().iter().enumerate() {
+            if schema.non_key_column(idx) {
+                let _ = col
```
ShiKaiWi, line 318: Maybe `ensure!` works better.

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+                    return StringKeyColumnRequired {
+                        type_str: col.data_type.to_string(),
+                    }
+                    .fail();
+                }
+                key_types.push(IndexedType {
+                    idx,
+                    data_type: col.data_type,
+                });
+            }
```
ShiKaiWi, line 342: I guess it will be better to organize this code block like this:
```rust
if idx == timetame_type.dix || idx == tsid_idx {
    continue
}
if schema.is_key_column(idx) {
...
} else {
...
}
```

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+                    })
+                    .map_err(|e| Box::new(e) as _)
+                    .context(EncodeRecordBatch)?;
+
+                non_key_types.push(IndexedType {
+                    idx,
+                    data_type: schema.column(idx).data_type,
+                });
+            } else if idx != timestamp_type.idx && idx != tsid_idx {
+                if !matches!(col.data_type, DatumKind::String) {
```
ShiKaiWi, line 332: Maybe `ensure!` works better.

analytic_engine/src/sst/parquet/reader.rs
```diff
@@ -332,13 +333,18 @@ impl ProjectAndFilterReader {
                 .context(DecodeRecordBatch)
             {
                 Ok(record_batch) => {
-                    row_num += record_batch.num_rows();
+                    let arrow_schema = record_batch.schema();
+                    let schema = Schema::try_from(arrow_schema).context(InvalidSchema)?;
+                    let record_batch = match schema.storage_format() {
+                        StorageFormat::Hybrid => todo!(),
```
ShiKaiWi, line 339: How about adding some messages about this `todo!()`.

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+    tsid_type: IndexedType,
+    timestamp_type: IndexedType,
+    key_types: Vec<IndexedType>,
+    // columns that can be collpased into list
+    non_key_types: Vec<IndexedType>,
+}
+
+impl HybridRecordEncoder {
+    fn try_new(write_props: WriterProperties, schema: &Schema) -> Result<Self> {
+        let tsid_idx = schema.index_of_tsid().context(TsidRequired)?;
```
waynexia, line 304: `tsid` is not required, we should support that situation (later)

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+        for (idx, col) in schema.columns().iter().enumerate() {
+            if schema.non_key_column(idx) {
+                let _ = col
+                    .data_type
+                    .size()
+                    .context(VariableLengthType {
+                        type_str: col.data_type.to_string(),
+                    })
+                    .map_err(|e| Box::new(e) as _)
+                    .context(EncodeRecordBatch)?;
```
waynexia, line 325: So tricky 🤣 takes seconds to think. Maybe a method like `is_type_supported()` would be better.

BTW the context `EncodeRecordBatch` can be more concrete, e.g. `UnsupportedType` or something

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+};
+use log::debug;
+use snafu::{OptionExt, ResultExt};
+
+use crate::sst::builder::{EncodeRecordBatch, Result, VariableLengthType};
+
+//  hard coded in https://github.com/apache/arrow-rs/blob/20.0.0/arrow/src/array/array_list.rs#L185
+const LIST_ITEM_NAME: &str = "item";
+
+/// ArrayHandle is used to keep different offsets of array, which can be concat
```
ShiKaiWi, line 25:
```suggestion
/// ArrayHandle is used to keep different offsets of array, which can be concated
```

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+/// together.
+///
+/// Note:
+/// 1. Array.slice(offset, length) don't work as expected, since the
+/// underlying buffer is still shared without slice.
+/// 2. Array shoule be [fixed-size primitive](https://arrow.apache.org/docs/format/Columnar.html#fixed-size-primitive-layout)
+#[derive(Debug, Clone)]
+struct ArrayHandle {
+    array: ArrayRef,
+    positions: Vec<(usize, usize)>, // (offset ,length)
```
ShiKaiWi, line 35: Is the `length` unnecessary (because it can be computed by `positions[idx+1] - positions[idx]`)?

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -150,3 +189,269 @@ pub fn decode_sst_meta_data(kv: &KeyValue) -> Result<SstMetaData> {
...
+                if !matches!(col.data_type, DatumKind::String) {
+                    return StringKeyColumnRequired {
+                        type_str: col.data_type.to_string(),
+                    }
+                    .fail();
+                }
+                key_types.push(IndexedType {
+                    idx,
+                    data_type: col.data_type,
+                });
```
waynexia, line 341: Maybe also leave a todo to support more type

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+            true,
+        ));
+        let array_data = ArrayData::builder(DataType::List(field))
+            .len(array_len)
+            .add_buffer(offsets.into())
+            .add_child_data(child_data);
+
+        // TODO: change to unsafe version?
+        // https://docs.rs/arrow/20.0.0/src/arrow/array/array_list.rs.html#192
+        // let array_data = unsafe { array_data.build_unchecked() };
```
ShiKaiWi, line 223: How about removing this comment? Basically, there is no need to consider usages of unsafe.

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+                } else {
+                    for i in 0..*length {
+                        bit_util::set_bit(null_slice, length_so_far as usize + i);
+                    }
+                }
+                length_so_far += *length as i32;
+                values.extend_from_slice(
+                    &shared_buffer[offset * data_type_size..(offset + length) * data_type_size],
+                );
+            }
```
ShiKaiWi, line 183: Can we concat the `null_bitmap` and `shared_buffer` directly to the `bitmap` and `values` rather than set them one by one?

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+                idx,
+                array: Arc::new(ListArrayBuilder::new(datum_type, handle).build()?),
+            })
+        })
+        .collect::<Result<Vec<_>>>()?;
+    let all_columns = vec![
+        vec![tsid_array, timestamp_array],
+        key_column_arrays,
+        non_key_column_arrays,
+    ]
```
ShiKaiWi, line 287:
```suggestion
    let all_columns = [
        vec![tsid_array, timestamp_array],
        key_column_arrays,
        non_key_column_arrays,
    ]
```

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+    .into_iter()
+    .flatten()
+    .map(|indexed_array| (indexed_array.idx, indexed_array.array))
+    .collect::<BTreeMap<_, _>>()
+    .into_values()
+    .collect::<Vec<_>>();
+
+    ArrowRecordBatch::try_new(arrow_schema, all_columns)
+        .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send + Sync>)
```
ShiKaiWi, line 296:
```suggestion
        .map_err(|e| Box::new(e) as _)
```

analytic_engine/src/sst/parquet/hybrid.rs
```diff
@@ -0,0 +1,441 @@
...
+
+/// Converts arrow record batch into hybrid record format describe in
+/// `StorageFormat::Hybrid`
+pub fn convert_to_hybrid_record(
+    tsid_type: &IndexedType,
+    timestamp_type: &IndexedType,
+    key_types: &[IndexedType],
+    non_key_types: &[IndexedType],
+    hybrid_arrow_schema: ArrowSchemaRef,
+    arrow_record_batch_vec: Vec<ArrowRecordBatch>,
```
ShiKaiWi, line 308:
```suggestion
    arrow_record_batches: Vec<ArrowRecordBatch>,
```

waynexia (approved):

waynexia:
🎉🎉

PR #186, opened 2022-08-10
docs: move architecture to user-guide

jiacai2050:

Closes #

# Rationale for this change
 
Architecture and user guide are in different places, I think it better to place them together, so users can browse them in https://docs.ceresdb.io/

- move architecture to user-guide directory in order to host via github pages
- add platform support under deploy section

Yes
image

# How does this change test

`mdbook serve` locally to check everything is expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,56 +13,11 @@ CeresDB is a high-performance, distributed, schema-less, cloud native time-serie
 ## Status
 The project is currently under rapid development. This early stage is not production ready and may incur data corruptions.
 
-## RoadMap
-
-See our [RoadMap](./docs/dev/roadmap.md)
 
 ## Documentation
-See https://docs.ceresdb.io/
-
-## Develop
-### Compile and run CeresDB
-#### Install dependencies
-In order to compile CeresDB, some relevant dependencies(including the `Rust` toolchain) should be installed.
-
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -64,24 +64,11 @@ Then you can run CeresDB using the default configuration file provided in the co
 ./target/release/ceresdb-server --config ./docs/example.toml
 ```
 
-## Platform Support
```
ShiKaiWi, line 67: The users may pay special attention to the supported platforms so how about keepping this in the README.md?

waynexia: unrelated issue: "platform" should not be part deploy section and should probably be at the top as well as "deploy"

waynexia: Good references. My point is that our "platform" is not for deployment purposes -- we only support one platform for now. That section is more about how to build and develop CeresDB. We can split "build" and "run" in https://github.com/CeresDB/ceresdb#platform-support and put "run" under "deploy" (maybe only one paragraph in "deploy" I guess).

waynexia: BTW, do you think we need a structured `dev-guide` like this `user-guide`? How about adding a top-level section. Then it will contain `Introduction`, `User Guide`, `Dev Guide`, and `Design`.

waynexia (approved):
Well done 👍

PR #188, opened 2022-08-11
feat: `RoleTable` trait and `LeaderTable` implementation

waynexia:

ref #175

# Rationale for this change
 

Following #175, this patch defines the trait `RoleTable` and implement three roles on top of it.

- Define `RoleTable`
- Draft `LeaderTable`
- Change some interfaces in `Instance`

no

# How does this change test

No, the logic and procedure are incomplete.


analytic_engine/src/instance/alter.rs
```diff
@@ -69,10 +84,11 @@ impl Instance {
     pub(crate) async fn process_alter_schema_command(
         self: &Arc<Self>,
         worker_local: &mut WorkerLocal,
-        space_table: &SpaceAndTable,
+        table_data: &TableDataRef,
         request: AlterSchemaRequest,
+        #[allow(unused_variables)] policy: TableAlterSchemaPolicy,
     ) -> Result<()> {
-        let table_data = space_table.table_data();
+        // let table_data = space_table.table_data();
```
ShiKaiWi, line 91: Why not just remove this line?

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -384,6 +394,11 @@ impl Instance {
         let _timer = local_metrics.flush_duration_histogram.start_timer();
 
         match policy {
+            TableFlushPolicy::Unknown => {
+                return Err(Error::UnknownPolicy {
+                    backtrace: Backtrace::generate(),
```
ShiKaiWi, line 399: I thought the backtrace will be generated automatically by snafu. Won't it?

analytic_engine/src/instance/write.rs
```diff
@@ -196,15 +212,18 @@ impl Instance {
     pub(crate) async fn process_write_table_command(
         self: &Arc<Self>,
         worker_local: &mut WorkerLocal,
-        space_table: &SpaceAndTable,
+        // space_table: &SpaceAndTable,
```
ShiKaiWi, line 215: Why not just remove this line?

analytic_engine/src/instance/write.rs
```diff
@@ -283,11 +302,14 @@ impl Instance {
...
+        space: &SpaceRef,
+        table_data: &TableDataRef,
         encode_ctx: &mut EncodeContext,
     ) -> Result<()> {
-        let space = space_table.space();
-        let table_data = space_table.table_data();
+        // let space = space_table.space();
+        // let space = table_data.sp
+        // let table_data = space_table.table_data();
```
ShiKaiWi, line 312: Why not just remove them?

analytic_engine/src/role_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+}
+
+struct LeaderTableInner {
+    state: AtomicU8,
+    table_data: TableDataRef,
+}
+
+// todo: handle `Result`
```
ShiKaiWi, line 103:
```suggestion
// TODO: handle `Result`
```

analytic_engine/src/role_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+#[allow(dead_code)]
+#[repr(u8)]
+pub enum TableRole {
+    Invalid = 0,
+    Leader = 1,
+    InSync = 2,
+    NoSync = 3,
+}
+
+pub struct LeaderTable {
```
ShiKaiWi, line 86: Shall we move the LeaderTable to a separate module?

analytic_engine/src/role_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    inner: Arc<LeaderTableInner>,
+}
+
+impl Drop for LeaderTable {
+    fn drop(&mut self) {
+        if Arc::strong_count(&self.inner) == 1 {
+            todo!("notify the state is completely changed")
```
ShiKaiWi, line 93: This will always cause panic, and maybe we should just leave a comment here instead of `todo!`?

analytic_engine/src/role_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+}
+
+struct LeaderTableInner {
```
ShiKaiWi, line 98: It seems that `LeaderTableInner` is the actual `LeaderTable` and the current `LeaderTable` should be called `LeaderTableRef` which focuses on reference counting.

analytic_engine/src/instance/alter.rs
```diff
@@ -29,6 +29,20 @@ use crate::{
     table_options,
 };
 
+/// Policy of how to perform flush operation.
+#[derive(Default, Debug, Clone, Copy)]
+pub enum TableAlterSchemaPolicy {
+    /// Unknown flush policy, this is the default value.
+    #[default]
+    Unknown,
```
ShiKaiWi, line 37: `Unknown` policy seems weird and unnecessary, and what's the difference between `Unknown` and `Noop`?

analytic_engine/src/role_table/mod.rs
waynexia: Not like other `T` & `TRef` pattern in our codebase, here `LeaderTableInner` is not public, others can only hold `LeaderTable`. And type alias `type LeaderTableRef = Arc<LeaderTable>` is not suitable since I customized the `Drop` fn for `LeaderTable`

analytic_engine/src/instance/alter.rs
waynexia: There are serval points
- I want each operation has its own policy, each policy name has meaning. E.g.: flush has `Purge` and `Dump` and alter has `Noop`.
- The only common policy variant is `Unknown`. And all operations report error when encount `Unknown` policy.
- The policy can only be set by `RoleTable`, other place can only set it to `Unknown` via `Default::default()`.

So the `Unknown` policy is a guard variant to avoid a operation has not checked by `RoleTable`. But the only problem I have is I cannot prevent others from constructing un-Default policy, maybe this need to be enforced via a runtime check.

analytic_engine/src/instance/flush_compaction.rs
waynexia: it would 👍

analytic_engine/src/instance/alter.rs
ShiKaiWi: Understood. How about giving some comments to the `UnKnown` Policy, especially for:
> And all operations report error when encount Unknown policy.

waynexia: > Understood. How about giving some comments to the `Unknown` Policy, especially for:
> 
> > And all operations report error when encount Unknown policy.

Ok 👌

ShiKaiWi (approved):

PR #189, opened 2022-08-11
refactor: try to replace dynamic-dispatch with pre-code in wal write

### LINKED ISSUE #162: Refactor write method in WalManager to improve performance
**Description**

After #159, we pass `dyn Payload` to write method of `WalManager`, this may has performance issue.
It is better to decouple encoder from write method. 

**Proposal**
We can define `WalEncoder`(maybe other name), like:
```rust
pub enum WalEncoder {
    Rocks(RocksLogEncoding),
    Obkv(ObkvLogEncoding)
}

impl WalEncoder {
   pub fn encode_to_write_batch<P:Payload>(&self, log_batch: &LogWriteBatch<P>) -> Result<(WriteBatch, u64)> {
       match self {
         WalEncoder::Rocks(encoder) => encoder.encode(payload),
         WalEncoder::Obkv(encoder) => encoder.encode(payload),
       }
   }
}
```

`Encoder` can get from `walManager`.
```rust
pub trait WalManager {
   fn encoder(&self) -> Encoder;
   fn write(&self, wb: WriteBatch) -> Result<()>,
}
```

So we need encode data before write to `WalManager`.
```rust
   let (wb, max_sequence_num) = self.wal.encoder().encode::<XXXPayload>(log_batch)?;
   self.wal.write(wb);
   ...
```

**Additional context**

cc: @waynexia @ShiKaiWi

ygf11:

Closes #162 

# Rationale for this change
`Pre-code` maybe more elegant and fast than `dynamic dispatch`.

- Using common `LogEncoding` for `rock_impl` and `table_kv_impl`.
- Add `WalManager::encoder()` to pre-encode payloads.
- Fix tests and benchmark. 

No

# How does this change test

Unit tests will test it.


ygf11:
I do a simple bench compared with the main branch. Since the result change a lot, I run 10 times for each, and compare time range.

Config:
```toml
bench_measurement_time = "60s"
bench_sample_size = 60
batch_size = 512
value_size = 1024
```

`Pre-Code`:
the time range is `720 µs ~ 840 µs`.

`main branch`:
the time range is `730 µs ~ 830 µs`.

Seems they are in the same level.

CC @waynexia

waynexia:
Sorry for the delay. With these results we can replace the late materialize approach confidently 🎊 I plan to review this today, appreciate your splendid work!

analytic_engine/src/meta/details.rs
```diff
@@ -311,17 +337,26 @@ impl MetaUpdateLogStore for RegionWal {
     }
 
     async fn store(&self, log_entries: &[MetaUpdateLogEntry]) -> Result<()> {
-        let mut log_batch = LogWriteBatch::new(self.region_id);
         let mut payload_batch = Vec::with_capacity(log_entries.len());
 
+        // TODO(ygf11): maybe we can build payload in encode loop.
```
waynexia, line 342: Do you mean encode payloads one by one? If so I prefer to encode them in a batch like now.

analytic_engine/src/meta/details.rs
```diff
@@ -311,17 +337,26 @@ impl MetaUpdateLogStore for RegionWal {
...
-            log_batch.push(LogWriteEntry { payload });
-        }
+        let region_id = self.region_id;
+        let log_batch_encoder = self
+            .wal_manager
+            .encoder(self.region_id, payload_batch.len() as u64)
+            .await
+            .context(GetLogBatchEncoder {
+                region_id,
+                entries_num: 1u64,
```
waynexia, line 355: `entries_num` should be `paylaod_batch.len()`

wal/src/kv_encoder.rs
```diff
@@ -463,3 +466,147 @@ impl MaxSeqMetaEncoding {
...
+        Self {
+            key_enc: LogKeyEncoder::newest(),
+            value_enc: LogValueEncoder::newest(),
+            value_enc_version: NEWEST_LOG_VALUE_ENCODING_VERSION,
+        }
+    }
+
+    // Encode [LogKey] into `buf` and caller should knows that the keys are ordered
+    // by ([RegionId], [SequenceNum]) so the caller can use this method to
+    // generate min/max key in specific scope(global or in some region).
```
waynexia, line 489:
```suggestion
    /// Encode [LogKey] into `buf` and caller should knows that the keys are ordered
    /// by ([RegionId], [SequenceNum]) so the caller can use this method to
    /// generate min/max key in specific scope(global or in some region).
```

waynexia (commented):
This is really great ❤️ I left some nitpicky comments

analytic_engine/src/meta/details.rs
ygf11: No, we still encode in a batch.

I am considering maybe we can pass `Into<Payload>` slice instead of `Payload` slice to the encode method, then the build loop can be eliminated.

current:
```rust
 pub fn encode<P: Payload>(self, payload_batch: &[P]) -> manager::Result<LogWriteBatch> {
 ... 
}
```

after:
```rust
pub fn encode<P: Payload, I:Into<Payload>>(self, payload_batch: &[I]) -> manager::Result<LogWriteBatch> {
 ... 
}
```

wal/src/manager.rs
```diff
@@ -255,10 +266,13 @@ pub trait WalManager: Send + Sync + fmt::Debug {
         req: &ReadRequest,
     ) -> Result<BatchLogIteratorAdapter>;
 
+    /// Provide the encoder for encoding payloads.
+    async fn encoder(&self, region_id: RegionId, entries_num: u64) -> Result<LogBatchEncoder>;
```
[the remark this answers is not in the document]
ygf11: Yes, it will brings some concurrent problems if thread model change.

This give me a new idea about `pre-code`:

- Because `SequenceNumber` is only used in `Encoding key`, maybe we can move encoding `key` to write method. Finally, it will never has concurrent problems, because `allocates SequenceNumber` and write is both in the same write lock.

Encode key:
```rust
// pub type LogKey = (RegionId, SequenceNumber);
pub fn encode_key(&self, buf: &mut BytesMut, log_key: &LogKey) -> Result<()>;
```

Encode Value:
```rust
// Never use SequenceNumber
pub fn encode_value(&self, buf: &mut BytesMut, payload: &impl Payload) -> Result<()>
```

Region in kv_table_impl:
```rust
pub struct Region {
    runtimes: WalRuntimes,
    state: RegionState,
    writer: Mutex<RegionWriter>,
}
```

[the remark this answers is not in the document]
waynexia: Sounds good. `LogKey` is fixed-length, payload-unrelated metadata, we don't have to encode both of them at the same time.

> because allocates SequenceNumber and write are in the protect of same mutex.

For now `SeqNum` is an atomic but yes, we can put "allocate `SeqNum`" and "consume(write) `SeqNum`)" together. The "mutex" is our thread model, let's leave it as it is.

analytic_engine/src/meta/details.rs
waynexia: Got it 👍

wal/src/manager.rs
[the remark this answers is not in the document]
ygf11: Addressed.

wal/src/kv_encoder.rs
ygf11: Addressed.

analytic_engine/src/meta/details.rs
ygf11: Since we do not allocate SequenceNumber in advance, entries_num is not in the Error now.

waynexia (approved):
Wonderful 🚀 Thanks @ygf11 !

PR #196, opened 2022-08-15
feat: support double quoted string.

### LINKED ISSUE #123: Support dummy select without table for string literal
**Description**

Support queries like
```sql
select "string literal";
```
Current the execution result is 
```
SELECT 'a';

Failed to execute query, err: Client("Failed to parse schema from JSON")
```

**Proposal**

Support this.

**Additional context**

ref #154

Rachelint:

Closes #123 

# Rationale for this change
Double quoted string is supported in Mysql and ceresdb plans to support such dialect.

+ Bump datafusion's version.
+ Add datum's conversion from double quoted string in ceresdb.

Users can use double quoted string like what they do in Mysql now.

# How does this change test
Test by integrated tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,8 +226,8 @@ name = "arrow_deps"
 version = "0.1.0"
 dependencies = [
  "arrow",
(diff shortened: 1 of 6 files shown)


waynexia (approved):
👍

Issue #92, opened 2022-07-07
Start CeresDB in distributed mode with CeresMeta


ShiKaiWi:
**Description**
The target is to run ceresdb in distributed mode and in order for that ceresdb in distributed mode, ceresdb must communicate with ceresmeta by the meta client. 

**Proposal**
CeresMeta's service is built by gRPC and the corresponding protocol has been defined by [the proto file](https://github.com/CeresDB/ceresdbproto/blob/main/protos/meta/service.proto) so first what we need to do is build a grpc client based on that proto file during ceresdb's starting procedure.

And there is no need to keep the old meta client, and the new meta client should provide the capabilities:
- Send keepalive heartbeats to ceresmeta;
- Register handlers for node heartbeats;
- Allocate schema/table id;
- Provide routing information of the cluster from the ceresmeta;

And based on the new meta client (meta_client_v2) a module called cluster will be responsible for managing information from the CeresMeta, including:
- Creating schema, creating/deleting tables;
- Tables to be opened in the ceresdb instance;
- Events such as change of the cluster topology from ceresmeta;
- ...

**Additional context**
steps:
- [x] Upgrade [ceresdbproto](https://github.com/CeresDB/ceresdbproto), remove old meta client and refactor meta_client_v2 (#100)
- [x] Enhance cluster module based on meta_client_v2 (#100)
- [x] Refactor volatile catalog implementation (#157)
- [x] Support startup in distributed mode (#190)
- [x] Remove the old meta_client (#205)
- [ ] Provide routing information (#215)

PR #205, opened 2022-08-18
refactor: remove old meta_client crate

ShiKaiWi:

Part of #92 

# Rationale for this change
 The old meta client is deprecated already, and should be replaced by the `meta_client_v2`. It is not trivial to remove it because the cluster topology and schema config is provided by the old meta client, and it is used across the whole server module.

- Remove the old meta client.
- Extend the capability of `meta_client_v2` to support to provide routing information.
- Refactor the `route` module and provides two implementations for `Router` one of which is based on the `cluster::Cluster` (but not implemented yet).
- Add one more trait `SchemaConfigProvider` used by the gRPC service which has two implementations: `ConfigBased` and `ClusterBased`.

The configuration about `meta_client` section will be removed.

# How does this change test
Unit test and integration test.


waynexia:
How about also removing the `_v2` in `meta_client_v2`

src/bin/ceresdb-server.rs
```diff
@@ -4,7 +4,7 @@
 
 // TODO(yingwen): ceresdb-server is a legacy name, maybe use a new name
```
waynexia, line 5: any thought?

cluster/src/lib.rs
```diff
@@ -44,6 +47,46 @@ pub type ClusterRef = Arc<dyn Cluster + Send + Sync>;
 
 pub type TableManipulatorRef = Arc<dyn TableManipulator + Send + Sync>;
 
+pub type ClusterTopologyRef = Arc<ClusterTopology>;
+
+pub type TableLocs = HashMap<String, ShardView>;
```
waynexia, line 52: what does `TableLocs` stand for

cluster/src/lib.rs
```diff
@@ -44,6 +47,46 @@ pub type ClusterRef = Arc<dyn Cluster + Send + Sync>;
...
+            auto_create_tables: false,
+            default_engine_type: ANALYTIC_ENGINE_TYPE.to_string(),
+            default_timestamp_column_name: TIMESTAMP_COLUMN.to_string(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Node {
+    pub addr: String,
```
waynexia, line 74: Would `SocketAddr` be better?

server/src/route/rule_based.rs
```diff
@@ -1,27 +1,40 @@
...
 const HASH_SEED: u64 = 0;
 
-pub type RouterRef = Arc<dyn Router + Sync + Send>;
+pub type ShardViews = HashMap<ShardId, ShardView>;
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct ShardView {
+    pub shard_id: ShardId,
+    pub node: Node,
+}
```
waynexia, line 32: This looks duplicated, other place defines `ShardView` too

server/src/config.rs
```diff
@@ -2,13 +2,23 @@
...
-use meta_client::MetaClientConfig;
+use cluster::{config::ClusterConfig, SchemaConfig};
+use common_types::schema::TIMESTAMP_COLUMN;
 use serde_derive::Deserialize;
+use table_engine::ANALYTIC_ENGINE_TYPE;
 
-use crate::router::RuleList;
+use crate::route::rule_based::{ClusterView, RuleList, ShardView};
 
+/// The deployment mode decides how to start up the CeresDB.
```
waynexia, line 15:
```suggestion
/// The deployment mode decides how to start the CeresDB.
```

"start up" is means another things

server/src/config.rs
```diff
@@ -28,6 +38,75 @@ pub struct RuntimeConfig {
...
+    pub auto_create_tables: bool,
+    pub default_engine_type: String,
+    pub default_timestamp_column_name: String,
+    pub shard_views: Vec<ShardView>,
+}
+
+impl Default for SchemaShardView {
+    fn default() -> Self {
+        Self {
+            schema: "".to_string(),
```
waynexia, line 61: would empty schema cause some issues? (I don't know, just confirm)

src/bin/ceresdb-server.rs
ShiKaiWi: It was actually `cse-server is legacy name, maybe use a new name` so I will remove this TODO.

cluster/src/lib.rs
ShiKaiWi: Actually, It stands for `TableLocations`, but let me figure out a better name. How about `TableShards`?

cluster/src/lib.rs
ShiKaiWi: `SocketAddr` introduces a limit that the `addr` must be an IP, but in some cases the `addr` may be a domain name.

server/src/route/rule_based.rs
ShiKaiWi: They are duplicated in name rather than content, so I will keep this intact and modify the `ShardView` in the cluster crate to `NodeShard`:
```rust
pub struct NodeShard {
    pub shard: ShardInfo,
    pub node: Node,
}
```

cluster/src/lib.rs
ShiKaiWi: The `ShardView` has been modified to:
```rust
pub struct NodeShard {
    pub shard: ShardInfo,
    pub node: Node,
}

pub struct TableNodeShards {
    pub table_id: TableId,
    pub node_shards: Vec<NodeShard>,
}
```


And the `TableLocs` should be `HashMap<TableName, TableNodeShards>`. Finally, this type alias is decided to be removed because I guess no need and no good name to alias such a type `HashMap<TableName, TableNodeShards>`.

server/src/config.rs
ShiKaiWi: It is just the default value for the `SchemaShardView` and the empty schema will be validated when used, so I think it is ok here.

server/src/route/rule_based.rs
waynexia:
```rust
pub struct ShardInfo {
    pub shard_id: ShardId,
    pub role: ShardRole,
}

pub enum ShardRole {
    LEADER,
    FOLLOWER,
}
```

But their contents are not that different I think 🧐 ?

cluster/src/lib.rs
waynexia: got it

cluster/src/lib.rs
waynexia: > Finally, this type alias is decided to be removed
👍

server/src/route/rule_based.rs
ShiKaiWi: It's true, but `rule_based.rs` needs such `ShardView` without `ShardRole` for configurations and `cluster` crate needs such `NodeShard` with `ShardRole` to describe the cluster topology. So the two similar but different structs have to exist at the same time.

ShiKaiWi: I have moved this `ShardView` to server/config.rs as just a config struct. And in the rule_based.rs, only `Node` is necessary.

waynexia (approved):
🌟

PR #208, opened 2022-08-19
feat: read hybrid format

jiacai2050:

part of #77 

# Rationale for this change
 
#185 implement write capability of hybrid format, this PR implement read

- Add `ParquetDecoder`, to convert record batch based on StorageFormat
  - In `HybridRecordDecoder`, stretch non-collapsible columns according to collapsible column's length
  - Extract ListArray's nested values out, and use it instead, so upper layer could see records just like normal columnar records 

No
# How does this change test

Add three unit tests in `encoding.rs`

### DIFF
--- a/analytic_engine/src/sst/parquet/encoding.rs
+++ b/analytic_engine/src/sst/parquet/encoding.rs
@@ -27,7 +27,7 @@ use common_types::{
     },
 };
 use common_util::define_result;
-use log::debug;
+use log::trace;
 use proto::sst::SstMetaData as SstMetaDataPb;
 use protobuf::Message;
 use snafu::{ensure, Backtrace, OptionExt, ResultExt, Snafu};
@@ -37,6 +37,7 @@ use crate::sst::{
     parquet::hybrid::{self, IndexedType},
 };
 
+// TODO: Only support i32 offset now, consider i64 here?
 const OFFSET_SIZE: usize = std::mem::size_of::<i32>();
 
 #[derive(Debug, Snafu)]
@@ -537,31 +538,28 @@ impl HybridRecordDecoder {
         array_ref: &ArrayRef,
         value_offsets: &[i32],
     ) -> Result<ArrayRef> {
-        assert!(!value_offsets.is_empty());
+        assert_eq!(array_ref.len() + 1, value_offsets.len());
 
         let values_num = *value_offsets.last().unwrap() as usize;
         let offset_slices = array_ref.data().buffers()[0].as_slice();
         let value_slices = array_ref.data().buffers()[1].as_slice();
         let null_bitmap = array_ref.data().null_bitmap();
-        debug!(
+        trace!(
             "raw buffer slice, offsets:{:#02x?}, values:{:#02x?}, bitmap:{:#02x?}",
             offset_slices,
             value_slices,
             null_bitmap.map(|v| v.buffer_ref().as_slice())
         );
 
-        let mut i32_offsets = Vec::with_capacity(offset_slices.len() / OFFSET_SIZE);
-        for i in (0..offset_slices.len()).step_by(OFFSET_SIZE) {
-            let offset = i32::from_le_bytes(offset_slices[i..i + OFFSET_SIZE].try_into().unwrap());
-            i32_offsets.push(offset);
-        }
-
+        let i32_offsets = Self::get_array_offsets(offset_slices);
         let mut value_bytes = 0;
         for (idx, (current, prev)) in i32_offsets[1..].iter().zip(&i32_offsets).enumerate() {
             let value_len = current - prev;
             let value_num = value_offsets[idx + 1] - value_offsets[idx];
             value_bytes += value_len * value_num;
         }
+
+        // construct new expanded array
         let mut new_offsets_buffer = MutableBuffer::new(OFFSET_SIZE * values_num);
         let mut new_values_buffer = MutableBuffer::new(value_bytes as usize);
         let mut new_null_buffer = MutableBuffer::new(values_num);
@@ -590,7 +588,7 @@ impl HybridRecordDecoder {
                 new_offsets_buffer.push(value_length_so_far);
             }
         }
-        debug!(
+        trace!(
             "new buffer slice, offsets:{:#02x?}, values:{:#02x?}, bitmap:{:#02x?}",
             new_offsets_buffer.as_slice(),
             new_values_buffer.as_slice(),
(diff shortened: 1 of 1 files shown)


analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+            SerializedFileReader::new(SliceableCursor::new(Arc::new(encoded_bytes))).unwrap();
+        let mut arrow_reader = ParquetFileArrowReader::new(Arc::new(reader));
+        let mut record_batch_reader = arrow_reader.get_record_reader(2048).unwrap();
+        let hybrid_record_batch = record_batch_reader.next().unwrap().unwrap();
+
+        let decoder = HybridRecordDecoder {};
+        let decoded_record_batch = decoder.decode(hybrid_record_batch).unwrap();
+        // Note: decode record batch's schema doesn't have metadata
+        // It's encoded in metadata of every fields
+        // assert_eq!(decoded_record_batch.schema(), input_record_batch.schema());
```
waynexia, line 940:
```suggestion
```

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+        Ok(array_data.into())
+    }
+
+    /// Returns offsets of nested items in ListArray, and populates the
+    /// `values_num`
+    ///
+    /// Note: caller should ensure `array_ref` is ListArray
+    fn get_list_array_offsets(array_ref: &ArrayRef) -> Vec<i32> {
+        let offset_slices = array_ref.data().buffers()[0].as_slice();
+        // *values_num = array_ref.data().child_data()[0].len();
```
waynexia, line 663:
```suggestion
```

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+            .len(values_num)
+            .build()
+            .map_err(|e| Box::new(e) as _)
+            .context(DecodeRecordBatch)?;
+
+        Ok(array_data.into())
+    }
+
+    /// Returns offsets of nested items in ListArray, and populates the
+    /// `values_num`
```
waynexia, line 658: `values_num` is commented out, is this comment outdated either?

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+    ///
+    /// Note: caller should ensure `array_ref` is ListArray
+    fn get_list_array_offsets(array_ref: &ArrayRef) -> Vec<i32> {
```
waynexia, line 661: Offset in arrow is a generic parameter. It can be `i32` or `i64`. The constant `OFFSET_SIZE` should also consider this

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+        let offset_slices = array_ref.data().buffers()[0].as_slice();
+        // *values_num = array_ref.data().child_data()[0].len();
+
+        let mut i32_offsets = Vec::with_capacity(offset_slices.len() / OFFSET_SIZE);
+        for i in (0..offset_slices.len()).step_by(OFFSET_SIZE) {
+            let offset = i32::from_le_bytes(offset_slices[i..i + OFFSET_SIZE].try_into().unwrap());
```
waynexia, line 667: Would it be more efficient to transmute directly if you use `unwrap()` here?

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+            .map_err(|e| Box::new(e) as _)
+            .context(DecodeRecordBatch)?;
+
+        Ok(array_data.into())
+    }
+
+    /// Returns offsets of nested items in ListArray, and populates the
+    /// `values_num`
+    ///
+    /// Note: caller should ensure `array_ref` is ListArray
```
waynexia, line 660: We'd better check this

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+
+impl RecordDecoder for HybridRecordDecoder {
+    /// Decode records from hybrid to columnar format
+    fn decode(&self, arrow_record_batch: ArrowRecordBatch) -> Result<ArrowRecordBatch> {
+        let new_arrow_schema = Self::convert_schema(arrow_record_batch.schema());
+        let arrays = arrow_record_batch.columns();
+        let mut value_offsets = None;
+        // Find value offsets from first `ListArray` column
+        for array_ref in arrays {
+            if matches!(array_ref.data_type(), DataType::List(_)) {
```
waynexia, line 683: The same question: what if it is a list array before hybrid encoding? I think we need to record which columns are collapsed rather than expand all those lists. (maybe we don't support creating a nested column for now, but I still think we cannot use `List` as a special array type)

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+                value_offsets = Some(Self::get_list_array_offsets(array_ref));
+                break;
+            }
+        }
+
+        ensure!(value_offsets.is_some(), ListArrayRequired {});
```
waynexia, line 689: Where does this constraint get introduced? I can't remember...

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+    ///
+    /// `array_ref` is `a b c`, `value_offsets` is `[0, 3, 5, 6]`, then
+    /// output array is `a a a b b c`
+    ///
+    /// Note: caller should ensure offsets is not empty.
+    fn stretch_variable_length_column(
+        array_ref: &ArrayRef,
+        value_offsets: &[i32],
+    ) -> Result<ArrayRef> {
+        assert!(!value_offsets.is_empty());
```
waynexia, line 540: This should be more strict, like: `array_ref.len() + 1 == value_offsets.len()`

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+
+        let values_num = *value_offsets.last().unwrap() as usize;
+        let offset_slices = array_ref.data().buffers()[0].as_slice();
+        let value_slices = array_ref.data().buffers()[1].as_slice();
+        let null_bitmap = array_ref.data().null_bitmap();
+        debug!(
```
waynexia, line 546: I think this log level is too high. This kind of detailed data should be avoided or at most at **trace** level. Another **debug** in this function is also included

analytic_engine/src/sst/parquet/encoding.rs
```diff
@@ -461,3 +487,457 @@ impl ParquetEncoder {
...
+            offset_slices,
+            value_slices,
+            null_bitmap.map(|v| v.buffer_ref().as_slice())
+        );
+
+        let mut i32_offsets = Vec::with_capacity(offset_slices.len() / OFFSET_SIZE);
+        for i in (0..offset_slices.len()).step_by(OFFSET_SIZE) {
+            let offset = i32::from_le_bytes(offset_slices[i..i + OFFSET_SIZE].try_into().unwrap());
+            i32_offsets.push(offset);
+        }
```
waynexia, line 557: We have a similar logic in `get_list_array_offsets()`. How about making `get_list_array_offsets` accept a slice and reuse it in both places?

analytic_engine/src/sst/parquet/encoding.rs
waynexia: Why can't without timestamp column? I can recall that the only required column is tsid, collapse-able and non-collapse-able columns are both optional, right?

waynexia (approved):

PR #212, opened 2022-08-23
fix: insert negative number value

### LINKED ISSUE #211: Fail to insert negative value
### Describe this problem

Now, CeresDB reports error when inserting a negative value (int or double).

### Steps to reproduce

Execute following sqls:
```sql
CREATE TABLE `test_negative_insert` (`t` timestamp not null, `v` int, TIMESTAMP KEY(t))

INSERT INTO test_negative_insert (t,  v) VALUES (1660233600000, -1);
```

Error message will be thrown:
```plaintext
err:Failed to create plan, query:INSERT INTO test_negative_insert (t,  v) VALUES (1660233600000, -1);, err:Failed to create plan, err:Invalid insert stmt, source expr is not value
```

### Expected behavior

Succeed in inserting negative values.

### Additional Information

_No response_

ShiKaiWi:

Closes #211

# Rationale for this change
 Now CeresDB reports failure when inserting negative number value and that is a bug needing fix. And the bug actually will lead to the failure of inserting any **explicitly signed** number.

Fix the parsing procedure from `Expr` to `Datum` to ensure numbers with sign are allowed to insert.

None.

# How does this change test
Add new UT tests and integration tests.

### DIFF
--- a/common_types/src/datum.rs
+++ b/common_types/src/datum.rs
@@ -485,6 +485,31 @@ impl Datum {
         }
     }
 
+    /// Generate a negative datum if possible.
+    ///
+    /// It will return `None` if:
+    /// - The data type has no negative value;
+    /// - The negative value overflows;
+    pub fn to_negative(self) -> Option<Self> {
+        match self {
+            Datum::Null => None,
+            Datum::Timestamp(_) => None,
+            Datum::Double(v) => Some(Datum::Double(-v)),
+            Datum::Float(v) => Some(Datum::Float(-v)),
+            Datum::Varbinary(_) => None,
+            Datum::String(_) => None,
+            Datum::UInt64(_) => None,
+            Datum::UInt32(_) => None,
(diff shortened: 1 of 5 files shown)


waynexia (approved):
The code is pretty and tests are concrete, LGTM 👍 

BTW, CI failed...

common_types/src/datum.rs
```diff
@@ -485,6 +485,31 @@ impl Datum {
         }
     }
 
+    /// Generate a negative datum if possible.
+    ///
+    /// It will return `None` if:
+    /// - The data type has no negative value;
+    /// - The negative value overflows;
```
chunshao90, line 492: Replace `;` wth `.`.

waynexia, line 492:
```suggestion
    /// - The data type has no negative value.
    /// - The negative value overflows.
```

PR #215, opened 2022-08-24
feat: support route in cluster mode

### LINKED ISSUE #92: Start CeresDB in distributed mode with CeresMeta

ShiKaiWi:

Closes #92 

# Rationale for this change
 The routing information when starting CeresDB with CeresMeta is stored in CeresMeta, and we need to provide such information to the clients by `route` grpc service.

Implement the router based on `Cluster`, and then provide the `route` gRPC service based on the router.

None because the `route` service is compatible.

# How does this change test
Some new unit tests to ensure this PR.

### DIFF
--- a/server/src/config.rs
+++ b/server/src/config.rs
@@ -5,7 +5,7 @@
 use std::collections::HashMap;
 
 use analytic_engine;
-use cluster::{config::ClusterConfig, Node, SchemaConfig};
+use cluster::{config::ClusterConfig, topology::SchemaConfig};
 use common_types::schema::TIMESTAMP_COLUMN;
 use meta_client::types::ShardId;
 use serde_derive::Deserialize;
@@ -44,6 +44,12 @@ pub struct RuntimeConfig {
 pub struct StaticRouteConfig {
     pub rule_list: RuleList,
     pub topology: StaticTopologyConfig,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Node {
+    pub addr: String,
+    pub port: u16,
 }
 
 #[derive(Debug, Clone, Deserialize)]
--- a/server/src/config.rs
+++ b/server/src/config.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 
 use analytic_engine;
 use ceresdbproto_deps::ceresdbproto::storage;
-use cluster::{config::ClusterConfig, topology::SchemaConfig};
+use cluster::config::{ClusterConfig, SchemaConfig};
 use common_types::schema::TIMESTAMP_COLUMN;
 use meta_client::types::ShardId;
 use serde_derive::Deserialize;
--- a/server/src/config.rs
+++ b/server/src/config.rs
@@ -47,22 +47,15 @@ pub struct StaticRouteConfig {
     pub topology: StaticTopologyConfig,
 }
 
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
 pub struct Endpoint {
     pub addr: String,
     pub port: u16,
 }
 
 impl Endpoint {
-    // Parse the raw endpoint which should be the form: <domain_name>:<port>
-    //
-    // Returns `None` if fail to parse.
-    pub fn try_from_str(s: &str) -> Option<Self> {
-        let (addr, raw_port) = s.rsplit_once(':')?;
-        raw_port.parse().ok().map(|port| Endpoint {
-            addr: addr.to_string(),
-            port,
-        })
+    pub fn new(addr: String, port: u16) -> Self {
+        Self { addr, port }
     }
 }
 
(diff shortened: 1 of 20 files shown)


server/src/schema_config_provider/cluster_based.rs
```diff
@@ -4,22 +4,28 @@
 
 #![allow(dead_code)]
```
waynexia, line 5: What will happen if we remove this

server/src/route/cluster_based.rs
```diff
@@ -19,8 +26,58 @@ impl ClusterBasedRouter {
     }
 }
 
+// Parse the raw endpoint which should be the form: <domain_name>:<port>
+//
+// Returns `None` if fail to parse.
+fn try_parse_endpoint(raw: &str) -> Option<Endpoint> {
```
waynexia, line 32: standard library has this trait https://doc.rust-lang.org/std/str/trait.FromStr.html#tymethod.from_str (which provides `.parse()`)

server/src/config.rs
```diff
@@ -46,6 +46,12 @@ pub struct StaticRouteConfig {
     pub topology: StaticTopologyConfig,
 }
 
+#[derive(Debug, Clone, Deserialize)]
+pub struct Node {
+    pub addr: String,
+    pub port: u16,
+}
```
waynexia, line 53: This is the same with `EndPoint`, do we have any way to reduce different naming on one thing?

server/src/schema_config_provider/cluster_based.rs
```diff
@@ -4,22 +4,28 @@
...
 
-use cluster::ClusterRef;
+use cluster::{topology::SchemaConfig, ClusterRef};
 
 use crate::schema_config_provider::{Result, SchemaConfigProvider};
 
 pub struct ClusterBasedProvider {
     cluster: ClusterRef,
+    default_schema_config: SchemaConfig,
```
waynexia, line 13: According to the `FIXME` beneath this `default_schema_config` is also temporary, right?

server/src/route/cluster_based.rs
```diff
@@ -19,8 +26,58 @@ impl ClusterBasedRouter {
...
+    let (domain, raw_port) = raw.split_once(':')?;
+    let port: u16 = raw_port.parse().map(Some).unwrap_or(None)?;
```
waynexia, line 34:
```suggestion
    let port: u16 = raw_port.parse().ok()?;
```

server/src/route/cluster_based.rs
```diff
@@ -19,8 +26,58 @@ impl ClusterBasedRouter {
...
+            .await
+            .map_err(|e| Box::new(e) as _)
+            .context(ErrWithCause {
+                code: StatusCode::InternalError,
+                msg: "Fail to route tables by cluster",
+            })?;
+
+        let mut routes = Vec::with_capacity(route_resp.entries.len());
+
+        // Now we pick up the nodes who own the leader shard for the route response.
```
waynexia, line 61: Is leader only here a hardcode? Maybe we need to add one field in `RouteRequest` to let the client specify what role it wants. Like `LeaderOnly`, `InSync`, `NoSync`, `DontCare` etc.

server/src/config.rs
ShiKaiWi: `Endpoint` is actually defined in `ceresdbproto` so we can't use it as some item of the configuration.

server/src/route/cluster_based.rs
ShiKaiWi: It's true, but for now the protocol hasn't supported such things yet. Let's add the information about roles afterwards when the protocol upgraded.

server/src/schema_config_provider/cluster_based.rs
ShiKaiWi: Yes. Now the schema information is not supported to be stored by CeresMeta. The schema configurations will be fetched from `Cluster` in the future.

server/src/route/cluster_based.rs
ShiKaiWi: I have tried, but I find that the Endpoint is defined in the ceresdbproto crate and I can't impl `FromStr` for it.

ShiKaiWi: I have tried, but I find that the Endpoint is defined in the ceresdbproto crate and I can't impl `FromStr` for it.

ShiKaiWi: I have tried, but I find that the Endpoint is defined in the ceresdbproto crate and I can't impl `FromStr` for it.

ShiKaiWi: I have tried, but I find that the Endpoint is defined in the ceresdbproto crate and I can't impl `FromStr` for it.

waynexia: I have a suggestion: redefine `Endpoint` in our lib and operate on it, and implement methods like `to_pb()` and `from_pb()`. By doing this https://github.com/CeresDB/ceresdb/pull/215#discussion_r954466140 can also be solved.

waynexia (approved):

PR #217, opened 2022-08-25
fix: deadlock when reconnect heartbeat channel

ShiKaiWi:

Closes #

# Rationale for this change
 Deadlock occurs when fail to send heartbeat stream and reconnect the gRPC server. Actually, `reconnect_hearbeat_channel` will take the write lock on the `grpc_client` but it is very implicit.

- Avoid taking the write lock on the `grpc_client` in the `reconnect_heartbeat_channel`, and let the caller to update the `grpc_client` manually.
- Remove topology cache update procedure because it can't handle the updates.

None.

# How does this change test
Test by manually.

### DIFF
--- a/cluster/src/cluster_impl.rs
+++ b/cluster/src/cluster_impl.rs
@@ -1,19 +1,15 @@
 // Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
 
 use std::{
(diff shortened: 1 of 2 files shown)


chunshao90 (approved):
LGTM

PR #222, opened 2022-08-29
docs: improve options section

jiacai2050:

Part of https://github.com/CeresDB/ceresdb/issues/77#issuecomment-1217376966

# Rationale for this change
 

The hybrid format has been merged in main, but no docs.

improve options section

Yes, all users can refer this doc to create hybrid table.

# How does this change test

`mdbook serve` locally to check page style is fine.

### DIFF
--- a/docs/guides/src/analytic_engine/README.md
+++ b/docs/guides/src/analytic_engine/README.md
@@ -1,1 +1,3 @@
 # Analytic Engine
+
+The engine mainly used for OLAP scenarios.
--- a/docs/guides/src/analytic_engine/README.md
+++ b/docs/guides/src/analytic_engine/README.md
@@ -1,3 +1,1 @@
 # Analytic Engine
-
-The engine mainly used for OLAP scenarios.
--- a/docs/guides/src/analytic_engine/options.md
+++ b/docs/guides/src/analytic_engine/options.md
@@ -1,1 +1,49 @@
 # Options
+
+Options below can be used when create table for analytic engine
+
+- `enable_ttl`, `bool`. When enable TTL on a table, rows older than `ttl` will be deleted and can't be querid, default `true`
+- `ttl`, `duration`, lifetime of a row, only used when `enable_ttl` is `true`. default `7d`.
+- `storage_format`, `string`. The underlying column's format. Availiable values:
+  - `columnar`, default
+  - `hybrid`
+
+  The meaning of those two values are in [Storage format](#storage-format) section.
+
+
+## Storage Format
+
+There are mainly two formats supported in analytic engine. One is `columnar`, which is the traditional columnar format, with one table column in one physical column:
+
+```plaintext
+| Timestamp | Device ID | Status Code | Tag 1 | Tag 2 |
+| --------- |---------- | ----------- | ----- | ----- |
+| 12:01     | A         | 0           | v1    | v1    |
+| 12:01     | B         | 0           | v2    | v2    |
+| 12:02     | A         | 0           | v1    | v1    |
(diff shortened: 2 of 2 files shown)


docs/guides/src/analytic_engine/options.md
```diff
@@ -1 +1,91 @@
...
+| 12:01     | A         | 0           | v1    | v1    |
+| 12:01     | B         | 0           | v2    | v2    |
+| 12:02     | A         | 0           | v1    | v1    |
+| 12:02     | B         | 1           | v2    | v2    |
+| 12:03     | A         | 0           | v1    | v1    |
+| 12:03     | B         | 0           | v2    | v2    |
+| .....     |           |             |       |       |
+```
+
+The other one is `hybrid`, a exprimental format used to simulate a row-oriented storage in columnar storage, in order to accelerate OLTP queries.
```
waynexia, line 30: We don't support "OLTP"

docs/guides/src/analytic_engine/README.md
```diff
@@ -1 +1,3 @@
 # Analytic Engine
+
+The engine mainly used for OLAP scenarios.
```
waynexia, line 3:
```suggestion
```

I prefer to not say this unless we have another engine.

docs/guides/src/analytic_engine/options.md
waynexia, line 30:
```suggestion
The other one is `hybrid`, an experimental format used to simulate row-oriented storage in columnar storage to accelerate OLTP queries.
```

docs/guides/src/analytic_engine/options.md
```diff
@@ -1 +1,91 @@
...
+
+In traditional time-series user case like IoT or DevOps, queries will typically first group its result by series id(or device-id), then by timestamp. In order to archieve good performance in those scenarios, the data physical layout should match this style, so the `hybrid` format is proposed like this:
```
waynexia, line 32:
```suggestion
In traditional time-series user cases like IoT or DevOps, queries will typically first group their result by series id(or device-id), then by timestamp. In order to achieve good performance in those scenarios, the data physical layout should match this style, so the `hybrid` format is proposed like this:
```

docs/guides/src/analytic_engine/options.md
```diff
@@ -1 +1,91 @@
...
+
+
+- With in one file, rows belong to same primary key(eg: series/device id) are collapsed in one row
+- The columns besides primary key are divided in two categroies:
+  - `collapsible`, those columns will be collapsed into a list. Used to encode `fields` in time-series table
+    - Note: only fixed-length type is supported now
+  - `non-collapsible`, those columns should only contains one distinct value. Used to encode `tags` in time-series table
+    - Note: only string type is supported now
+- Two more columns are added, `minTime` and `maxTime`. Those are used to cut unnecessary rows out in query.
+  - Note: Not implemented yet.
```
waynexia, line 51:
```suggestion
- Within one file, rows belonging to the same primary key(eg: series/device id) are collapsed into one row
- The columns besides primary key are divided into two categories:
  - `collapsible`, those columns will be collapsed into a list. Used to encode `fields` in time-series table
    - Note: only fixed-length type is supported now
  - `non-collapsible`, those columns should only contain one distinct value. Used to encode `tags` in time-series table
    - Note: only string type is supported now
- Two more columns are added, `minTime` and `maxTime`. Those are used to cut unnecessary rows out in query.
  - Note: Not implemented yet.
```

waynexia (approved):

PR #204, opened 2022-08-18
chore: check lock file in CI

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

# Which issue does this PR close?

Closes #52

# Rationale for this change
 
We track `Cargo.lock` file with SCM but it's not checked in CI. It happens serval times that someone updates `.toml` and forgets to commit changes in `.lock`. Those ignored changes will then be submitted in another unrelated PR by another innocent.

# What changes are included in this PR?

Add a CI job to check if the lock file is up-to-date

# Are there any user-facing changes?

No

# How does this change test

CI itself

PR #223, opened 2022-08-29
chore: check Cargo.lock in CI

### LINKED ISSUE #52: CI: check diff on `Cargo.lock`
**Description**
Add CI job to check file diff of `Cargo.lock`.

We are tracking `Cargo.lock` with SCM, and should also check it in CI. An incorrect (usually is just not up-to-date) lock file won't fail a build but is annoying.

**Proposal**

diff lock file after `cargo build`. Fail if there is any diff.

**Additional context**

jiacai2050:

Closes #52

# Rationale for this change
 

Supersede #204 , to check if Cargo.lock has been changed all by ourselves, without any other dependencies.

No

# How does this change test

Testing with those command
```bash
          cp ${LOCK_FILE} ${LOCK_FILE}.bak
          git clone https://github.com/apache/parquet-testing.git components/parquet-testing
          make test-ut
          echo "# test" >> ${LOCK_FILE}
          diff ${LOCK_FILE} ${LOCK_FILE}.bak
```
It will fail https://github.com/jiacai2050/ceresdb/runs/8065062616?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,8 +98,12 @@ jobs:
           sudo apt update
           sudo apt install --yes gcc g++ libssl-dev pkg-config cmake git
       - name: Run Tests
+        env:
+          LOCK_FILE: Cargo.lock
         run: |
+          cp ${LOCK_FILE} ${LOCK_FILE}.bak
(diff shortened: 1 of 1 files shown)


waynexia (approved):
👍

PR #225, opened 2022-08-29
chore: add minimal config

jiacai2050:

Closes #

# Rationale for this change
 
Local is useful to test locally

# How does this change test

Start server locally with minimal.toml


waynexia:
can we add another `local-config.toml` in somewhere else like `etc/`? This one under `docs/` is used as the config file's example.

chunshao90:
remove `[analytic.storage.remote_store]`.

waynexia (approved):

Issue #235, opened 2022-09-06
Refactor Cluster and Catalog mods


waynexia:
### Describe This Problem

A rough dependency graph of related structures:

![cluster](https://user-images.githubusercontent.com/15380403/188592338-48dc88dd-a484-4352-94d6-7d139030bdca.svg)

`Cluster` and `CatalogManager` are public, they may be used by `Client` or `CeresMeta` to modify catalog information (create, open or delete tables). This graph shows two paths of creating/droping a table. The **green** one is triggered by server itself and another **orange** one is triggered by `CeresMeta`

- Green path
  `CatalogManager` can create/drop tables when it has write access (the leader role). `3-1` in this path is notifying `CeresMeta` that a new table has been created, and `3-2` is an operation that happens simultaneously that create/drop that table in the leader itself. 
- Orange path
  This path can be treated as counter part of the green one. `CeresMeta` may notify non-leader node that something has changed. It also goes through `EventHandler` to notify `Cluster` and let it take action.

### Proposal

From my first impression, the dependency and call graph is complicated. I'd like to change relative parts in the following ways:
- [x] ~~Combine `VolatileCatalog` and `Cluster` #245
  `Cluster` is responsible to manage `Table`s and `Shard`s while `Catalog` is responsible to manage `Schema`s which is a group of tables in another perspective different from `Shard`. Their functionalities have a big part that overlaps. It's natural to implement both `Cluster` and `Catalog` traits over one struct, so they can use one memory state and maybe one operation logic.~~
- [x] Remove `SchemaIdAlloc` and `TableIdAlloc` #238
  `Id` should be allocated by `CeresMeta`, not the server. And furthermore, the server does need not to allocate ID then create table. The entire procedure is accomplished by `CeresMeta`.
- [x] Split `MetaClient`
  Current `MetaClient` provides two functionalities: poll update from `CeresMeta` and invoke `CeresMeta`'s method. These correspond to the two ways above: green arrows are from server to `CeresMeta` and orange are from `CeresMeta` to server. We can split these two aspects, by making `MetaClient` an actual client that invokes RPC to `CeresMeta` and an `Eventloop` that keeps fetching updates from `CeresMeta`. This can also resolve the cyclic dependency between `Cluster` and `MetaClient`.

Due to https://github.com/CeresDB/ceresdb/issues/235#issuecomment-1275722731, here are the new tracking tasks:
- [x] Remove the current catalog based on cluster, and add table manager to volatile catalog; #295
[...]

PR #238, opened 2022-09-07
refactor: remove `SchemaIdAlloc` and `TableIdAlloc`

waynexia:

ref #235 

# Rationale for this change
 

>Id should be allocated by CeresMeta, not the server. And furthermore, the server does need not to allocate ID then create table. The entire procedure is accomplished by CeresMeta.

- Remove `TableIdAlloc` and `SchemaIdAlloc` traits' definition and impls
- volatile catalog impl uses `MetaClient` directly
- serval minor changes on comments

no

# How does this change test

No new case is added


ShiKaiWi (approved):
Good Job！

PR #248, opened 2022-09-14
feat: add scan batch size options

jiacai2050:

Closes #

# Rationale for this change
 
For now, iter scan size is hard coded, add an option to configure it.

# How does this change test

Start server with this config
```
[analytic]
wal_path = "/path/to/data"
sst_data_cache_cap = 10000
sst_meta_cache_cap = 10000
scan_batch_size = 10000
```


ShiKaiWi (approved):
LGTM

PR #254, opened 2022-09-14
feat: primary key definition support specify tsid column

jiacai2050:

Closes #

# Rationale for this change
 

Since `TSID` is a reserved column, we can't define a table with a tsid column, `TSID` is used when create table without setting primary key, but in this case the primary key is hard coded to `timestamp, tsid`, there is no way to change this order now.

1. When `create_table_to_plan`, if `TSID` is in primary key definition, skip existence check
2. Adjust some log level to info, since those are useful to analyze query performance.

No

# How does this change test

```
CREATE TABLE test(ts timestamp timestamp key not null, c1 string tag, c2 int, primary key (tsid,ts))  ENGINE=Analytic;
show create table test;
```
Output
```
{
  "rows": [
    {
      "Table": "test",
      "Create Table": "CREATE TABLE `test` (`tsid` uint64 NOT NULL, `ts` timestamp NOT NULL, `c1` string TAG, `c2` int, PRIMARY KEY(tsid,ts), TIMESTAMP KEY(ts)) ENGINE=Analytic WITH(arena_block_size='2097152', compaction_strategy='default', compression='UNCOMPRESSED', enable_ttl='true', num_rows_per_row_group='8192', segment_duration='', storage_format='COLUMNAR', ttl='7d', update_mode='OVERWRITE', write_buffer_size='3221225472')"
    }
  ]
}
```

### DIFF
--- a/query_engine/src/executor.rs
+++ b/query_engine/src/executor.rs
@@ -2,10 +2,11 @@
 
 //! Query executor
 
-use std::sync::Arc;
+use std::{sync::Arc, time::Instant};
 
 use async_trait::async_trait;
 use common_types::record_batch::RecordBatch;
+use common_util::time::InstantExt;
 use futures::TryStreamExt;
 use log::{debug, info};
 use snafu::{ResultExt, Snafu};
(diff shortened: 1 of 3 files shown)


ShiKaiWi (approved):
LGTM

PR #264, opened 2022-09-20
feat: add tsbs CI

jiacai2050:

Closes #

# Rationale for this change
 
We currently use [tsbs](https://github.com/CeresDB/tsbs) to benchmark CeresDB, it would be convenient to run it in CI

- Add a new CI, triggered by manually or 00:02 UTC every day
  - After benchmark, it will output artifacts like [bench-2022-09-21T09_59_31](https://github.com/CeresDB/ceresdb/suites/8389523337/artifacts/370511312), which contains server.log for trouble shooting
- If cached dependencies is hit, the CI will take 20min approximately.

No

# How does this change test

- https://github.com/jiacai2050/ceresdb/actions/runs/3090091094
- TSBS CI itself

### DIFF
--- a/.github/workflows/tsbs.yml
+++ b/.github/workflows/tsbs.yml
@@ -35,6 +35,9 @@ jobs:
         run: |
           ./scripts/run-tsbs.sh
           echo "NOW=$(TZ=':Asia/Shanghai' date +'%Y-%m-%dT%H_%M_%S')" >> $GITHUB_ENV
+      - name: Update Summary
+        run: |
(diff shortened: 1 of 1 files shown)


waynexia (approved):
Splendid 🤩 

How about adding one more change that renders the `result.md` in the action's result page, like here https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/

From the post this should be done by adding
```yaml
steps:
  - name: Adding markdown
    run: cat tsbs/result.md >> $GITHUB_STEP_SUMMARY 
```

PR #260, opened 2022-09-16
refactor: use TableManager to implement `CatalogManager` and `SchemaManager`

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

# Which issue does this PR close?

Part of #245.

# Rationale for this change
 

`CatalogManager` and `SchemaManager`'s interfaces are redundant in some sorts. This is one step towards `Cluster` and `catalog`'s combination. The next step should be to implement table manipulating operations on top of this.

# What changes are included in this PR?

This patch implements `CatalogManager` and `SchemaManager` using `TableManager`, and implement `catalog::manager::Manager` for `ClusterImpl`

The new `catalog::manager::Manager` implementation should replace `volatile`, but I haven't removed the old one to keep this patch small.

# Are there any user-facing changes?

No so far.

# How does this change test

untested.

Issue #245, opened 2022-09-13
Combine `VolatileCatalog` and `Cluster`


waynexia:
### Describe This Problem

Reference to #235:
> `Cluster` is responsible to manage `Table`s and `Shard`s while `Catalog` is responsible to manage `Schema`s which is a group of tables in another perspective different from `Shard`. Their functionalities have a big part that overlaps. It's natural to implement both `Cluster` and `Catalog` traits over one struct, so they can use one memory state and maybe one operation logic.

### Proposal

- [ ] Make a triple to act as `TableRef`'s full identifier(token)
  ```rust
  struct TableToken {
      catalog: CatalogId,
      schema: SchemaId,
      table: TableId,
  }
  ```
  We currently use the three layers separately: first get a `Catalog` instance, and get `Schema` instance from it, then get `Table` instance. And the storage needs a nested map struct to put all this infos. The entire procedure can be simplified.

- [ ] Remove `CatalogManager` and `SchemaManager` to reduce the complexity and simplify interfaces. Now only keep a top-level `Manager` and low-level `TableManager` 

  Follows the above description. `CatalogManager` and `SchemaManager`'s interfaces are redundant in some sorts.

- [ ] Manage all the `TableRef` in one place, here is `TableManager` insides `Cluster`

  The one who keeps all `TableRef` should be able to serve two requests:
    - get tables by shard
    - get tables by catalog/schema/table
  
  the fields in `TableManager` may look like
  ```rust
  struct TableManager {
      // ...
      tables: BTreeMap<TableToken, TableRef>,
      shards: HashMap<ShardId, HashSet<TableToken>>,
  }
  ```

### Additional Context

- [ ] (optional) Unify the name and id. Like `SchemaName` and `SchemaId`
```rust
enum CatalogToken{
    Id(CatalogId),
    Name(String)
}
```

PR #265, opened 2022-09-21
feat: Implement some cluster commands

waynexia:

Part of #245

# Rationale for this change
 

The `TableManager` is changed in #260, some command implementations need to be re-implement.

- Remove `TableManipulator`.
  - It doesn't provide full functionalities. Only open and close are not enough
  - It's just another wrapper over `TableEngine`, which is unnecessary.
- Draft `Open`, `Close`, `Create` and `Drop` commands. But they are going to be changed again in the later task of #245 (with new `MetaClient` and `MetaEndpoint`)

No for now

# How does this change test

untested


cluster/src/cluster_impl.rs
```diff
@@ -134,22 +140,81 @@ impl EventHandler for Inner {
...
+
+                self.table_engine.drop_table(req).await?;
+
+                Ok(())
+            }
+            ActionCmd::MetaNoneCmd(_) => Ok(()),
+            ActionCmd::MetaCloseCmd(close_cmd) => {
+                let mut execute_errors = vec![];
+
+                for shard in &close_cmd.shard_ids {
```
ShiKaiWi, line 192:
```suggestion
                for shard_id in &close_cmd.shard_ids {
```

cluster/src/cluster_impl.rs
```diff
@@ -134,22 +140,81 @@ impl EventHandler for Inner {
...
+                    let tokens = self.table_manager.tokens_by_shard(*shard);
+                    for token in tokens {
+                        let req = CloseTableRequest {
+                            catalog_name: DEFAULT_CATALOG.to_string(),
+                            schema_name: "TODO".to_string(),
```
ShiKaiWi, line 197: Maybe we should use empty string and add TODO comment : Remove `schema_name` and `table_name` from `CloseTableRequest` in the future.

ShiKaiWi (approved):
LGTM

Issue #258, opened 2022-09-16
Integration tests are not stable


ygf11:
### Describe this problem

For create table test, 

https://github.com/CeresDB/ceresdb/blob/25557c73adaa6815875aaea3ccd5b61bf54e291f/tests/cases/local/05_ddl/create_tables.sql#L64

it will report error like:
```
Failed to execute query, err: Server(ServerError { code: 500, msg: "Failed to execute interpreter, query: CREATE TABLE `05_create_tables_t8`(c1 int, t1 timestamp NOT NULL TIMESTAMP KEY) ENGINE = Analytic with (storage_format= 'unknown');. Caused by: Failed to execute create table, err:Failed to create table, name:05_create_tables_t8, err:Failed to create table, err:Invalid arguments, err:Invalid options, space_id:2, table:05_create_tables_t8, table_id:2199023255725, err:Unknown storage format. value:\"unknown\"." })
```

The `table_id` in error message is not stable in my local pc, it will change if I run tests again.

The issue maybe is in test config, we use `Local storage` as the server config for integration tests now.

cc: @waynexia @jiacai2050

### Steps to reproduce

Run tests again.

### Expected behavior

The output should the same as before when we rerun tests.

ygf11:
> For your case, you can ignore this error for now, we will support special interceptor to replace those random string in near future(maybe one month I guess).

Thanks, got it.

ShiKaiWi:
#270 will fix this. The reason for this bug is that the test data won't be cleaned before any run, while the table id is auto-incremented, so the created, dropped, created-again table may have different table ids every time run the tests.

ShiKaiWi:
> I think of one corner case even when #270 is merged
> 
> > if users drop a new table in `00_dummy`, then all table id after `00_dummy` will be affected
> 
> So an interceptor may also need to deal with this case.

Actually, just update the result files if the test cases are modified.

ShiKaiWi:
I guess #270 has fixed this, so let's close it.

PR #270, opened 2022-09-26
test: make harness test stable

### LINKED ISSUE #107: Implement `ignore` interceptor in integration test framework
**Description**

Some cases are not expected to run due to unstable features, temporary changes or anything else. It should be supported to just ignore them.

**Proposal**

Implement the `ignore` interceptor. Proposed syntax:
```
-- CERESDB ignore: REASON
```
Follows the K-V format. Value for this interceptor is the reason why this case got ignored. It won't take effort and is a comment in fact.

**Additional context**

TBD: should we count/record those ignored cases and report it in the end?
ref #154

ShiKaiWi:

Closes #107 #258

# Rationale for this change
In #107, an `ignore` interceptor is proposed to support to handle some random stuff in the harness. However, there is nothing random for now in ceresdb sequential read/write procedure.

As for the __random__  table id, its reason is that the data used by harness is not reset, and the new table id for the same table is auto incremented, so what we should do is just to clean the data before running harness. Besides this, the execution order of the test case sqls contributes to the __random__ table ids.

As for the random partition number in the optimized plan, it is determined by the cpu number of the machine. Actually, this partition number can be configurable, so let's use the config to make it determined too.

- harness
  - Clean the data used by previous harness;
  - Ensure the order of execution of the cases;
  - Run harness with specified config;
  - Add `build` `clean` dependencies for `run` rule;
  - Update some failed test results;
  - Enable the harness test again;
- ceresdb server
  - Add config for query engine;
  - Refactor `context` module for query engine;

None.

# How does this change test
No need.


chunshao90 (approved):
LGTM

PR #285, opened 2022-10-09
feat: create schema in static routing

chunshao90:

Closes #

# Rationale for this change
In `static routing` mode, ceresdb  should open the `schema`s defined in config.
 

Open the `schema`s defined in config.

No.

# How does this change test

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -8,6 +8,7 @@ use analytic_engine::{
     self,
     setup::{EngineBuilder, ReplicatedEngineBuilder, RocksEngineBuilder},
 };
+use catalog::manager::ManagerRef;
 use catalog_impls::{table_based::TableBasedManager, volatile, CatalogManagerImpl};
 use cluster::cluster_impl::ClusterImpl;
 use common_util::runtime;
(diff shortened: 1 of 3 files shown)


src/setup.rs
```diff
@@ -204,3 +212,27 @@ async fn build_in_standalone_mode<Q: Executor + 'static>(
...
+
+async fn create_static_topology_schema(
+    catalog_mgr: ManagerRef,
+    static_topology_config: StaticTopologyConfig,
+) {
+    let default_catalog = catalog_mgr
+        .catalog_by_name(catalog_mgr.default_catalog_name())
+        .expect("Fail to retrieve default catalog")
+        .expect("Default catalog doesn't exist");
+    for schema_shard_view in static_topology_config.schema_shards {
```
ShiKaiWi, line 224: Is it possible that duplicate schemas exists in the `schema_shards`? However, it is still fine if the `create_schema` won't fail for creating more than one same schema.

ShiKaiWi, line 224: Is it possible that duplicate schemas exists in the `schema_shards`? However, it is still fine if the `create_schema` won't fail for creating more than one same schema.

ShiKaiWi (approved):

PR #287, opened 2022-10-09
feat: enhance sql handle, accept both json and text request

jiacai2050:

Closes #

# Rationale for this change
 
Fow now, HTTP endpoint `sql` only accept JSON request, it's not convenient to wrap sql inside `{"query": "sql"}`

Enhance sql handle endpoint, to accept both json and raw text request

Yeah, you can make http request like this directly:
```
$ curl -d 'show tables' -H 'content-type: text/plain' 0:5440/sql
{"rows":[]}
```

# How does this change test

Manually

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -46,6 +46,7 @@ tokio-stream = { version = "0.1", features = ["net"] }
 tonic = { workspace = true }
 twox-hash = "1.6"
 warp = "0.3"
+bytes = "1.2.1"
 
 [dev-dependencies]
 sql = { workspace = true , features=["test"] }
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -46,7 +46,7 @@ tokio-stream = { version = "0.1", features = ["net"] }
 tonic = { workspace = true }
 twox-hash = "1.6"
 warp = "0.3"
-bytes = "1.2.1"
+bytes = { workspace = true }
 
 [dev-dependencies]
 sql = { workspace = true , features=["test"] }
--- a/server/src/http.rs
+++ b/server/src/http.rs
@@ -67,6 +67,12 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Invalid JSON. body:{}, source:{}.", body, source))]
+    InvalidJSON {
+        body: String,
+        source: serde_json::Error,
+    },
+
     #[snafu(display("Internal err:{}.", source))]
     Internal {
         source: Box<dyn StdError + Send + Sync>,
@@ -76,6 +82,8 @@ pub enum Error {
 define_result!(Error);
 
 impl reject::Reject for Error {}
+
+const MAX_BODY_SIZE: u64 = 4096;
 
 /// Http service
 ///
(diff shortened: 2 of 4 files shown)


server/Cargo.toml
```diff
@@ -46,6 +46,7 @@ tokio-stream = { version = "0.1", features = ["net"] }
 tonic = { workspace = true }
 twox-hash = "1.6"
 warp = "0.3"
+bytes = "1.2.1"
```
ShiKaiWi, line 49: `bytes` has been defined as a workspace dependency, so here we can put it as:
```
bytes = { workspace = true }
```

server/src/http.rs
```diff
@@ -116,25 +124,36 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
...
-                    })
-                    .context(HandleRequest);
-                match result {
-                    Ok(res) => Ok(reply::json(&res)),
-                    Err(e) => Err(reject::custom(e)),
-                }
-            })
+            .and_then(
+                |content_type: String, bytes: bytes::Bytes, ctx, instance| async move {
+                    let req = if content_type.to_lowercase().contains("json") {
```
ShiKaiWi, line 133: Shall we define two const content types:
```
const JSON_CONTENT_TYPE: &str = "application/json";
const TEXT_CONTENT_TYPE: &str = "text/plain";
```

And then put it like:
```rust
let req = if context_type == JSON_CONTENT_TYPE {
    ...
} else if context_type == TEXT_CONTENT_TYPE {
    ...
} else {
   InvalidContextType.context({}).fail()?;
};
```

ShiKaiWi: Ok. And shall we define the constants and do check also for `text/plain`?

ShiKaiWi: I find a better way deserving a try, but I'm not sure whether it will work:
```
warp::path!("sql")
            .and(warp::post())
            .and(warp::header("content-type"))
            .and(warp::body::content_length_limit(MAX_BODY_SIZE)
            .and(warp::body::json().or(warp::body::bytes())))
            .and(self.with_context())
            .and(self.with_instance())
```

ShiKaiWi: After some tries, I find this works:
```rust
        let extract_sql = warp::body::json()
            .or(
                warp::body::bytes().map(|raw_sql: common_types::bytes::Bytes| {
                    Request::from(String::from_utf8_lossy(&raw_sql).to_string())
                }),
            )
            .unify();
        warp::path!("sql")
            .and(warp::post())
            // TODO(yingwen): content length limit
            .and(extract_sql)
            .and(self.with_context())
            .and(self.with_instance())
```

server/src/http.rs
```diff
@@ -116,25 +134,44 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
...
-                        Box::new(e)
-                    })
-                    .context(HandleRequest);
-                match result {
-                    Ok(res) => Ok(reply::json(&res)),
-                    Err(e) => Err(reject::custom(e)),
-                }
-            })
+            .and_then(
+                |content_type: String, bytes: bytes::Bytes, ctx, instance| async move {
```
ShiKaiWi, line 143: How about:
```rust
let extract_sql = warp::body::json()
            .or(
                warp::body::bytes().map(|raw_sql: common_types::bytes::Bytes| {
                    Request::from(String::from_utf8_lossy(&raw_sql).to_string())
                }),
            )
            .unify();
        warp::path!("sql")
            .and(warp::post())
            // TODO(yingwen): content length limit
            .and(extract_sql)
            .and(self.with_context())
            .and(self.with_instance())
            .and_then(|req, ctx, instance| async {
```

server/src/http.rs
ShiKaiWi: Yes, but there is a pattern for the or filter.  Take a look at the example metioned above.

ShiKaiWi: > After some tries, I find this works:
> ```rust
>         let extract_sql = warp::body::json()
>             .or(
>                 warp::body::bytes().map(|raw_sql: common_types::bytes::Bytes| {
>                     Request::from(String::from_utf8_lossy(&raw_sql).to_string())
>                 }),
>             )
>             .unify();
>         warp::path!("sql")
>             .and(warp::post())
>             // TODO(yingwen): content length limit
>             .and(extract_sql)
>             .and(self.with_context())
>             .and(self.with_instance())
>             .and_then(|req, ctx, instance| async {
> ```

Take a look at this.

ShiKaiWi (approved):
It's more simple and elegant!

PR #288, opened 2022-10-10
refactor: modify exist wal

### LINKED ISSUE #283: Modify the exist wal interface to adapt the new design
**Description**
Now the wal interface's design is for the mapping `one table to one region (wal unit)`. For adapting to the new wal implementation based on kafka and solving some exist problem (for example, replay is too slow), we plan to modify the wal interface. 

**Proposal**
+ Make `region id` to `WalLocation` (include `region id` and `table id` now).
+ Try best to keep current process in other modules (in fact, I prefer to add new components for avoiding direct interacting between `Instance` and `WalManager` to keep current process in later development).
+ Add a new method `scan` to `WalManager`(implement it in wal on kafka firstly, other implementations will implement and make use of it after).

**Additional context**

Rachelint:

Closes #283 

# Rationale for this change
see #283 

+ Add `WalLocation` to replace original `RegionId`. 
+ Add `scan` method to `WalManager`.

None.

# How does this change test
Test by exist ut.


wal/src/manager.rs
```diff
@@ -119,12 +119,28 @@ pub mod error {
...
-use common_types::{MAX_SEQUENCE_NUMBER, MIN_SEQUENCE_NUMBER};
+use common_types::{table::TableId, MAX_SEQUENCE_NUMBER, MIN_SEQUENCE_NUMBER};
 pub use error::*;
 
 pub type RegionId = u64;
 pub const MAX_REGION_ID: RegionId = u64::MAX;
 
+/// Used to wals of the table.
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct WalLocation {
```
ShiKaiWi, line 130: `Location` seems enough. After all, there is no such a struct called `WalRegionId`.

wal/src/manager.rs
```diff
@@ -119,12 +119,28 @@ pub mod error {
...
 }
 
-use common_types::{MAX_SEQUENCE_NUMBER, MIN_SEQUENCE_NUMBER};
+use common_types::{table::TableId, MAX_SEQUENCE_NUMBER, MIN_SEQUENCE_NUMBER};
 pub use error::*;
 
 pub type RegionId = u64;
 pub const MAX_REGION_ID: RegionId = u64::MAX;
 
+/// Used to wals of the table.
```
ShiKaiWi, line 128: Used to describe where to store the wals of the table.

wal/src/manager.rs
```diff
@@ -119,12 +119,28 @@ pub mod error {
     define_result!(Error);
 }
 
-use common_types::{MAX_SEQUENCE_NUMBER, MIN_SEQUENCE_NUMBER};
+use common_types::{table::TableId, MAX_SEQUENCE_NUMBER, MIN_SEQUENCE_NUMBER};
```
ShiKaiWi, line 122: These import statements seem weird to be put here, and how about put them in the beginning of the file?

wal/src/manager.rs
```diff
@@ -207,15 +223,23 @@ impl ReadBoundary {
...
     /// Start bound
     pub start: ReadBoundary,
     /// End bound
     pub end: ReadBoundary,
 }
 
+#[derive(Debug, Clone)]
+pub struct ScanRequest {
+    /// WalLocation of the wal to read
+    pub wal_location: WalLocation,
```
ShiKaiWi, line 238: Shouldn't `scan` read the data of the whole region? Otherwise, it seems no essential difference between the `read_batch` and `scan`.

wal/src/manager.rs
Rachelint: It seems reasonable in wal crate. But I think it should be renamed to `WalLocation` while use in other modules by `use as`.

wal/src/manager.rs

wal/src/manager.rs

wal/src/manager.rs

wal/src/manager.rs
ShiKaiWi: `Location` should not be a usual name. In my memory, there is no struct called `Location` now.

Rachelint: > `Location` should not be a usual name. In my memory, there is no struct called `Location` now.

I mean `Location`s in other crate, such as `std::panic::Location`, `snafu::Location`. And If just say `Location`, I will think it is used to locate the table in database first but not the location in wal module.

Rachelint: Done.

common_types/src/table.rs
```diff
@@ -2,3 +2,17 @@
 
 pub type TableId = u64;
 pub type TableName = String;
+pub type ShardId = u32;
+pub const DEFAULT_SHARD_ID: u32 = 0;
+
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
```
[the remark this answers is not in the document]
Rachelint: Done.

catalog_impls/src/volatile.rs
```diff
@@ -302,6 +302,8 @@ impl Schema for SchemaImpl {
             .map(|v| TableId::from(v.id))?;
 
         let request = request.into_engine_create_request(table_id);
+        // TODO(kamille) should get and assign the `shard_id` to create table request
```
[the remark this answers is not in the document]
Rachelint: Done.

analytic_engine/src/instance/open.rs
```diff
@@ -196,33 +196,39 @@ impl Instance {
...
 
-        info!("Instance recover table:{} meta end", table_id);
+        info!("Instance recover table:{} meta end", request.table_id);
 
         Ok(table_data)
     }
 
     /// Apply manifest data to instance
-    async fn apply_table_manifest_data(
+    async fn recover_table_data(
```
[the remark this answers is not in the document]
Rachelint: Done.

ShiKaiWi (approved):
Good job! Let's wait for the CI finishes.

PR #292, opened 2022-10-11
chore: fix CI cache key, release mode should not mix with debug

jiacai2050:

Closes #

# Rationale for this change
 
In current CI, the cache size is 4G after gzipped, which is too large, one issue for this is that the cache may contains both debug and release target, which is useless
- https://github.com/CeresDB/ceresdb/actions/runs/3224489192/jobs/5275667928

Update CI cache key, so release target will not be mixed with debug mode.

No

# How does this change test

Not required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,11 +53,11 @@ jobs:
           path: |
             ~/.cargo
             ./target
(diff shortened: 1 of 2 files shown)


ShiKaiWi (approved):
LGTM

ShiKaiWi (approved):

PR #293, opened 2022-10-11
feat: support meta event service

ShiKaiWi:

Part of #235

# Rationale for this change
 After the event loop mechanism is removed, we support a meta event service instead.

- Support the meta event service (only the skeleton without any implementations).

None.

# How does this change test
Not ready for test.


chunshao90 (approved):
LGTM

PR #295, opened 2022-10-12
feat: remove cluster catalog and refactor volatile catalog

ShiKaiWi:

Part of #235 

# Rationale for this change
 Refer to #235 

- Remove the cluster catalog;
- Refactor volatile catalog;
- Now the catalog-schema-table information is managed by `Catalog`, and the shard-table information is managed by `Cluster.ShardTableManager`;
- Update the ceresdbproto to the latest version;

# How does this change test


cluster/src/cluster_impl.rs
```diff
@@ -174,6 +175,54 @@ impl Inner {
 
         Ok(resp)
     }
+
+    async fn open_shard(&self, req: &OpenShardRequest, _opts: OpenShardOpts) -> Result<()> {
+        let shard_info = req.shard.as_ref().context(OpenShard {
+            shard_id: 0u32,
```
Rachelint, line 181: Replace by a const? Such as `DEFAULT_SHARD_ID`?

catalog_impls/src/volatile.rs
```diff
@@ -251,163 +233,119 @@ impl Schema for SchemaImpl {
...
+        if let Some(table) = self.get_table_with_check(
             &request.catalog_name,
             &request.schema_name,
             &request.table_name,
         )? {
             return Ok(table);
         }
 
-        todo!("The implementation will be removed after cluster implementation is ready");
+        let table = self
```
Rachelint, line 265: It seems too strange if nothing done here.

cluster/src/table_manager.rs
```diff
@@ -52,6 +52,14 @@ impl TableManager {
...
 
+    pub fn contains_shard(&self, shard_id: ShardId) -> bool {
+        self.inner
+            .read()
+            .unwrap()
+            .shard_infos
+            .contains_key(&shard_id)
+    }
+
     pub fn update_table_info(&self, shard_table: &HashMap<ShardId, ShardTables>) {
```
Rachelint, line 63: I think the name ShardTables is too strange, how about TablesInShard?
And rename shard_table to shard_to_tables?

cluster/src/table_manager.rs
```diff
@@ -148,6 +133,30 @@ impl TableManager {
         }
     }
 
+    // todo: should accept schema id as param instead of schema name?
```
Rachelint, line 136: TODO

cluster/src/table_manager.rs
```diff
@@ -148,6 +133,30 @@ impl TableManager {
...
+    pub fn tables_by_schema(&self, catalog: &str, schema: &str) -> Vec<TableRef> {
```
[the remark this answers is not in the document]
ShiKaiWi: `tables_by_catalog_and_schema` seems too verbose, and `by_schema` denotes implicitly that the catalog should be provided because no schema can exists outside the catalog.

cluster/src/table_manager.rs
ShiKaiWi: There is no significant difference between `TablesInShard` and `ShardTables`, but it is hard to change such a name because the protobuf files should be changed also. So I prefer to keep this name.

As for the parameter name `shard_table`, I will replace it with `shard_tables`.

cluster/src/cluster_impl.rs
ShiKaiWi: Actually, this is the error field of `OpenShard` and in this case, the error denotes the shard is not found in the request, so it is wired to set it as DEFAULT_SHARD_ID.

cluster/src/table_manager.rs
Rachelint: When I see shard table, I will think it is a table used to record the information of shard.

cluster/src/table_manager.rs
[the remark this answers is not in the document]
Rachelint: I see `SpaceAndTable` in ceresdb and `RecordAndOffset` in rskafka. I think such naming is not so rare actually.

cluster/src/cluster_impl.rs
Rachelint: Ok.

cluster/src/table_manager.rs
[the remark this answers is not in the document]
ShiKaiWi: Ok, but the code is outdated, and in the new codes, there is no such hard naming issue. :stuck_out_tongue_winking_eye:

Rachelint (approved):

Issue #291, opened 2022-10-11
replace custom ParquetSstReader to DataFusion's ParquetExec


jiacai2050:
### Describe This Problem

In current implementation
- `ParquetSstReader` use CacheableSerializedFileReader to read parquet file, and
- `CacheableSerializedFileReader` is modeled after https://github.com/apache/arrow-rs/blob/5.2.0/parquet/src/file/serialized_reader.rs, which contains lots of low-level details about how to parse parquet, and there are already some issues with it, such as #271

### Proposal

By leverage DataFusion's [ParquetExec](https://docs.rs/datafusion/13.0.0/datafusion/physical_plan/file_format/struct.ParquetExec.html) to implement this, we can remove all those low-level details and it's already feature rich, for example:
- Support predicate pushdown
- Support custom [ParquetFileReaderFactory](https://docs.rs/datafusion/13.0.0/datafusion/physical_plan/file_format/trait.ParquetFileReaderFactory.html), which can be used to implement row-group level cache, just as what we do now 

### Additional Context

IOx already does this.
- https://github.com/influxdata/influxdb_iox/pull/5531/files

ygf11:
Seems this issue is relative to:

* https://github.com/CeresDB/ceresdb/issues/256

In #256, we want to use datafusion's expr to implement filter logic.

If we replace `ParquetSstReader` with `ParquetExec`, I think the filter logic is also there.

ygf11:
> @ygf11 yeah, those issue are related, but they work in different levels.

Thanks, I almost got it. If I have mistake, please take it out.

* We will read data from two places, `Memtable` and `Sst files`. And we will read sst with `ParquetExec` after this task.
* For `sst files`, we can directly pass `filter expr` to `ParquetExec`.
* For `Memtable`, we still need implement filer logics manually, that's what #256 need do.
* Further more, after https://github.com/CeresDB/ceresdb/issues/256 and this task, we can remove `FilterExec`, since we have pushed down all filters.

ShiKaiWi:
There are some mistakes here:
> For sst files, we can directly pass filter expr to ParquetExec.

This is true, but one thing worth mentioning that the data is not ensured to meet the requirements by pushed down predicates (actually the filter is implemented by the min-max index on the row-group level.)

> For Memtables, we still need implement filter logics manually, that's what https://github.com/CeresDB/ceresdb/issues/256 need do.

[...]

PR #296, opened 2022-10-12
feat: bump object store to 0.5.1, same with datafusion

jiacai2050:

Part of #291 

# Rationale for this change
 
Now ceresdb use object store 0.1.0, which is different datafusion, this block #291, and compile fails 
image

Bump object store to 0.5.1
- https://github.com/apache/arrow-rs/blob/master/object_store/CHANGELOG.md

No
# How does this change test

Existing UT

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1476,7 +1476,7 @@ dependencies = [
  "lazy_static",
  "log",
  "num_cpus",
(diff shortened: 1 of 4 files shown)


Rachelint (commented):
LGTM

Rachelint (approved):

PR #300, opened 2022-10-13
chore: rename ShardTables to TablesOfShard

ShiKaiWi:

Closes #

# Rationale for this change
 In the #295, @Rachelint proposes that the `ShardTables` is an ambiguous name, so let's rename it to `TablesOfShard`.

- Renaming;
- Update the version of ceresdbproto for renaming protobuf struct name;
- Remove some unused depenecies;

No.

# How does this change test
CI.


Rachelint (approved):
LGTM

PR #303, opened 2022-10-14
feat: add TabelCreator/TableDroper for executing create/drop table

ShiKaiWi:

Part of #235 

# Rationale for this change
 Refer to #235 

Add `TableCreator`/`TableDropper` to Create/Drop interpreter so that we can choose different implements of create/drop table. After this PR, two kinds of implementations will be provided: `catalog_based` and `meta_based`.

None.

# How does this change test
Existing UT.


interpreters/src/drop/table_dropper/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+use crate::{context::Context, drop::Result, interpreter::Output};
+
+pub mod catalog_based;
+pub mod meta_based;
+
+pub type TableDropperRef = Arc<dyn TableDropper + Send + Sync>;
+
+#[async_trait]
+pub trait TableDropper {
+    async fn drop_table(
```
Rachelint, line 18: I think `drop_table` in `TableDropper` may be a bit verbose. Same as `TableCreator`.

server/src/server.rs
```diff
@@ -218,14 +219,25 @@ impl<Q: QueryExecutor + 'static> Builder<Q> {
...
         let function_registry = self.function_registry.context(MissingFunctionRegistry)?;
-        let instance = Instance {
-            catalog_manager,
-            query_executor,
-            table_engine,
-            function_registry,
-            limiter: self.limiter,
+
+        let instance = {
+            let creator = Arc::new(table_creator::catalog_based::TableCreatorImpl::new(
```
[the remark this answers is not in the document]
ShiKaiWi: `table_creator`has been already imported as the module name.

interpreters/src/drop/table_dropper/mod.rs
ShiKaiWi: It's true, and I named them as `drop/create` at first. However, the `drop` method has been used by `Drop` trait, and I have to name it `drop_table` here.

Rachelint: Why not make `Manipulator` and place drop and create logic into it?

Rachelint (approved):
LGTM

PR #305, opened 2022-10-16
fix: avoid running blocked-on task when building meta client

### LINKED ISSUE #304: Can't start ceresdb-server in cluster mode
### Describe this problem

It will panic if start ceresdb-server in cluster mode:
```
ERRO [common_util/src/panic.rs:42] thread 'main' panicked 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.' at "/Users/wangzuliang/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.20.1/src/runtime/thread_pool/mod.rs:89"
```

### Steps to reproduce

Run the ceresdb-server with: `./target/release/ceresdb-server --config ./docs/example-cluster-0.toml`.

### Expected behavior

It won't panic.

### Additional Information

_No response_

ShiKaiWi:

Closes #304 

# Rationale for this change
 A blocked-on task for building meta client is started in the init procedure, leading to the panic mentioned by #304.

Avoid running blocked-on task in the init procedure.

# How does this change test
Now this is tested by manual.

### DIFF
--- a/meta_client/src/lib.rs
+++ b/meta_client/src/lib.rs
@@ -27,11 +27,13 @@ pub enum Error {
     MissingHeader { backtrace: Backtrace },
 
     #[snafu(display(
-        "Failed to connect the service endpoint, err:{}\nBacktrace:\n{}",
+        "Failed to connect the service endpoint:{}, err:{}\nBacktrace:\n{}",
+        addr,
         source,
         backtrace
     ))]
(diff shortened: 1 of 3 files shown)


Rachelint (approved):
LGTM

PR #306, opened 2022-10-16
feat: message queue and its kafka impl 

### LINKED ISSUE #298: Define a message queue interface and its Kafka implementation
**Description**
In the future, maybe we need to support other message queue, so make an abstract message queue interface may be useful. Otherwise, such an interface is convenient for test, too.

**Proposal**
+ Define a abstract message queue.
+ Implement it for Kafka. 

**Additional context**
The scan method is needed in CeresDB, it will return a iterator and high-level should poll the iterator to get all data in topic. For ensuring scan process's normal running, freezing the partition during scanning is necessary, otherwise error will occur. For reaching the target, following points should be considered:
+ Kafka's message retention. Too old logs may be deleted by Kafka (`default 7 days`, normally won't be reset), but it make almost no difference to us, because CeresDB will flush data periodically (`default 5 hours`), and drop or close tables will trigger flush, too. So it is almost impossible to leave such too old data in wal, and it is abnormal to keep data in wal for such a long time in theory because wal is a write buffer actually.

+ Two leader shard. If there are two leader in a specific shard, it may make the scanning mad, and everything will be unpredictable. To ensure such situation will not happen is the responsibility of `Ceresmeta`.

+ Other threads concurrently operate the partition in a single node. Ensuring no others will operate the partition while it is being scanned due to recovering is necessary. It is easy to control.

Rachelint:

Closes #298 

# Rationale for this change
 See #298 

# How does this change test

### DIFF
--- a/components/message_queue/Cargo.toml
+++ b/components/message_queue/Cargo.toml
@@ -12,13 +12,17 @@ common_util = { workspace = true }
 serde = { workspace = true }
 serde_derive = { workspace = true }
 snafu = { workspace = true }
-rskafka = { version = "0.3.0", default-features = false, features = ["compression-gzip", "compression-lz4", "compression-snappy"] }
-# rskafka = "0.3.0"
-time = "0.3.15"
+chrono = { workspace = true }
 async-trait = { workspace = true }
-dashmap = "5.4.0"
 log = { workspace = true }
 futures = { workspace = true }
+tokio = { workspace = true }
+
+[dependencies.rskafka]
+git = "https://github.com/influxdata/rskafka.git"
+rev = "00988a564b1db0249d858065fc110476c075efad"
+default-features = false
+features = ["compression-gzip", "compression-lz4", "compression-snappy"]
 
 [dev-dependencies]
-rand = { workspace = true }
+uuid = { version = "1.0", features = ["v4"] }
--- a/components/message_queue/Cargo.toml
+++ b/components/message_queue/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "message_queue"
+version = "0.1.0"
+
+[package.edition]
+workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+common_util = { workspace = true }
+serde = { workspace = true }
+serde_derive = { workspace = true }
+snafu = { workspace = true }
+rskafka = { version = "0.3.0", default-features = false, features = ["compression-gzip", "compression-lz4", "compression-snappy"] }
+# rskafka = "0.3.0"
+time = "0.3.15"
+async-trait = { workspace = true }
+dashmap = "5.4.0"
+log = { workspace = true }
+futures = { workspace = true }
+
+[dev-dependencies]
+rand = { workspace = true }
--- a/components/message_queue/src/kafka/config.rs
+++ b/components/message_queue/src/kafka/config.rs
@@ -1,3 +1,7 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+//! Kafka implementation's config
+
 use serde_derive::{Deserialize, Serialize};
 
 /// Generic client config that is used for consumers, producers as well as admin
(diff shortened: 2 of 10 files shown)


components/message_queue/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+rskafka = { version = "0.3.0", default-features = false, features = ["compression-gzip", "compression-lz4", "compression-snappy"] }
+# rskafka = "0.3.0"
+time = "0.3.15"
+async-trait = { workspace = true }
+dashmap = "5.4.0"
+log = { workspace = true }
+futures = { workspace = true }
+
+[dev-dependencies]
+rand = { workspace = true }
```
ShiKaiWi, line 24: This dep seems unused?

components/message_queue/src/kafka/config.rs
```diff
@@ -0,0 +1,90 @@
...
+            timeout_ms: 5000,
+        }
+    }
+}
+
+/// Config for consumers.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct WalConfig {
+    /// Will wait for at least `min_batch_size` bytes of data
```
ShiKaiWi, line 60: It seems that the comments doesn't correspond to the three fields.

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,396 @@
...
+};
+use snafu::{ensure, ResultExt, Snafu};
+
+use super::config::WalConfig;
+use crate::{kafka::config::Config, ConsumeAllIterator, Message, MessageAndOffset, MessageQueue};
+
+/// The topic(with just one partition) client for Kafka
+//
+/// `Arc` is needed to ensure its lifetime. Because in future's gc process,
+/// it may has removed from pool but still is still being used.
```
ShiKaiWi, line 26:
```suggestion
/// The topic (with just one partition) client for Kafka
//
/// `Arc` is needed to ensure its lifetime because in future's gc process,
/// it may has removed from pool but be still in use.
```

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,396 @@
...
+type TopicClientRef = Arc<PartitionClient>;
+const PARTITIONS_NUM: i32 = 1;
```
ShiKaiWi, line 28:
```suggestion
const PARTITION_NUM: u32 = 1;
```

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,396 @@
...
+        source: RskafkaError,
+        when: ConsumeAllWhen,
+    },
+
+    #[snafu(display(
+        "Race happened in scanning partition in topic:{}, when:{}",
+        topic_name,
+        when
+    ))]
+    ConsumeAllRace {
```
ShiKaiWi, line 67: For such error without cause error, we may need to attach backtrace to it.

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,396 @@
...
+        let records: Vec<Record> = message.into_iter().map(|m| m.into()).collect();
+        Ok(topic_client
+            .produce(records, Compression::default())
+            .await
+            .context(Produce {
+                topic_name: topic_name.to_string(),
+            })?)
+    }
+
+    async fn consume_all(&self, topic_name: &str) -> Result<KafkaConsumeAllIterator> {
```
ShiKaiWi, line 215: Why not just call it consume? And call `ConsumeAllIterator` as `ConsumeIterator`?

Rachelint: Because in message queue, `consume` usually means consume one by one and should return one message per time, but the method here will return all the data in the topic by a iterator.
I propose to  implement the `consume` method in the future for the logic of fetching process from master in the slave.

Renaming the iterator to `ConsumeIterator ` is reasonable, I will do it.

components/message_queue/Cargo.toml
Rachelint: It will be used in the test later.

components/message_queue/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::{collections::BTreeMap, result::Result};
+
+use async_trait::async_trait;
+use chrono::{DateTime, Utc};
+
+/// Topic's producer
+///
+/// In ceresdb every topic just has one partition.
+/// Because there won't be a huge number of topic(the upper bound is just about
+/// 1000).
```
ShiKaiWi, line 18: It is hard to understand these comments, and how about just removing them here?

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,448 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ConsumeAllWhen::Start => f.write_str("start"),
+            ConsumeAllWhen::InitIterator => f.write_str("init_iterator"),
+            ConsumeAllWhen::PollStream => f.write_str("poll_stream"),
+        }
+    }
+}
+
+define_result!(Error);
```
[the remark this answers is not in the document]
Rachelint: Done.

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,448 @@
...
+        }
+
+        // Otherwise, we should make a double-check first,
+        // and if still not found(other thread may has inserted it),
+        // we should create it.
+        let mut topic_client_pool = self.topic_client_pool.write().await;
+        if let Some(client) = topic_client_pool.get(topic_name) {
+            Ok(client.clone())
+        } else {
+            Ok(topic_client_pool
```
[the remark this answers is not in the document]
Rachelint: Done.

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,448 @@
...
+                .get_offset(OffsetAt::Latest)
+                .await
+                .context(ConsumeAll {
+                    topic_name: topic_name.to_string(),
+                    when: ConsumeAllWhen::InitIterator,
+                })?;
+        ensure!(
+            start_offset <= high_watermark,
+            ConsumeAllRace {
+                topic_name: topic_name.to_string(),
```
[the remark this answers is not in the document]
Rachelint: Done.

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -0,0 +1,448 @@
...
+            None => {
+                return None;
+            }
+        };
+
+        // Return message and offset from buffer.
+        match stream.next().await {
+            Some(Ok((record, high_watermark))) => match high_watermark.cmp(&self.high_watermark) {
+                Ordering::Greater => Some(
+                    ConsumeAllRace {
```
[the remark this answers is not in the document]
Rachelint: Done.

components/message_queue/src/lib.rs

ShiKaiWi (approved):

PR #307, opened 2022-10-17
feat: support open&close shards

ShiKaiWi:

Part of #235 

# Rationale for this change
 Implemenat the logic for close/open shards.

Support close/open shards.

None.

# How does this change test
No test for this, and this will be tested with meta by manual.

### DIFF
--- a/cluster/src/shard_table_manager.rs
+++ b/cluster/src/shard_table_manager.rs
@@ -68,7 +68,7 @@ impl ShardTableManager {
 #[derive(Debug, Default)]
 struct Inner {
     // Tables organized by shard.
-    // TODO: take considerations about the role.
+    // TODO: The shard roles should be also taken into considerations.
     tables_by_shard: HashMap<ShardId, TablesOfShard>,
 }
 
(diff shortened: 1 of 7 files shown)


Rachelint (commented):
I think the file `server/src/grpc/mod.rs`  may be too huge. How about make a `service` directory as a mod, split the logic in mod.rs to `storage.rs` and `meta.rs`(or a better name)  and place them in `service`. 
It can be done in the future.

cluster/src/shard_table_manager.rs
```diff
@@ -43,6 +43,11 @@ impl ShardTableManager {
         self.inner.read().unwrap().contains_shard(shard_id)
```
Rachelint, line 43: In fact, I think `ShardTableManager` is just a struct like xxx_map. How about name it like`ShardToTablesMap`? And rename `update_tables_of_shard` to `insert_or_update`, `remove_shard` to `remove` ?

ShiKaiWi: I don't think it is a good practice to suffix a struct with its type. Furthermore, current `ShardTableManager` provides more functionalities than the mapping from shard to tables, e.g. `find_table_by_name`. In conclusion, such renaming is not a good idea for me.

Issue #268, opened 2022-09-23
Centralize the logic of choosing worker for table into one place


ShiKaiWi:
### Describe This Problem

When carry on operations (write, flush etc.) on one table, a specific worker should be chosen. However, the logic for choosing such a worker distributes across the codebase:

https://github.com/CeresDB/ceresdb/blob/ccf988f5629436927479721b63df53b220967160/analytic_engine/src/instance/write_worker.rs#L307

https://github.com/CeresDB/ceresdb/blob/ccf988f5629436927479721b63df53b220967160/analytic_engine/src/instance/write_worker.rs#L669

https://github.com/CeresDB/ceresdb/blob/ccf988f5629436927479721b63df53b220967160/analytic_engine/src/table/data.rs#L541

### Proposal

We hope the logic of choosing the worker for a specific table should be centralized in one place (one function).

QuintinTao:
could I ?
I want add a function in sst_util.rs
pub fn find_worker(table_id: usize, worker_num: usize) -> usize {
    table_id % worker_num
}

replace the three lines you mentioned.
 if find_worker(table_id, worker_num) != worker_id {

ShiKaiWi:
> could I ? I want add a function in sst_util.rs pub fn find_worker(table_id: usize, worker_num: usize) -> usize { table_id % worker_num }
> 
> replace the three lines you mentioned. if find_worker(table_id, worker_num) != worker_id {

@QuintinTao Thanks. Any contributions are welcome. The proposal is nice, and is it better to put the function `find_worker` in `writer_worker.rs`?

QuintinTao:
> > could I ? I want add a function in sst_util.rs pub fn find_worker(table_id: usize, worker_num: usize) -> usize { table_id % worker_num }
> > replace the three lines you mentioned. if find_worker(table_id, worker_num) != worker_id {
> 
> @QuintinTao Thanks. Any contributions are welcome. The proposal is nice, and is it better to put the function `find_worker` in `writer_worker.rs`?

OK

PR #309, opened 2022-10-17
Centralize the logic of choosing worker for table into one place #268

QuintinTao:
I added a function in the write_worker.rs to centralize the logic of choosing worker:

```Rust
pub fn find_worker(table_id: usize, worker_num: usize) -> usize {
table_id % worker_num
}
```

### DIFF
--- a/analytic_engine/src/instance/write_worker.rs
+++ b/analytic_engine/src/instance/write_worker.rs
@@ -304,7 +304,7 @@ impl WorkerLocal {
         worker_num: usize,
     ) -> Result<()> {
         let worker_id = self.data.as_ref().id;
-        if table_id % worker_num != worker_id {
+        if find_worker(table_id, worker_num) != worker_id {
             return DataNotLegal {
                 table: table_name,
                 worker_id,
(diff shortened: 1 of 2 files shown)


analytic_engine/src/instance/write_worker.rs
```diff
@@ -989,7 +989,10 @@ impl WriteWorker {
         self.local.data.background_notify.notified().await;
     }
 }
-
+///Centralize the logic of choosing worker for table into one place #268
+pub fn find_worker(table_id: usize, worker_num: usize) -> usize {
```
ShiKaiWi, line 993: Maybe we can replace the type usize of `table_id` with `TableId`.

ShiKaiWi:
@QuintinTao Thanks for your contributions. I have left some comments about your PR, and you can address them when you are convenient.

ShiKaiWi (commented):
I guess the ci may report some style errors, which will block the pr merge, so you can fix them when the ci results are ready.

ShiKaiWi:
> I guess the ci may report some style errors, which will block the pr merge, so you can fix them when the ci results are ready.

BTW, you can execute `cargo fmt` locally to see which codes are need fixing.

QuintinTao:
> report some style errors, which will block the pr merge, so you can fix them when the ci results are ready.

OK I will check it. thanks.

PR #312, opened 2022-10-18
WIP: add datafusion based parquet reader

### LINKED ISSUE #291: replace custom ParquetSstReader to DataFusion's ParquetExec

jiacai2050:

Closes #291

# Rationale for this change
 
As described in #291, this PR also fix object store cache isn't working.
After #14, parquet reader will read all bytes out, ignoring whether if it's already cached.

Replace hand-rolled parquet reader with datafusion's ParquetExec, and add `CachableParquetFileReader` to implement row-group level cache

No

# How does this change test

Using existing UT


analytic_engine/src/sst/parquet/reader.rs
```diff
@@ -3,393 +3,506 @@
...
 };
+use log::{debug, info};
+use object_store::{ObjectMeta, ObjectStoreRef, Path};
+use parquet::arrow::async_reader::AsyncFileReader;
+use parquet_ext::{DataCacheRef, MetaCacheRef};
 use snafu::{ensure, OptionExt, ResultExt};
 use table_engine::predicate::PredicateRef;
-use tokio::sync::mpsc::{self, Receiver, Sender};
 
+use super::encoding::{self, ParquetDecoder};
```
ShiKaiWi, line 45: Avoid the relative importing path.

analytic_engine/src/sst/parquet/reader.rs
```diff
@@ -3,393 +3,506 @@
...
-                );
-                arrow_reader.get_record_reader_by_columns(proj_mask, self.batch_size)
+impl Drop for CachableParquetFileReader {
+    fn drop(&mut self) {
+        info!(
+            "CachableParquetFileReader meta_size_hint:{:?}, cache_hit:{}, cache_miss:{}, bytes_scanned:{}",
+            self.metadata_size_hint, self.cache_hit, self.cache_miss, self.metrics.bytes_scanned.value()
+        );
+    }
+}
```
ShiKaiWi, line 322: Miss one newline here.

common_types/src/time.rs
```diff
@@ -276,6 +280,18 @@ impl TimeRange {
             self.exclusive_end.min(other.exclusive_end),
         )
     }
+
+    /// Creates expression like:
+    /// start <= time && time < end
+    pub fn df_expr(&self, column_name: impl AsRef<str>) -> Expr {
```
ShiKaiWi, line 286:
```suggestion
    pub fn to_df_expr(&self, column_name: impl AsRef<str>) -> Expr {
```

PR #319, opened 2022-10-18
test: set correct exit code for test harness

jiacai2050:

Closes #

# Rationale for this change
 
For now, test harness will always exit with 0, even when there are failed cases.

Return `Result` in main, so exit code will not be 0 when case failed

No

# How does this change test

Run test harness locally.


ShiKaiWi (approved):
LGTM

PR #320, opened 2022-10-18
feat: add async inner to `BatchLogIteratorAdapter`

Rachelint:
…and add its adapting logic to `BatchLogIteratorAdapter`.

Closes #

# Rationale for this change
 The `BatchLogIteratorAdapter` just support `BlockingLogIterator` as its inner. It is not enough, because in other wal implementation, the iterator return from `read_batch` may be originally async, the  `BatchLogIteratorAdapter` should take  it in consideration.

+ Define a new trait  `AsyncLogIterator`.
+ Add its adapting logic to `BatchLogIteratorAdapter`.

None.

# How does this change test
By exist ut.

### DIFF
--- a/wal/src/manager.rs
+++ b/wal/src/manager.rs
@@ -403,7 +403,7 @@ mod tests {
     use std::{collections::VecDeque, sync::Arc};
 
     use async_trait::async_trait;
-    use common_util::runtime;
+    use common_util::runtime::{self, Runtime};
 
     use super::{AsyncLogIterator, BatchLogIteratorAdapter, InnerIterator, SyncLogIterator};
     use crate::{log_batch::LogEntry, tests::util::TestPayloadDecoder};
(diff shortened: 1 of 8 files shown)


wal/src/manager.rs
```diff
@@ -286,44 +278,41 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
...
+    Sync(Box<dyn SyncLogIterator>, Arc<Runtime>),
+    Async(Box<dyn AsyncLogIterator>),
+}
+
+/// Adapter to convert a sync/async iterator to a batch async iterator.
 #[derive(Debug)]
 pub struct BatchLogIteratorAdapter {
-    blocking_iter: Option<Box<dyn BlockingLogIterator>>,
-    runtime: Arc<Runtime>,
+    inner_iterator: Option<InnerIterator>,
```
ShiKaiWi, line 290: Just call it iter.

wal/src/manager.rs
```diff
@@ -286,44 +278,41 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
...
     batch_size: usize,
 }
 
 impl BatchLogIteratorAdapter {
-    pub fn new(
-        blocking_iter: Box<dyn BlockingLogIterator>,
-        runtime: Arc<Runtime>,
-        batch_size: usize,
-    ) -> Self {
+    pub fn new(inner_iterator: InnerIterator, batch_size: usize) -> Self {
```
ShiKaiWi, line 295: Maybe we should define two methods: `new_sync` & `new_async`.

wal/src/manager.rs
```diff
@@ -286,44 +278,41 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
     async fn scan(&self, ctx: &ScanContext, req: &ScanRequest) -> Result<BatchLogIteratorAdapter>;
 }
 
-/// Adapter to convert a blocking interator to a batch async iterator.
+#[derive(Debug)]
+pub enum InnerIterator {
+    Sync(Box<dyn SyncLogIterator>, Arc<Runtime>),
```
ShiKaiWi, line 283: Let's give the two fields names:
```rust
Sync {
    iter: Box<dyn SyncLogIterator>,
    runtime: Arc<Runtime>,
}
```

wal/src/manager.rs
```diff
@@ -286,44 +278,41 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
     async fn scan(&self, ctx: &ScanContext, req: &ScanRequest) -> Result<BatchLogIteratorAdapter>;
 }
 
-/// Adapter to convert a blocking interator to a batch async iterator.
+#[derive(Debug)]
+pub enum InnerIterator {
```
ShiKaiWi, line 282: How about naming it `BatchLogIterator`?

wal/src/manager.rs
```diff
@@ -286,44 +278,41 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
...
             batch_size,
         }
     }
 }
 
-#[async_trait]
-impl BatchLogIterator for BatchLogIteratorAdapter {
-    async fn next_log_entries<D: PayloadDecoder + Send + 'static>(
+impl BatchLogIteratorAdapter {
+    async fn simulated_async_next<D: PayloadDecoder + Send + 'static>(
```
ShiKaiWi, line 304: Add comments to explain what it means when `InnerIterator` is None.

wal/src/manager.rs
Rachelint: Yes, quite verbose to new it now.

wal/src/manager.rs
Rachelint: But the inner's return is not a batch, but is just a single log entry.
I propose to rename it to `LogIterator`.

wal/src/manager.rs

wal/src/manager.rs

wal/src/manager.rs

wal/src/manager.rs

wal/src/manager.rs
Rachelint: Done.

ShiKaiWi (approved):

PR #322, opened 2022-10-19
chore: minimize release binary size

chunshao90:

Closes #

# Rationale for this change
Minimize release binary size. 
Refer to [min-sized-rust](https://github.com/johnthagen/min-sized-rust).

* Enable [lto](https://doc.rust-lang.org/cargo/reference/profiles.html#lto).
* Set [opt-level](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level) to "z".
The new setting minimizes the release binary size from 1.3G to 510M.

# How does this change test

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -131,7 +131,8 @@ vergen = { version = "5", default-features = false, features = ["build", "git"] 
 
 [profile.release]
 debug = true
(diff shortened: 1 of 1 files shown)


archerny (approved):
LGTM

PR #323, opened 2022-10-19
feat: impl create drop table service

### LINKED ISSUE #235: Refactor Cluster and Catalog mods

ShiKaiWi:

Closes #235 

# Rationale for this change
 This is the final part of #235, implement the service that do create/drop table on the specific shard.

- Implement the service that handle creating/dropping table on specific shard;
- Make storage_service and meta_event_service as separate module;
- Use macro to simplify the codes of meta service;

None.

# How does this change test
Will be tested by manual.


cluster/src/shard_tables_cache.rs
```diff
@@ -55,6 +58,46 @@ impl ShardTablesCache {
...
+        self.inner.write().unwrap().try_insert_table_to_shard(
+            prev_shard_version,
+            curr_shard,
+            new_table,
+        )
+    }
+
+    /// Try to remove a table from the shard with a newer version.
+    ///
+    /// It will fail if:
```
Rachelint, line 84: The fail situations are same as insert_table_to_shard, how about: 
`It will fail in the same situations as insert_table_to_shard`

cluster/src/shard_tables_cache.rs
```diff
@@ -113,4 +156,82 @@ impl Inner {
         self.tables_by_shard
             .insert(tables_of_shard.shard_info.id, tables_of_shard);
     }
+
+    fn try_insert_table_to_shard(
+        &mut self,
+        prev_shard_version: u64,
```
Rachelint, line 162: How about `type ShardVersion = u64`?

cluster/src/shard_tables_cache.rs
```diff
@@ -113,4 +156,82 @@ impl Inner {
...
+
+        let table = tables_of_shard.tables.iter().find(|v| v.id == new_table.id);
+        ensure!(
+            table.is_none(),
+            TableAlreadyExists {
+                msg: "the table to insert already exists",
+            }
+        );
+
+        // Update the tables of shard.
```
Rachelint, line 192: Update tables of shard.

cluster/src/shard_tables_cache.rs
ShiKaiWi: Actually, there are some difference between the two cases, so let's keep it.

cluster/src/shard_tables_cache.rs
ShiKaiWi: Ok, I will define this alias in meta/type.rs.

cluster/src/shard_tables_cache.rs
Rachelint: Ok.

cluster/src/shard_tables_cache.rs
```diff
@@ -55,6 +58,46 @@ impl ShardTablesCache {
...
+    /// Try to remove a table from the shard with a newer version.
+    ///
+    /// It will fail if:
+    ///  - the shard doesn't exist, or
+    ///  - the shard version in the cache is not equal to the provided
+    ///    `prev_shard_version`, or
+    ///  - the table doesn't exist.
+    pub fn try_remove_table_from_shard(
+        &self,
+        prev_shard_version: u64,
```
Rachelint, line 91: The same problem here.

cluster/src/lib.rs
```diff
@@ -59,23 +59,34 @@ pub enum Error {
...
+        "Shard version mismatch, shard_info:{:?}, expect version:{}.\nBacktrace:\n{}",
+        shard_info,
+        expect_version,
         backtrace
     ))]
-    SchemaNotFound {
-        schema_name: SchemaName,
+    ShardVersionMismatch {
+        shard_info: ShardInfo,
+        expect_version: u64,
```
Rachelint, line 89: Replace to ShardVersion?

Rachelint (approved):
LGTM

PR #329, opened 2022-10-21
feat: add async parquet reader bench

jiacai2050:

Closes #

# Rationale for this change

Perf is always good, add a new bench case for async reader.

Bench https://docs.rs/parquet/latest/parquet/arrow/async_reader/index.html

No
# How does this change test

Not required


ShiKaiWi (approved):
LGTM

PR #332, opened 2022-10-21
feat: separate status code between storage service and meta event service

ShiKaiWi:

Closes #

# Rationale for this change
Now the status code of storage service and meta event service is defined as one, but the two services have different status code set. 

- Separate the definition of error for storage service and meta event service.

None.

# How does this change test
Test by manual with CeresMeta.


server/src/route/mod.rs
```diff
@@ -5,14 +5,56 @@ use std::sync::Arc;
...
+    },
+
+    #[snafu(display("Failure caused by others, msg:{}, err:{}", msg, source))]
+    OtherWithCause {
+        msg: String,
+        source: Box<dyn std::error::Error + Sync + Send>,
+    },
+
+    #[snafu(display("Failure caused by others, msg:{}.\nBacktrace:\n{}", msg, backtrace))]
+    OtherNoCause { msg: String, backtrace: Backtrace },
```
Rachelint, line 54: I see `OtherNoCause` is unused now?

server/src/grpc/meta_event_service/error.rs
```diff
@@ -0,0 +1,70 @@
...
+                let err_string = source.to_string();
+                let first_line = error_util::first_line_in_error(&err_string);
+                format!("{}. Caused by: {}", msg, first_line)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum StatusCode {
```
Rachelint, line 50: I think we should explain why two status codes are needed. Because they are so similar and a bit confusing.

server/src/grpc/meta_event_service/error.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn build_err_header(err: Error) -> ResponseHeader {
+    ResponseHeader {
+        code: err.code() as u32,
+        error: err.error_message(),
+    }
+}
+
+pub fn build_ok_header() -> ResponseHeader {
+    ResponseHeader {
+        code: StatusCode::Ok as u32,
```
Rachelint, line 67: How about implementing a method like `as_u32()` to do this.

server/src/grpc/storage_service/mod.rs
```diff
@@ -587,35 +610,55 @@ fn build_schema_from_metric(schema_config: &SchemaConfig, metric: &WriteMetric)
...
-    schema_builder.build().context(BuildTableSchema {
-        metric: &metric.metric,
-    })
+    schema_builder
+        .build()
+        .map_err(|e| Box::new(e) as _)
+        .context(ErrWithCause {
+            code: StatusCode::BAD_REQUEST,
+            msg: "invalid schema to build",
+        })
```
Rachelint, line 661: invalid -> Invalid

server/src/grpc/storage_service/mod.rs
```diff
@@ -489,8 +506,9 @@ fn build_schema_from_metric(schema_config: &SchemaConfig, metric: &WriteMetric)
 
     ensure!(
         !write_entries.is_empty(),
-        InvalidArgument {
-            msg: format!("Empty write entires to write table:{}", table_name),
+        ErrNoCause {
+            code: StatusCode::BAD_REQUEST,
+            msg: format!("empty write entires to write table:{}", table_name),
```
Rachelint, line 511: empty -> Empty? It make me confused, the first character lowercase or uppercase?

ShiKaiWi: The style seems to keep the first character lowercase.

server/src/route/mod.rs
ShiKaiWi: It is used in `server/src/route/cluster_based.rs`.

server/src/grpc/storage_service/mod.rs
ShiKaiWi: No need to make the first word with capital letter.

server/src/grpc/meta_event_service/error.rs
ShiKaiWi: Done.

Rachelint (approved):

PR #333, opened 2022-10-21
fix: use right table manipulator when setup

ShiKaiWi:

Closes #

# Rationale for this change
 Now the table manipulator is hard coded when setup.

- Choose correct table manipulator when setup ceresdb-server.

None.

# How does this change test
Test by manual.


Rachelint (approved):
LGTM

PR #336, opened 2022-10-24
fix: create default schema automatically

ShiKaiWi:

Closes #

# Rationale for this change
 For now, the default schema won't be created when setting up ceresdb-server, which leads to the following failures.

- Create default schema when setting up ceresdb-server.

None.

# How does this change test
Test with meta.


ShiKaiWi:
The default schema is already created when setting up.

PR #337, opened 2022-10-24
refactor: make mq interface simpler.

Rachelint:

Closes #

# Rationale for this change
The race's checking should not be done in mq interface level, it should be done in high-level. So I refactor it to make it simpler.
 

Refactor the mq interface and its kafka implemetation.

None.

# How does this change test
By new and exist ut.

### DIFF
--- a/components/message_queue/src/kafka/kafka_impl.rs
+++ b/components/message_queue/src/kafka/kafka_impl.rs
@@ -2,7 +2,7 @@
 
 //! Kafka implementation's detail
 
-use std::{cmp::Ordering, collections::HashMap, fmt::Display, sync::Arc};
+use std::{collections::HashMap, fmt::Display, sync::Arc};
 
 use async_trait::async_trait;
 use common_util::define_result;
@@ -10,7 +10,7 @@ use futures::StreamExt;
 use log::info;
 use rskafka::{
     client::{
-        consumer::{StartOffset, StreamConsumer, StreamConsumerBuilder},
+        consumer::{StartOffset as KafkaStartOffset, StreamConsumer, StreamConsumerBuilder},
         controller::ControllerClient,
         error::{Error as RskafkaError, ProtocolError},
         partition::{Compression, OffsetAt, PartitionClient, UnknownTopicHandling},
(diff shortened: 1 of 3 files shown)


components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -249,22 +253,62 @@ impl MessageQueue for KafkaImpl {
             })?)
     }
 
-    async fn consume_all(&self, topic_name: &str) -> Result<KafkaConsumeIterator> {
-        info!("Need to consume all data in kafka topic:{}", topic_name);
+    async fn get_earliest_offset(&self, topic_name: &str) -> Result<Offset> {
```
ShiKaiWi, line 256: How about naming it `fetch_earliest_offset`? `Get` seems a lightweight operation.

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -249,22 +253,62 @@ impl MessageQueue for KafkaImpl {
...
+        topic_client
+            .get_offset(OffsetAt::Earliest)
+            .await
+            .context(FetchOffset {
+                topic_name: topic_name.to_string(),
+                offset_type: OffsetType::EarliestOffset,
+            })
+    }
 
+    async fn get_high_watermark(&self, topic_name: &str) -> Result<Offset> {
```
ShiKaiWi, line 274:
```suggestion
    async fn fetch_high_watermark(&self, topic_name: &str) -> Result<Offset> {
```

BTW, `get_earliest_offset` and `get_high_watermark` seems the same procedure except the different offsets.

components/message_queue/src/lib.rs
```diff
@@ -50,5 +61,27 @@ pub struct MessageAndOffset {
...
+    async fn next_message(&mut self) -> Result<(MessageAndOffset, Offset), Self::Error>;
+}
+
+/// At which position shall the stream start.
+#[derive(Debug, Clone, Copy)]
+pub enum StartOffset {
+    /// At the earlist known offset.
+    ///
+    /// This might be larger than 0 if some records were already deleted due to
+    /// a retention policy or delete records.
```
ShiKaiWi, line 73:
```suggestion
    /// a retention policy or delete operations.
```

components/message_queue/src/lib.rs
```diff
@@ -50,5 +61,27 @@ pub struct MessageAndOffset {
...
+    Earliest,
+
+    /// At the latest known offset.
+    ///
+    /// This is helpful if you only want ot process new data.
```
ShiKaiWi, line 78:
```suggestion
    /// This is helpful if you only want to process new data.
```

components/message_queue/src/lib.rs

components/message_queue/src/lib.rs

components/message_queue/src/kafka/kafka_impl.rs
Rachelint: Done.

ShiKaiWi (approved):
LGTM

PR #338, opened 2022-10-24
fix: make open shard idempotent

ShiKaiWi:

Closes #

# Rationale for this change
 For now, opening a shard is not idempotent, which is not expected by CeresMeta.

- Make opening a shard idempotent.

None.

# How does this change test
Tested with CeresMeta.


Rachelint (approved):
LGTM

PR #339, opened 2022-10-24
fix: add missing version for meta request

ShiKaiWi:

Closes #

# Rationale for this change
The shard version in the heartbeat is not set now, which leading to wrong message to CeresMeta.

- Set the correct shard version in the message to CeresMeta;
- Add log for handling requests from CeresMeta;

None.

# How does this change test
No need to test.


chunshao90 (approved):
LGTM

PR #340, opened 2022-10-24
edit: column order use user define

### LINKED ISSUE #187: Column order is not what user define when create table
### Describe this problem

Table' columns order is not what user define when create table

### Steps to reproduce

Execute those SQL
```sql
CREATE TABLE test_order (
    host string tag,
    v2 int,
    ts timestamp NOT NULL,
    v1 int,
    timestamp KEY (ts)
)ENGINE = Analytic WITH (
    enable_ttl = 'false'
);

desc test_order;
```
`desc` will output 

```
ts	timestamp	1	0	0
tsid	uint64	1	0	0
host	string	0	1	1
v1	int	0	1	1
v2	int	0	1	0
```

### Expected behavior

The column order should be same with the order when created.

### Additional Information

https://github.com/CeresDB/ceresdb/blob/3e826ffbb98d2c5351e63b28dd9dd4644e7d4c1e/sql/src/planner.rs#L265

The root cause is currently we use a BTreeMap when build create table plan, it should use a vector to keep the ordering

ref #154

dust1:

Closes #187 

# Rationale for this change
 
Table' columns order is not what user define when create table

I modified the 'add_key_column' and 'add_normal_column' methods in 'schema.builder' and added 'primary_key_index' to record additional field primary key information.

But I had a serialization problem, so the job wasn't done.

In my understanding, these changes are sufficient, but when I test the new table, I find that the primary_key_index is empty in the data re-read by wal log. I need more time to check the reason

# How does this change test

### DIFF
--- a/common_types/src/schema.rs
+++ b/common_types/src/schema.rs
@@ -511,6 +511,8 @@ pub struct Schema {
     arrow_schema: ArrowSchemaRef,
     /// The number of primary key columns
     num_key_columns: usize,
+    /// The primary key index list in columns
+    primary_key_index: Vec<usize>,
     /// Index of timestamp key column
     // TODO(yingwen): Maybe we can remove the restriction that timestamp column must exists in
     //  schema (mainly for projected schema)
@@ -537,6 +539,7 @@ impl fmt::Debug for Schema {
             .field("enable_tsid_primary_key", &self.enable_tsid_primary_key)
             .field("column_schemas", &self.column_schemas)
             .field("version", &self.version)
+            .field("primary_key_index", &self.primary_key_index)
             .finish()
     }
 }
@@ -564,13 +567,36 @@ impl Schema {
     }
 
     /// Returns an immutable reference of the key column vector.
-    pub fn key_columns(&self) -> &[ColumnSchema] {
-        &self.columns()[..self.num_key_columns]
+    pub fn key_columns(&self) -> Vec<ColumnSchema> {
+        let columns = self.column_schemas.columns();
+        let mut result = vec![];
+        for (idx, col) in columns.iter().enumerate() {
+            if idx == self.timestamp_index {
+                result.push(col.clone());
+                continue;
+            }
+
+            if self.primary_key_index.contains(&idx) {
+                result.push(col.clone());
+            }
+        }
+        result
     }
 
     /// Returns an immutable reference of the normal column vector.
-    pub fn normal_columns(&self) -> &[ColumnSchema] {
-        &self.columns()[self.num_key_columns..]
+    pub fn normal_columns(&self) -> Vec<ColumnSchema> {
+        let columns = self.column_schemas.columns();
+        let mut result = vec![];
+        for (idx, col) in columns.iter().enumerate() {
+            if idx == self.timestamp_index {
+                continue;
+            }
+
+            if !self.primary_key_index.contains(&idx) {
+                result.push(col.clone());
+            }
+        }
+        result
     }
 
     /// Returns index of the tsid column.
@@ -824,12 +850,17 @@ impl TryFrom<common_pb::TableSchema> for Schema {
         let mut builder = Builder::with_capacity(schema.columns.len())
             .version(schema.version)
             .enable_tsid_primary_key(schema.enable_tsid_primary_key);
+        let mut primary_key_index = schema.primary_key_index;
+        if primary_key_index.is_empty() && schema.num_key_columns > 0 {
+            for i in 0..schema.num_key_columns {
+                primary_key_index.push(i as u64);
+            }
+        }
 
         for (i, column_schema_pb) in schema.columns.into_iter().enumerate() {
             let column =
                 ColumnSchema::try_from(column_schema_pb).context(ColumnSchemaDeserializeFailed)?;
-
-            if i < schema.num_key_columns as usize {
+            if primary_key_index.contains(&(i as u64)) {
                 builder = builder.add_key_column(column)?;
             } else {
                 builder = builder.add_normal_column(column)?;
@@ -850,6 +881,8 @@ impl From<Schema> for common_pb::TableSchema {
             table_schema.columns.push(column_schema);
         }
 
+        table_schema.primary_key_index =
+            schema.primary_key_index.iter().map(|i| *i as u64).collect();
         table_schema.num_key_columns = schema.num_key_columns as u32;
         table_schema.timestamp_index = schema.timestamp_index as u32;
         table_schema.enable_tsid_primary_key = schema.enable_tsid_primary_key;
@@ -865,6 +898,7 @@ pub struct Builder {
     columns: Vec<ColumnSchema>,
     /// The number of primary key columns
     num_key_columns: usize,
+    primary_key_index: Vec<usize>,
     /// Timestamp column index
     timestamp_index: Option<usize>,
     column_names: HashSet<String>,
@@ -895,6 +929,7 @@ impl Builder {
         Self {
             columns: Vec::with_capacity(capacity),
             num_key_columns: 0,
+            primary_key_index: Vec::new(),
             timestamp_index: None,
             column_names: HashSet::with_capacity(capacity),
             column_ids: HashSet::with_capacity(capacity),
@@ -923,9 +958,10 @@ impl Builder {
                     given_column: column.name,
                 }
             );
-            self.timestamp_index = Some(self.num_key_columns);
+            self.timestamp_index = Some(self.columns.len());
         }
 
+        self.primary_key_index.push(self.columns.len());
         self.insert_new_key_column(column);
 
         Ok(self)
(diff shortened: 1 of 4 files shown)


dust1:
> 

ok

ShiKaiWi:
> In my understanding, these changes are sufficient, but when I test the new table, I find that the primary_key_index is empty in the data re-read by wal log. I need more time to check the reason

@dust1 Have you figured out this problem? I can help do some troubleshooting if you will. Don't worry about the time because this pr is not urgent for us.

dust1:
> > In my understanding, these changes are sufficient, but when I test the new table, I find that the primary_key_index is empty in the data re-read by wal log. I need more time to check the reason
> 
> @dust1 Have you figured out this problem? I can help do some troubleshooting if you will. Don't worry about the time because this pr is not urgent for us.

thank you @ShiKaiWi 🥰. Maybe I've already done this pr

common_types/src/schema.rs
```diff
@@ -534,8 +543,8 @@ pub struct Schema {
     /// The underlying arrow schema, data type of fields must be supported by
     /// datum
     arrow_schema: ArrowSchemaRef,
-    /// The number of primary key columns
-    num_key_columns: usize,
```
jiacai2050, line 538: primary_key_indexes

common_types/src/schema.rs
```diff
@@ -556,12 +565,12 @@ impl fmt::Debug for Schema {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Schema")
             // arrow_schema is ignored.
-            .field("num_key_columns", &self.num_key_columns)
             .field("timestamp_index", &self.timestamp_index)
             .field("tsid_index", &self.tsid_index)
             .field("enable_tsid_primary_key", &self.enable_tsid_primary_key)
             .field("column_schemas", &self.column_schemas)
             .field("version", &self.version)
+            .field("primary_key_index", &self.primary_key_index)
```
jiacai2050, line 573: primary_key_indexes

common_types/src/schema.rs
```diff
@@ -589,13 +598,36 @@ impl Schema {
     }
 
     /// Returns an immutable reference of the key column vector.
-    pub fn key_columns(&self) -> &[ColumnSchema] {
-        &self.columns()[..self.num_key_columns]
+    pub fn key_columns(&self) -> Vec<ColumnSchema> {
+        let columns = self.column_schemas.columns();
+        let mut result = vec![];
+        for (idx, col) in columns.iter().enumerate() {
+            if idx == self.timestamp_index {
```
jiacai2050, line 605: Isn't timestamp_index already in `primary_key_index`?

common_types/src/schema.rs
```diff
@@ -617,6 +649,11 @@ impl Schema {
         self.column_schemas.num_columns()
     }
 
+    /// Returns true if idx is primary key idnex
+    pub fn contains_primary_key_index(&self, idx: &usize) -> bool {
```
jiacai2050, line 653: is_primary_key_index

common_types/src/schema.rs
```diff
@@ -652,10 +689,9 @@ impl Schema {
         self.column_schemas.index_of(name)
     }
 
-    /// Returns the number of columns in primary key
-    #[inline]
-    pub fn num_key_columns(&self) -> usize {
-        self.num_key_columns
+    /// Return the number of columns index in primary key
+    pub fn primary_key_idx(&self) -> &[usize] {
```
jiacai2050, line 693: primary_key_indexes

common_types/src/schema.rs
```diff
@@ -849,12 +881,17 @@ impl TryFrom<common_pb::TableSchema> for Schema {
         let mut builder = Builder::with_capacity(schema.columns.len())
             .version(schema.version)
             .enable_tsid_primary_key(schema.enable_tsid_primary_key);
+        let mut primary_key_index = schema.primary_key_index;
+        if primary_key_index.is_empty() && schema.num_key_columns > 0 {
```
jiacai2050, line 885: No need to keep compatibility, we are still before 1.0, so break change is acceptable now.

common_types/src/schema.rs
```diff
@@ -1135,7 +1183,26 @@ impl Builder {
 
     /// Build the schema
     pub fn build(self) -> Result<Schema> {
-        let timestamp_index = self.timestamp_index.context(MissingTimestampKey)?;
+        let retry_timestamp_key = if self.timestamp_index.is_none() {
```
jiacai2050, line 1186: if timestamp_index is not set, we should return an Error just like before.

proto/protos/common.proto
```diff
@@ -44,16 +44,18 @@ message ColumnSchema {
...
+    // Version of the schema
+    uint32 version = 2;
+    // Key column num
+    uint32 num_key_columns = 3;
+    // Timestamp index in columns
+    uint32 timestamp_index = 4;
+    // Enable auto generated tsid as primary key
+    bool enable_tsid_primary_key = 5;
+    // Primary key index
+    repeated uint64 primary_key_index = 6;
```
jiacai2050, line 58: primary_key_indexes

jiacai2050 (changes requested):

sql/src/planner.rs
```diff
@@ -397,46 +417,39 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
...
 
-        // Build primary key, the builder will check timestamp column is in primary key.
-        if let Some(idx) = primary_key_constraint_idx {
-            // If primary key is already provided, use that primary key.
-            if let TableConstraint::Unique { columns, .. } = &stmt.constraints[idx] {
-                for col in columns {
-                    let key_column = if TSID_COLUMN == col.value {
+        let timestamp_col_idx = timestamp_column_idx.context(RequireTimestamp)?;
+        // The key columns have been consumed.
+        for (idx, col) in stmt
```
jiacai2050, line 422: This iter is not required, it's already collected in `name_column_index_map`

common_types/src/schema.rs
jiacai2050: `num_key_columns` should be removed, and `insert_new_key_column` & `insert_new_normal_column ` could be merged into one method.

jiacai2050 (changes requested):

sql/src/planner.rs
```diff
@@ -397,46 +417,39 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
...
+            .columns
+            .iter()
+            .enumerate()
+            .map(|(idx, column)| Ok((idx, parse_column(column)?)))
+            .collect::<Result<Vec<_>>>()?
+        {
+            if !primary_key_column_idxs.is_empty() {
```
jiacai2050, line 429: After insert_new_key_column & insert_new_normal_column is merged into one method, I think this place could be simplified further.

sql/src/planner.rs
```diff
@@ -397,49 +417,39 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
...
-                        name_column_map
-                            .remove(&*col.value)
-                            .with_context(|| PrimaryKeyNotFound {
-                                name: col.value.clone(),
-                            })?
-                    };
-                    // The schema builder will checks there is only one timestamp column in primary
-                    // key.
+        let timestamp_col_idx = timestamp_column_idx.context(RequireTimestamp)?;
+        // The key columns have been consumed.
```
jiacai2050, line 421: Is this still right?
Loop below still use key columns

sql/src/planner.rs
```diff
@@ -379,7 +395,11 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
                         timestamp_column.data_type == DatumKind::Timestamp,
                         InvalidTimestampKey
                     );
+                    let column_idx = name_column_index_map
+                        .get(name as &str)
+                        .context(TimestampColumnNotFound { name })?;
 
+                    timestamp_column_idx = Some(*column_idx);
```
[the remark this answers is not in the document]
jiacai2050: Is this addressed?

common_types/src/schema.rs
```diff
@@ -457,7 +457,7 @@ impl TryFrom<ArrowSchemaRef> for RecordSchema {
 #[derive(Debug, Clone, PartialEq)]
 pub struct RecordSchemaWithKey {
     record_schema: RecordSchema,
-    num_key_columns: usize,
+    key_index: Vec<usize>,
```
jiacai2050, line 460: primary_key_indexes

common_types/src/schema.rs
```diff
@@ -873,22 +909,28 @@ impl From<&Schema> for common_pb::TableSchema {
...
     }
 }
 
 /// Schema builder
 #[must_use]
 pub struct Builder {
     columns: Vec<ColumnSchema>,
-    /// The number of primary key columns
-    num_key_columns: usize,
+    /// The indexes with primary key columns
```
jiacai2050, line 932: The indexes of...

jiacai2050 (changes requested):

sql/src/planner.rs
dust1: I need to keep the order of the columns in the loop, so I have to loop through the columns in the `stmt.column`.
the `timestamp_col_idx `  is evaluated in the loop against the column index and a timestamp key is added.

Issue #284, opened 2022-10-08
Implement wal manager on kafka


Rachelint:
**Description**
After modifying the existing wal desigin, we can implement our wal on kafka now. 

**Proposal**
+ Define and implement the log record format.
+ Implement the basic region management, and read, write on region (region's recovery is not considered yet).
+ Define the meta record format and implement region's recovery.
    + The meta in mq implement will includes a shard's all tables' max sequence number.
    + The meta key should be able to be distinguished from normal logs.
    + Before pushing the meta messages, the partition should be frozen firstly.
    + While recovering the region, it will perform the reverse traversal on partition util find one meta log.

**Additional context**
The work is a bit big, I spit it into three stage:

- [x] implement the region meta module.
- [x] define record format .
- [ ] implement record deletion, record reading、record writing method.
- [ ] implement the region manager, region recovery,  and add some config.

PR #345, opened 2022-10-27
feat:  region meta in wal manager on kafka

Rachelint:

Part of #284 

# Rationale for this change
 

See  #284.

+ The record format used in wal on message queue.
+ The region meta module.

None.

# How does this change test
Will add test later.

### DIFF
--- a/wal/src/message_queue_impl/region_meta.rs
+++ b/wal/src/message_queue_impl/region_meta.rs
@@ -2,11 +2,7 @@
 
 //! Region meta data
 
-use std::{
-    cmp,
-    collections::{BTreeMap, HashMap},
-    sync::atomic::{AtomicI64, Ordering},
-};
+use std::collections::{BTreeMap, HashMap};
 
 use common_types::{table::TableId, SequenceNumber};
 use common_util::define_result;
@@ -107,10 +103,6 @@ impl RegionMeta {
             .update_after_write(write_offset_range, updated_num)
             .await;
 
-        inner
-            .local_latest_offset
-            .fetch_add(updated_num as i64, Ordering::Relaxed);
-
         Ok(())
     }
 
@@ -132,36 +124,6 @@ impl RegionMeta {
         Ok(())
     }
 
-    /// Scan the table meta entry in it and get the safe(minimum) offset among
-    /// them to return.
-    ///
-    /// NOTICE: Need to freeze the whole region meta on high-level before
-    /// calling.
-    pub async fn get_safe_delete_offset(&self) -> Result<Offset> {
-        let inner = self.inner.read().await;
-        let mut safe_delete_offset = Offset::MAX;
-        let mut high_watermark = 0;
-        // Calc the min offset in message queue.
-        for table_meta in inner.table_metas.values() {
-            let meta_data = table_meta.get_meta_data().await?;
-            if let Some(offset) = meta_data.safe_delete_offset {
-                safe_delete_offset = cmp::min(safe_delete_offset, offset);
-            }
-            high_watermark = cmp::max(high_watermark, meta_data.current_high_watermark);
-        }
-
-        if safe_delete_offset == Offset::MAX {
-            // All tables are in such states:
-            // + has init, but not written
-            // + has written, but not flushed
-            // + has flushed, but not written again
-            // So, we can directly delete it up to the high_watermark.
-            Ok(high_watermark)
-        } else {
-            Ok(safe_delete_offset)
-        }
-    }
-
     /// Scan the table meta entry in it and get the snapshot about tables' next
     /// sequences.
     ///
@@ -176,10 +138,7 @@ impl RegionMeta {
             entries.push(meta_data);
         }
 
-        Ok(RegionMetaSnapshot {
-            next_offset: inner.local_latest_offset.load(Ordering::Relaxed),
-            entries,
-        })
+        Ok(RegionMetaSnapshot { entries })
     }
 }
 
@@ -187,9 +146,6 @@ impl RegionMeta {
 #[derive(Default, Debug)]
 struct RegionMetaInner {
     table_metas: HashMap<TableId, TableMeta>,
-    /// It will fall behind the high watermark for ensuring successive
-    /// increment.
-    local_latest_offset: AtomicI64,
 }
 
 /// Wrapper for the [TableMetaInner].
@@ -320,7 +276,6 @@ pub struct TableMetaData {
 ///
 /// Include all tables(of current shard) and their next sequence number.
 pub struct RegionMetaSnapshot {
-    pub next_offset: Offset,
     pub entries: Vec<TableMetaData>,
 }
 
--- a/wal/src/message_queue_impl/region_meta.rs
+++ b/wal/src/message_queue_impl/region_meta.rs
@@ -56,7 +56,7 @@ define_result!(Error);
 
 // TODO: will be made use of later.
 #[allow(unused)]
-#[derive(Default)]
+#[derive(Default, Debug)]
 pub struct RegionMeta {
     inner: RwLock<RegionMetaInner>,
 }
@@ -64,9 +64,10 @@ pub struct RegionMeta {
 // TODO: will be made use of later.
 #[allow(unused)]
 impl RegionMeta {
-     // TODO: Need to implement the init method using the [RegionMetaSnapshot] which will be persisted.
+    // TODO: Need to implement the init method using the [RegionMetaSnapshot] which
+    // will be persisted.
 
-     pub async fn prepare_for_table_write(&self, table_id: TableId) -> Result<SequenceNumber> {
+    pub async fn prepare_for_table_write(&self, table_id: TableId) -> Result<SequenceNumber> {
         {
             let inner = self.inner.read().await;
             if let Some(table_meta) = inner.table_metas.get(&table_id) {
@@ -88,8 +89,8 @@ impl RegionMeta {
     pub async fn update_after_table_write(
         &self,
         table_id: TableId,
-        next_sequence_num: SequenceNumber,
-        offset: Offset,
+        updated_start_offset: Offset,
+        updated_num: u64,
     ) -> Result<()> {
         let inner = self.inner.read().await;
         let table_meta = inner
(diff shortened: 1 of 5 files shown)


wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,595 @@
...
+            .local_latest_offset
+            .fetch_add(updated_num as i64, Ordering::Relaxed);
+
+        Ok(())
+    }
+
+    pub async fn mark_table_deleted(
+        &self,
+        table_id: TableId,
+        next_sequence_num: SequenceNumber,
```
ShiKaiWi, line 120: Maybe it is better to name this parameter as `sequence_num`.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,595 @@
...
+            high_watermark = cmp::max(high_watermark, meta_data.current_high_watermark);
+        }
+
+        if safe_delete_offset == Offset::MAX {
+            // All tables are in such states:
+            // + has init, but not written
+            // + has written, but not flushed
+            // + has flushed, but not written again
+            // So, we can directly delete it up to the high_watermark.
+            Ok(high_watermark)
```
ShiKaiWi, line 159: Can't understand why it is safe to delete up to high_watermark if data of tables has written but not flushed.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,595 @@
...
+        let _ = inner
+            .start_sequence_offset_mapping
+            .insert(old_next_sequence_num, write_offset_range.0);
+        inner.current_high_watermark = write_offset_range.1 + 1;
+    }
+
+    async fn mark_deleted(&self, latest_marked_deleted: SequenceNumber) {
+        let mut inner = self.inner.lock().await;
+        assert!(latest_marked_deleted <= inner.next_sequence_num);
+        inner.latest_marked_deleted = latest_marked_deleted;
```
ShiKaiWi, line 236: Shall we avoid updating to a smaller `latest_marked_deleted`?

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,595 @@
...
+                next_sequence_num: inner.next_sequence_num,
+                latest_marked_deleted: inner.latest_marked_deleted,
+                current_high_watermark: inner.current_high_watermark,
+                safe_delete_offset: None,
+            })
+        } else {
+            let offset = inner
+                .start_sequence_offset_mapping
+                .get(&inner.latest_marked_deleted);
+            ensure!(inner.next_sequence_num > inner.latest_marked_deleted && offset.is_some(),
```
ShiKaiWi, line 264: It will be more clear for troubleshooting if we divide this check into two parts: 
```
ensure!(inner.next_sequence_num > inner.latest_marked_deleted, ...);
let offset = inner..start_sequence_offset_mapping
                .get(&inner.latest_marked_deleted)
                .context()?;
```
```

File: wal/src/message_queue_impl/region_meta.rs [hunk 5]
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    async fn get_meta_data(&self) -> Result<TableMetaData> {
+        let inner = self.inner.lock().await;
+
+        // Only two situations exist:
+        // + no log of the table has ever been written
+        //  (next sequence num == latest marked deleted).
+        // + all logs of the table can be deleted
+        //  (next_sequence_num > latest_marked_deleted).
```
ShiKaiWi, line 251: This comment seems confusing to me -- Why are all the logs of table able to be deleted when `next_sequence_num > latest_marked_deleted`.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,595 @@
...
+            .is_none());
+        // New table, so returned next sequence num is zero.
+        Ok(SequenceNumber::MIN)
+    }
+
+    pub async fn update_after_table_write(
+        &self,
+        table_id: TableId,
+        write_offset_range: (Offset, Offset),
+        updated_num: u64,
```
ShiKaiWi, line 93: Maybe we should define a struct call OffsetRange with specific field name: `start` & `end`. And the `update_num` can be replaced by its method called `len`.

wal/src/message_queue_impl/region_meta.rs
Rachelint: The comment has problems indeed...
The high watermark should be returned just while all tables are flushed but not written again and just initialized.

I have removed this function, it should be the logic in cleaner and can be performed using the data in snapshot.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,595 @@
...
+            let inner = self.inner.read().await;
+            if let Some(table_meta) = inner.table_metas.get(&table_id) {
+                return Ok(table_meta.get_meta_data().await?.next_sequence_num);
+            }
+        }
+
+        // Double check is not needed, due to just one task will write the specific
+        // table.
+        let mut inner = self.inner.write().await;
+        debug_assert!(inner
```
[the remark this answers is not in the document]
Rachelint: I propose to keep the logic and just add an assert now.
I want to let making it adapt to other thread model be an enhancement which will be implemented later. Because for reaching it, the key task should be modifying the lock granularity but not the logic here. As I see, it may be not so suitable to do in one pr.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,587 @@
...
+
+    /// Update following meta data of table after each writing:
+    /// + mapping of start sequence number to start offset
+    /// + high watermark
+    /// + next sequence number
+    pub async fn update_after_table_write(
+        &self,
+        table_id: TableId,
+        write_offset_range: OffsetRange,
+        updated_num: u64,
```
ShiKaiWi, line 82: Is `updated_num` redundent? Because it can be computed by `(range.end - range.start + 1)`.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,587 @@
...
+    ) -> Result<()> {
+        let inner = self.inner.read().await;
+        let table_meta = inner
+            .table_metas
+            .get(&table_id)
+            .with_context(|| MarkDeleted {
+                table_id,
+                msg: format!("table:{}'s meta not found while mark its deleted", table_id),
+            })?;
+        table_meta.mark_deleted(sequence_num + 1).await;
```
ShiKaiWi, line 119: Why should we add 1 to the `sequence_num`?

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,587 @@
...
+        // Update the mapping and high water mark.
+        let _ = inner
+            .start_sequence_offset_mapping
+            .insert(old_next_sequence_num, write_offset_range.start);
+        inner.current_high_watermark = write_offset_range.end + 1;
+    }
+
+    async fn mark_deleted(&self, latest_marked_deleted: SequenceNumber) {
+        let mut inner = self.inner.lock().await;
+        assert!(
```
ShiKaiWi, line 188: For this case, maybe we should return an error rather than panic.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,587 @@
...
+            latest_marked_deleted <= inner.next_sequence_num,
+            "latest marked deleted should be less than or 
+            equal to next sequence number, now are:{} and {}",
+            latest_marked_deleted,
+            inner.next_sequence_num
+        );
+        assert!(
```
ShiKaiWi, line 195: For this case, how about just leaving a warning log and then returning?

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,587 @@
...
+}
+
+/// Message queue implementation's meta value.
+///
+/// Include all tables(of current shard) and their next sequence number.
+pub struct RegionMetaSnapshot {
+    pub entries: Vec<TableMetaData>,
+}
+
+pub struct OffsetRange {
```
ShiKaiWi, line 299: Add comments to explain that the end offset is inclusive. BTW, it can't express an empty range if we make the end offset inclusive, and is that OK?

wal/src/message_queue_impl/region_meta.rs
Rachelint: > Is `updated_num` redundent? Because it can be computed by `(range.end - range.start + 1)`.

I think we should not assume that the returned offsets are continuous ?
Although it is indeed now.

wal/src/message_queue_impl/region_meta.rs
Rachelint: Maybe following error, we should return to tell : 
+ returned from the third party or os
+ returned from other module
+ caused by Distributed Inconsistency

And the pure logic error, assert maybe proper?

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,606 @@
...
+        msg,
+        backtrace
+    ))]
+    MarkDeleted {
+        table_id: TableId,
+        msg: String,
+        backtrace: Backtrace,
+    },
+}
+define_result!(Error);
```
ShiKaiWi, line 39: Missing one newline.

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -0,0 +1,606 @@
...
+
+        Ok(())
+    }
+
+    /// Scan the table meta entry in it and get the snapshot about tables' next
+    /// sequences.
+    ///
+    /// NOTICE: Need to freeze the whole region meta on high-level before
+    /// calling.
+    pub async fn get_snapshot(&self) -> RegionMetaSnapshot {
```
ShiKaiWi, line 132: How about replacing `get` with `making` because it's not a trivial thing to build the snapshot: 
```suggestion
    pub async fn make_snapshot(&self) -> RegionMetaSnapshot {
```

Rachelint: Done.

wal/src/message_queue_impl/region_meta.rs
Rachelint: Done.

ShiKaiWi (approved):
LGTM

PR #348, opened 2022-10-27
chore: fix error message when table is not found

ShiKaiWi:

Closes #

# Rationale for this change
The error message is not readable when table is not found.

- Refactor the message for table is not found.

The error message will be different.

# How does this change test
By the harness test.


chunshao90 (approved):
LGTM

PR #352, opened 2022-10-28
Refactor replace protobuf with prost

### LINKED ISSUE #346: Replace protobuf with prost used by protos crate so to remove grpcio dependencies
### Describe This Problem

Current the `protos` crate uses the `protobuf` dependency to generate codes, but it introduces a huge dependency `grpcio`, which makes the binary size larger and compiling time longer.

### Proposal

Let's replace `protobuf` crate with the `prost` crate.

### Additional Context

This may be helpful for #294

ShiKaiWi:

Closes #346

# Rationale for this change
 See the #346

- Replace the protobuf with prost;
- Modify the `MemBuf`/`MemBufMut` trait to extenstion for `Buf`/`BufMut` because the prost's apis depend on `BufMut`/`Buf` and such modification (`impl<T> MemBuf for T where Buf`) is more elegant (some codes can be removed);
- Format the proto files.

None.

# How does this change test
Existing UT.


proto/protos/analytic_common.proto
```diff
@@ -6,68 +6,68 @@ package analytic_common;
 
 // Options of a table that need to persist
 message TableOptions {
-    // Segment duration in ms.
```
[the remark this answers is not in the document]
ShiKaiWi: Yes. I will file an issue for this.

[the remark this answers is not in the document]
ShiKaiWi: Currently, `clang-format -style=google` is used, and it is applied to ceresdbproto repo.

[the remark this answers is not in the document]
ShiKaiWi: #353 is filed for this.

rustfmt.toml
```diff
@@ -9,4 +9,8 @@ format_code_in_doc_comments = true
 # Reorder impl items. type and const are put first, then macros and methods.
 reorder_impl_items = true
 # Discard existing import groups, and create three groups for std, external crates, crates
-group_imports = "StdExternalCrate"
\ No newline at end of file
+group_imports = "StdExternalCrate"
+
+ignore = [
+    "proto",
+]
```
Rachelint, line 16: add a new line?

common_types/src/datum.rs
```diff
@@ -269,43 +269,43 @@ impl TryFrom<u8> for DatumKind {
...
-            DatumKind::UInt64 => Self::UINT64,
-            DatumKind::UInt32 => Self::UINT32,
-            DatumKind::UInt16 => Self::UINT16,
-            DatumKind::UInt8 => Self::UINT8,
-            DatumKind::Int64 => Self::INT64,
-            DatumKind::Int32 => Self::INT32,
-            DatumKind::Int16 => Self::INT16,
-            DatumKind::Int8 => Self::INT8,
-            DatumKind::Boolean => Self::BOOL,
+            DatumKind::Null => Self::Null,
```
Rachelint, line 272: Upper case in pb, and Mixed case after compiling. It seems a bit strange...?

ShiKaiWi: It is indeed strange, but I prefer this naming style, which is uniform with the enum naming in the Rust.

Rachelint: Ok, indeed I have felt strange before, because upper case usually used by constant in rust.

PR #355, opened 2022-10-31
feat: add async parquet reader

jiacai2050:

Closes #

# Rationale for this change
In #312, performance has regression, so we can't replace old reader with the async reader, this PR add an async reader beside the old reader, which can be chosen by environment below
```
export ASYNC_PARQUET_READER=true
```

Add a new parquet reader based on DataFusion's ParquetExec, which use `ParquetRecordBatchStream` to parse parquet file.

No

# How does this change test

UT `test_parquet_build_and_read`


ShiKaiWi (approved):
LGTM

PR #358, opened 2022-11-01
feat: encoding in wal's message queue impl.

Rachelint:

Part of #284 

# Rationale for this change

see #284  

+ Add common log key and value for all wal implementations to use.
+ Add message queue's meta key and value.

None.

# How does this change test
Test by ut.

### DIFF
--- a/wal/src/message_queue_impl/encoding.rs
+++ b/wal/src/message_queue_impl/encoding.rs
@@ -1,0 +1,408 @@
+use common_types::{
+    bytes::{self, BytesMut, SafeBuf, SafeBufMut, Buf, BufMut},
+    table::TableId,
+    SequenceNumber,
+};
+use common_util::{
+    codec::{Decoder, Encoder},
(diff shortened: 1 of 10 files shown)


wal/src/message_queue_impl/encoding.rs
```diff
@@ -0,0 +1,427 @@
...
+    fn estimate_encoded_size(&self, _log_key: &MetaKey) -> usize {
+        // Refer to key format.
+        1 + 1 + 8
+    }
+}
+
+impl Decoder<MetaKey> for MetaKeyEncoder {
+    type Error = Error;
+
+    fn decode<B: SafeBuf>(&self, buf: &mut B) -> Result<MetaKey> {
```
ShiKaiWi, line 231: I guess that the B's trait bound should only be `Buf`?

ShiKaiWi (approved):
LGTM

PR #361, opened 2022-11-02
feat: add tools sst-convert

jiacai2050:

Closes #

# Rationale for this change
 
For now, we need to test different sst options to determine which is best for performance, but there is no fast way to convert existing sst to other options, so this PR add a cli to do this job.  

Add tools workspace member, and there is one cli: `sst-convert`, 
```
cargo run --bin sst-convert -- -h

tools 0.4.0
CeresDB Authors <ceresdbservice@gmail.com>

USAGE:
    sst-convert [OPTIONS] --store-path <STORE_PATH> --input <INPUT> --output <OUTPUT>

OPTIONS:
    -b, --batch-size <BATCH_SIZE>      Row group size of new sst file [default: 8192]
    -c, --compression <COMPRESSION>    Compression of new sst file(values:
                                       uncompressed/lz4/snappy/zstd) [default: uncompressed]
    -h, --help                         Print help information
    -i, --input <INPUT>                Input sst file(relative to store_path)
    -o, --output <OUTPUT>              Output sst file(relative to store_path)
    -s, --store-path <STORE_PATH>      Root dir of storage
    -V, --version                      Print version information
```

No

# How does this change test

Manually

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,6 +129,9 @@ name = "anyhow"
 version = "1.0.59"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c91f1f46651137be86f3a2b9a8359f9ab421d04d941c62b5982e1ca21113adf9"
+dependencies = [
+ "backtrace",
+]
 
 [[package]]
 name = "arc-swap"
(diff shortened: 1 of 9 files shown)


ShiKaiWi (approved):
LGTM

PR #362, opened 2022-11-02
feat: region in wal's message queue impl.  

### LINKED ISSUE #367: feat: implement region's delete, read、write method
**Description**
After modifying the existing wal desigin, we can implement our wal on kafka now. 

**Proposal**
The implementation will be split into following three phase:
+ read, write, mark deleted.
+ sync region meta snapshot.
+ clean old logs.

##### Read, write, mark deleted
For the region(highest level lock):
+ read and write can be concurrent(read lock).
+ mark deleted is exclusive to them(write lock, using write lock to lock the whole process is not so reasonable, should optimize later).
+ after marking deleted, a region meta snapshot sync will be triggered(it is why we use write lock in marking deleted).

##### Sync region meta snapshot
+ temporarily, we just sync it after mark delete and clean outdated logs.

**Additional context**
Part of #284

Rachelint:

Part of #284 
Close #367 

# Rationale for this change

see #284  

+ add region module in wal's message queue implementation.
+ add tests for it.

None.

# How does this change test
Test by ut.


wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+    encoding::{MetaEncoding, MetaKey},
+    region_meta::RegionMetaSnapshot,
+    to_message,
+};
+use crate::manager::RegionId;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display(
+        "Failed to sync snapshot of region, region id:{}, topic:{}, err:{}",
```
chunshao90, line 23: Remove `Failed to`.

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    pub async fn maybe_clean_logs(&self, snapshot: &RegionMetaSnapshot) -> Result<()> {
+        info!(
+            "Begin to check and clean logs, region id:{} snapshot:{:?}, topic:{}",
+            self.region_id, snapshot, self.log_topic
+        );
+
+        // Get prepare delete to offset.
```
chunshao90, line 163: The comment is confusing.

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+                    Some(last_deleted_offset)
+                } else {
+                    None
+                }
+            }
+
+            None => Some(safe_delete_offset),
+        };
+
+        // Delete to and sync snapshot
```
chunshao90, line 182: The comment is confusing.

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+        let log_topic = format_wal_data_topic_name(namespace, region_id);
+        let meta_topic = format_wal_meta_topic_name(namespace, region_id);
+        message_queue
+            .create_topic_if_not_exist(&log_topic)
+            .await
+            .map_err(|e| Box::new(e) as _)
+            .context(Open {
+                namespace,
+                region_id,
+            });
```
chunshao90, line 156:
```suggestion
        message_queue
            .create_topic_if_not_exist(&log_topic)
            .await
            .map_err(|e| Box::new(e) as _)
            .context(Open {
                namespace,
                region_id,
            })?;
```

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+        message_queue
+            .create_topic_if_not_exist(&meta_topic)
+            .await
+            .map_err(|e| Box::new(e) as _)
+            .context(Open {
+                namespace,
+                region_id,
+            });
```
chunshao90, line 164:
```suggestion
        message_queue
            .create_topic_if_not_exist(&meta_topic)
            .await
            .map_err(|e| Box::new(e) as _)
            .context(Open {
                namespace,
                region_id,
            })?;
```

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+                .await
+                .unwrap();
+
+            (
+                inner.make_meta_snapshot().await,
+                self.snapshot_synchronizer.lock().await,
+            )
+        };
+
+        // TODO: a temporary and stupid implementation...
```
chunshao90, line 228:
```suggestion
        // TODO: a temporary and rough implementation...
```

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+            log_encoding,
+            message_queue,
+            log_topic,
+        }
+    }
+
+    /// Write logs of table.
+    async fn write(
+        &self,
+        _ctx: &manager::WriteContext,
```
chunshao90, line 304:
```suggestion
        ctx: &manager::WriteContext,
```

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+        log_batch: &LogWriteBatch,
+    ) -> Result<SequenceNumber> {
+        // TODO: make use of ctx.
+        debug!(
+            "Begin to write to wal region, ctx:{:?}, region_id:{}, location:{:?}, log_entries_num:{}",
+            _ctx,
```
chunshao90, line 310:
```suggestion
            ctx,
```

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+
+        Ok(sequence_num - 1)
+    }
+
+    /// Read all logs from region, will return a iterator for high-level to
+    /// poll.
+    // TODO: take each read's timeout in consideration.
+    async fn read_all(
+        &self,
+        _ctx: &manager::ReadContext,
```
chunshao90, line 377:
```suggestion
        ctx: &manager::ReadContext,
```

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+            .map_err(|e| Box::new(e) as _)
+            .context(SyncSnapshot {
+                region_id: self.region_id,
+                topic: self.meta_topic.clone(),
+            })?;
+        assert_eq!(offsets.len(), 1);
+
+        // Delete old snapshots.
+        self.message_queue
+            .delete_up_to(&self.meta_topic, offsets[0])
```
ShiKaiWi, line 117: Is the offset passed to `delete_up_to`exclusive? It is weird that `up_to` seems inclusive if not. Is it wrong if it is.

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+            "Begin to check and clean logs, region id:{} snapshot:{:?}, topic:{}",
+            self.region_id, snapshot, self.log_topic
+        );
+
+        // Get prepare delete to offset.
+        let safe_delete_offset = Self::calc_safe_delete_offset(snapshot);
+        let prepare_delete_to_offset = match self.last_deleted_offset {
+            Some(last_deleted_offset) => {
+                assert!(safe_delete_offset >= last_deleted_offset,
+                    "impossible the new safe delete offset less than the last deleted offset, inner state inconsistent,
```
ShiKaiWi, line 168:
```suggestion
                    "the new safe delete offset should be larger than the last deleted offset, inner state inconsistent, 
```

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+        let safe_delete_offset = Self::calc_safe_delete_offset(snapshot);
+        let prepare_delete_to_offset = match self.last_deleted_offset {
+            Some(last_deleted_offset) => {
+                assert!(safe_delete_offset >= last_deleted_offset,
+                    "impossible the new safe delete offset less than the last deleted offset, inner state inconsistent, 
+                    safe delete offset:{}, last deleted offset:{}", safe_delete_offset, last_deleted_offset
+                );
+
+                if safe_delete_offset > last_deleted_offset {
+                    Some(last_deleted_offset)
```
ShiKaiWi, line 173: It is confusing to me that `last_deleted_offset` seems being always used afterwards if use the smaller one here.

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+                } else {
+                    None
```
ShiKaiWi, line 175: Why not just return here? Then there is no need to wrap the offset in an option.

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+        debug!(
+            "Produce to topic success, ctx:{:?}, region_id:{}, location:{:?}, topic:{}",
+            _ctx, self.region_id, log_batch.location, self.log_topic,
+        );
+
+        // Update after write.
+        self.region_meta
+            .update_after_table_write(
+                location.table_id,
+                OffsetRange::new(*offsets.first().unwrap(), *offsets.last().unwrap()),
```
ShiKaiWi, line 360: Shall we check the offsets to ensure it is not empty?

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+            previous_value: Vec::new(),
+        }
+    }
+}
+
+#[allow(unused)]
+impl<C: ConsumeIterator> MessageQueueLogIterator<C> {
+    pub async fn next_log_entry(&mut self) -> Result<Option<LogEntry<&'_ [u8]>>> {
+        if self.is_terminated && self.terminate_offset.is_some() {
+            info!(
```
ShiKaiWi, line 498: I guess this log should be set debug level.

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+            _ctx,
+            self.region_id,
+            log_batch.location,
+            log_batch.entries.len()
+        );
+        let location = &log_batch.location;
+        let log_write_entries = &log_batch.entries;
+
+        // Create messages and prepare for write.
+        let mut sequence_num = self
```
ShiKaiWi, line 319: Maybe we should name this `sequence_num` as `next_seq_num`, so in the end of this method it is natural to  return `next_seq_num - 1`.

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+            log_topic,
+        }
+    }
+
+    /// Write logs of table.
+    async fn write(
+        &self,
+        _ctx: &manager::WriteContext,
+        log_batch: &LogWriteBatch,
+    ) -> Result<SequenceNumber> {
```
ShiKaiWi, line 306: Maybe we should return error if the log batch is empty.

wal/src/message_queue_impl/region.rs
```diff
@@ -0,0 +1,761 @@
...
+                self.region_id, self.terminate_offset
+            );
+            return Ok(None);
+        }
+
+        let (message_and_offset, high_watermark) = self.iter.next_message().await.unwrap();
+
+        // Check has race happened.
+        if let Some(terminate_offset) = &self.terminate_offset {
+            ensure!(*terminate_offset == high_watermark, ReadAllLogsNoCause {
```
ShiKaiWi, line 509: I can't see where this condition is ensured to be true.

wal/src/message_queue_impl/helpers.rs
Rachelint: When current `safe_delete_offset` is not greater than `last_deleted_offset`, this represents no new progress, we should return None to tell this  and don't send the delete request.

wal/src/message_queue_impl/helpers.rs
Rachelint: Yes... it should be `safe_delete_offset`.

wal/src/message_queue_impl/helpers.rs
```diff
@@ -0,0 +1,221 @@
...
+///
+/// It will be locked before being called to keep the order of snapshots.
+pub struct SnapshotSynchronizer<Mq: MessageQueue> {
+    region_id: RegionId,
+    message_queue: Arc<Mq>,
+    meta_topic: String,
+    meta_encoding: MetaEncoding,
+}
+
+impl<Mq: MessageQueue> SnapshotSynchronizer<Mq> {
```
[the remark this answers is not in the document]
Rachelint: Done.

wal/src/message_queue_impl/helpers.rs
Rachelint: I modify `up_to` to `to` now.

wal/src/message_queue_impl/region.rs
Rachelint: It will just print once while finishing to poll, I think `info` may be ok?

wal/src/message_queue_impl/region.rs
Rachelint: It is actually unused now, I will remove the `_` after making use of it.

wal/src/message_queue_impl/region.rs
Rachelint: Yes, it should not be the logic here, because it is a basic component used for `scanning region during recovering`, `scanning the range of a table`, `fetching for slave node`. 

The only restriction for the `terminate_offset ` should be less than or equal to high watermark.

I plan to define some high-level iterator wrapping it as the inner later, for processing the high-level logic. But the pr is a bit larger, so let us keep it now

wal/src/message_queue_impl/snapshot_synchronizer.rs
```diff
@@ -0,0 +1,125 @@
...
+        let offsets = self
+            .message_queue
+            .produce(&self.meta_topic, messages)
+            .await
+            .map_err(|e| Box::new(e) as _)
+            .context(SyncSnapshot {
+                region_id: self.region_id,
+                topic: self.meta_topic.clone(),
+            })?;
+        assert_eq!(offsets.len(), 1);
```
ShiKaiWi, line 106: Maybe we should throw error instead of panic if a wrong result comes from an outside service.

wal/src/message_queue_impl/helpers.rs
ShiKaiWi: It seems no such modification now.

Rachelint: I propose to make modification about message queue component together in a new pr, just keep it now?

wal/src/message_queue_impl/helpers.rs
Rachelint: Done.

ShiKaiWi (approved):

Issue #363, opened 2022-11-03
Tracking issue for query performance


jiacai2050:
### Describe This Problem

For now, the main bottleneck within a query is sst scan, there are several improvement can be made to accelerate scan process:
- [x] Cache, as described in #312, the row-group level cache is broken, the async reader fix this but has performance regression, we need to investigate this
- [x] min/max filter is broken in main, #360 fix this
- [x] min/max filter may not enough for columns with high cardinality, in this case bloom filter may be a better choice

### Proposal

As above.

### Additional Context

https://github.com/apache/parquet-format/blob/master/BloomFilter.md

PR #368, opened 2022-11-03
feat: replace ParquetExec with ParquetRecordBatchStream

jiacai2050:

Part of #363  

# Rationale for this change
 
As describe in #363, there is perf regression in ParquetExec, and its interface is not flexible enough for us to control row group filter.

`ParquetRecordBatchStream` is the basic struct in parquet to read parquet, we can more control how to parse parquet by using it.

1. Remove ParquetExec 
2. Move row group filter logic into a new file `prune.rs` in `parquet_ext` module, so this logic can be used in both sync and async parquet reader.
3. Refactor some tests to make it more readable.

No

# How does this change test

UT


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -234,93 +178,36 @@ impl<'a> Reader<'a> {
...
     ) -> Self {
         Self {
             storage,
+            path,
             data_cache,
-            meta,
-            metrics,
-            metadata_size_hint,
+            file_size,
+            bytes_scanned: 0,
```
ShiKaiWi, line 210: How about moving the three statistics variable into a struct to make the struct clean?

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -447,6 +350,7 @@ impl Stream for RecordBatchProjector {
                             .map_err(|e| Box::new(e) as _)
                             .context(DecodeRecordBatch)?;
 
+                        info!("batch is {:?}", record_batch);
```
ShiKaiWi, line 353: Remove this line?

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -333,28 +230,25 @@ impl CachableParquetFileReader {
 
 impl Drop for CachableParquetFileReader {
     fn drop(&mut self) {
-        info!(
-            "CachableParquetFileReader meta_size_hint:{:?}, cache_hit:{}, cache_miss:{}, bytes_scanned:{}",
-            self.metadata_size_hint, self.cache_hit, self.cache_miss, self.metrics.bytes_scanned.value()
-        );
+        info!("CachableParquetFileReader metrics:{:?}", self.metrics,);
```
ShiKaiWi, line 233:
```suggestion
        info!("CachableParquetFileReader metrics:{:?}", self.metrics);
```

ShiKaiWi (approved):
LGTM

PR #370, opened 2022-11-04
feat: add bloom filter when write sst

jiacai2050:

Part of #363 

# Rationale for this change
 
Described in #363, bloom filter is beneficial for columns with high cardinality, so this PR append bloom filter to sst's meta when write SST files.

Update sst write process, loop record batch twice:
- first to build bloom filter
- second to write to underlying storage.

> Note: for now, all columns is appended with bloom filter, some columns may not suitable for it, we can remove them in future to reduce meta size.

No

# How does this change test

Add UT: test_partition_record_batch

After we support prune by bloom filter, we can add more test cases.

### DIFF
--- a/analytic_engine/src/sst/file.rs
+++ b/analytic_engine/src/sst/file.rs
@@ -58,14 +58,14 @@ pub enum Error {
     StorageFormatOptionsNotFound { backtrace: Backtrace },
 
     #[snafu(display("Bloom filter options are not found.\nBacktrace\n:{}", backtrace))]
-    BloomFitlerNotFound { backtrace: Backtrace },
+    BloomFilterNotFound { backtrace: Backtrace },
 
     #[snafu(display(
         "Bloom filter should be 256 byte, current:{}.\nBacktrace\n:{}",
         size,
         backtrace
     ))]
-    InvalidBloomFilter { size: usize, backtrace: Backtrace },
+    InvalidBloomFilterSize { size: usize, backtrace: Backtrace },
 
     #[snafu(display("Failed to convert time range, err:{}", source))]
     ConvertTimeRange { source: common_types::time::Error },
@@ -448,21 +448,23 @@ impl BloomFilter {
     pub fn new(filters: Vec<Vec<Bloom>>) -> Self {
         Self { filters }
     }
+}
 
-    pub fn to_pb(&self) -> sst_pb::SstBloomFilter {
-        let filters = self
+impl From<BloomFilter> for sst_pb::SstBloomFilter {
+    fn from(bloom_filter: BloomFilter) -> Self {
+        let row_group_filters = bloom_filter
             .filters
             .iter()
             .map(|row_group_filter| {
-                let bloom_filter = row_group_filter
+                let column_filters = row_group_filter
                     .iter()
                     .map(|column_filter| column_filter.data().to_vec())
                     .collect::<Vec<_>>();
-                sst_pb::sst_bloom_filter::RowGroupFilter { bloom_filter }
+                sst_pb::sst_bloom_filter::RowGroupFilter { column_filters }
             })
             .collect::<Vec<_>>();
 
-        sst_pb::SstBloomFilter { filters }
+        sst_pb::SstBloomFilter { row_group_filters }
     }
 }
 
@@ -471,18 +473,18 @@ impl TryFrom<sst_pb::SstBloomFilter> for BloomFilter {
 
     fn try_from(src: sst_pb::SstBloomFilter) -> Result<Self> {
         let filters = src
-            .filters
+            .row_group_filters
             .into_iter()
             .map(|row_group_filter| {
                 row_group_filter
-                    .bloom_filter
+                    .column_filters
                     .into_iter()
                     .map(|encoded_bytes| {
                         let size = encoded_bytes.len();
                         let bs: [u8; 256] = encoded_bytes
                             .try_into()
                             .ok()
-                            .with_context(|| InvalidBloomFilter { size })?;
+                            .context(InvalidBloomFilterSize { size })?;
 
                         Ok(Bloom::from(bs))
                     })
@@ -529,7 +531,7 @@ impl From<SstMetaData> for sst_pb::SstMetaData {
             size: src.size,
             row_num: src.row_num,
             storage_format_opts: Some(src.storage_format_opts.into()),
-            bloom_filter: Some(src.bloom_filter.to_pb()),
+            bloom_filter: Some(src.bloom_filter.into()),
         }
     }
 }
@@ -551,7 +553,7 @@ impl TryFrom<sst_pb::SstMetaData> for SstMetaData {
                 .context(StorageFormatOptionsNotFound)?,
         );
         let bloom_filter = {
-            let pb_filter = src.bloom_filter.context(BloomFitlerNotFound)?;
+            let pb_filter = src.bloom_filter.context(BloomFilterNotFound)?;
             BloomFilter::try_from(pb_filter)?
         };
 
--- a/analytic_engine/src/sst/file.rs
+++ b/analytic_engine/src/sst/file.rs
@@ -27,6 +27,7 @@ use common_util::{
     metric::Meter,
     runtime::{JoinHandle, Runtime},
 };
+use ethbloom::Bloom;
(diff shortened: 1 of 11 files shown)


tustvold:
FWIW parquet has a form of bloom filter support, and I, and likely others, would be very interested in collaborating to get support added to arrow-rs/DataFusion if this was something you were amenable to. Just thought I'd mention it, as it has come up in a few IOx discussions lately :smile:

alamb:
This is very cool work @jiacai2050  👍 

Along with @tustvold I would love to help collaborate on getting Bloom filter support into parquet (and then also datafusion) -- my first contribution to the process is writing up a ticket https://github.com/apache/arrow-rs/issues/3023

jiacai2050:
@tustvold @alamb Thanks for your tips. Great write-up.

The bloom filter support in CeresDB is in early stage, we have just drafted an initial design how to add bloom filter, which is different from what [parquet-format](https://github.com/apache/parquet-format/blob/master/BloomFilter.md) said.

Simply put, bloom filter is row-group based, and we intend to encode them in `key_value_metadata` instead of in each row group's meta, so we can read them out in one IO.

We will verify this design in our case ASAP, if it's works as expected, we would be happy to contribute it to parquet crates.

alamb:
> Simply put, bloom filter is row-group based, and we intend to encode them in key_value_metadata instead of in each row group's meta, so we can read them out in one IO.

Sounds like a good plan -- thank you @jiacai2050 ! I think using a custom format in the metadata sounds like a good idea for CeresDB initially (it will be much faster than working it through the rest of the ecosystem). We would love to hear about your experience in implementing it

analytic_engine/src/sst/parquet/builder.rs
```diff
@@ -80,28 +74,78 @@ impl RecordBytesReader {
...
+        .context(EncodeRecordBatch)?;
+
+        // process record batch stream
+        let mut arrow_record_batch_vec = Vec::new();
+        for record_batches in self.fetched_record_batch {
+            for batch in record_batches {
+                arrow_record_batch_vec.push(batch.into_record_batch().into_arrow_record_batch());
+            }
+
+            let buf_len = arrow_record_batch_vec.len();
```
ShiKaiWi, line 142: One thing I'm worried about is that the real row group distribution may not be the exactly same as the `self.fetched_record_batch`. The bloom filters for row groups will be wrong if it is indeed not what we expect.

analytic_engine/src/sst/file.rs
```diff
@@ -425,6 +436,64 @@ impl FileMeta {
...
+            .into_iter()
+            .map(|row_group_filter| {
+                row_group_filter
+                    .bloom_filter
+                    .into_iter()
+                    .map(|encoded_bytes| {
+                        let size = encoded_bytes.len();
+                        let bs: [u8; 256] = encoded_bytes
+                            .try_into()
+                            .ok()
```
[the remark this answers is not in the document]
jiacai2050: try_into return `Result`

analytic_engine/src/sst/parquet/builder.rs
jiacai2050: In current implementation, rows inside a row group(except the last one) will be larger than given `num_rows_per_row_group`.

I don't see where might be wrong this this case.

analytic_engine/src/sst/parquet/builder.rs
```diff
@@ -54,23 +55,16 @@ struct RecordBytesReader {
     compression: Compression,
     meta_data: SstMetaData,
     total_row_num: Arc<AtomicUsize>,
-
-    fetched_row_num: usize,
+    fetched_record_batch: Vec<Vec<RecordBatchWithKey>>,
```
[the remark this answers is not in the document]
jiacai2050: Change to `partitioned_record_batch` add some comments

analytic_engine/src/sst/file.rs
```diff
@@ -425,6 +436,64 @@ impl FileMeta {
...
+    // 2. column
+    filters: Vec<Vec<Bloom>>,
+}
+
+impl BloomFilter {
+    pub fn new(filters: Vec<Vec<Bloom>>) -> Self {
+        Self { filters }
+    }
+
+    pub fn to_pb(&self) -> sst_pb::SstBloomFilter {
```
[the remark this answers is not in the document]
jiacai2050: fixed

analytic_engine/src/sst/file.rs
```diff
@@ -56,6 +57,16 @@ pub enum Error {
...
 
+    #[snafu(display("Bloom filter options are not found.\nBacktrace\n:{}", backtrace))]
+    BloomFitlerNotFound { backtrace: Backtrace },
+
+    #[snafu(display(
+        "Bloom filter should be 256 byte, current:{}.\nBacktrace\n:{}",
+        size,
+        backtrace
+    ))]
+    InvalidBloomFilter { size: usize, backtrace: Backtrace },
```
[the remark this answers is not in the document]
jiacai2050: fixed

common_types/src/datum.rs
```diff
@@ -484,6 +484,32 @@ impl Datum {
         }
     }
 
+    pub fn as_bytes(&self) -> Vec<u8> {
```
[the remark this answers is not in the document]
jiacai2050: fixed

analytic_engine/src/sst/file.rs
```diff
@@ -425,6 +436,64 @@ impl FileMeta {
...
+            .map(|row_group_filter| {
+                row_group_filter
+                    .bloom_filter
+                    .into_iter()
+                    .map(|encoded_bytes| {
+                        let size = encoded_bytes.len();
+                        let bs: [u8; 256] = encoded_bytes
+                            .try_into()
+                            .ok()
+                            .with_context(|| InvalidBloomFilter { size })?;
```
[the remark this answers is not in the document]
jiacai2050: fixed

analytic_engine/src/sst/parquet/builder.rs
ShiKaiWi: What I mean is that one `arrow_record_batch_vec` from `partitioned_record_batch` here maybe not the real row group in the parquet, which leads to wrong usage of the bloom filters because it is built from the `partitioned_record_batch`.

jiacai2050: Got your point.

https://github.com/jiacai2050/arrow-rs/blob/a881083c425dfbc1a87fb510bc32644bca16fbf3/parquet/src/arrow/arrow_writer/mod.rs#L152

In `flush_completed` of parquet,  it will write one row group once `max_row_group_size` is reached, it's the same with what we expect, at least for now.

ShiKaiWi: According to https://github.com/jiacai2050/arrow-rs/blob/a881083c425dfbc1a87fb510bc32644bca16fbf3/parquet/src/arrow/arrow_writer/mod.rs#L193-L198, one row group contains at most `max_row_group_size` which is different from the distribution of our partitioned_record_batch.

jiacai2050: fixed

analytic_engine/src/sst/parquet/builder.rs
```diff
@@ -331,4 +441,75 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_partition_record_batch() {
+        // rows per group: 10
+        let testcases = vec![
+            // input, expected
+            (vec![10, 10], vec![10, 10]),
+            (vec![10, 10, 1], vec![10, 10, 1]),
+            (vec![10, 10, 21], vec![10, 10, 10, 10, 1]),
+            (vec![5, 6, 10], vec![10, 10, 1]),
+            (vec![5, 4, 4, 30], vec![10, 10, 10, 10, 3]),
```
ShiKaiWi, line 454:
```suggestion
            (vec![5, 4, 4, 30], vec![10, 10, 10, 10, 3]),
            (vec![], vec![]),
            (vec![20,7,23,20], vec![10, 10, 10, 10, 10, 10, 10]),
            (vec![21], vec![10, 10, 1]),
```

ShiKaiWi (approved):
LGTM

alamb:
🎉

PR #372, opened 2022-11-07
feat: support equal pruner

ShiKaiWi:

Part of #363

# Rationale for this change
 Prepare for using bloom filter (and maybe dictionary) to do row group filtering.

- Add a new pruner called `EqPrunner`.

None.

# How does this change test
New ut.


components/parquet_ext/src/prune/equal.rs
```diff
@@ -0,0 +1,343 @@
...
+            }
+            Operator::Eq => normalize_equal_expr(left, right, true),
+            Operator::NotEq => normalize_equal_expr(left, right, false),
+            _ => unhandled,
+        },
+        Expr::InList {
+            expr,
+            list,
+            negated,
+        } if !list.is_empty() && list.len() < MAX_ELEMS_IN_LIST_FOR_FILTER => {
```
jiacai2050, line 109: If list is empty, we can add a `NormalizedExpr::False` to handle this.

jiacai2050 (approved):
lgtm

PR #375, opened 2022-11-07
feat: enhancement to mq's kafka implementation

### LINKED ISSUE #366: Tracking issue about todo in distributed wal
**Description**
There are some enhancements in wal needing to do in the futures, just mark them now. 

**Proposal**
### On message
- [x] Now manifest based on wal need to scan twice from storage while doing snapshot. Add a cache for this may be better.
- [ ] In wal's message queue implementation, we now sync the snapshot to remote in every flushing. Indeed, we should just need to do the sync work while flushing after dropping table.
- [x] Now, the desigin of wal on message queue is based on the thread model of `one writer for one table`, maybe we should remove this assumption.
- [x] Define some iterators with more explicit semantics.
- [ ] Now we init region(heavy work with network IO) even if it is just created, we should change  `MessageQueue` to know whether it is a new region.
- [x] We should disable the ttl of topic in kafka, but it is not supported in `rskafka` now. It is important so I plan to use another crate to finish it temporarily.
### On OBKV
- [ ] Rename the table name prefix.
- [x] Change the meta table's sharding key to `table id`.
### Manifest(on wal)
- [x] Use the table id as region id.

**Additional context**

Rachelint:

Close #366 

# Rationale for this change
 Enhancement to mq's kafka implementation.

+ add some useful traits.
+ add a test about multiple consumers on a same topic.
+ rename `delete_up_to` to `delete_to`.

None.

# How does this change test
Test by ut.

### DIFF
--- a/components/message_queue/src/lib.rs
+++ b/components/message_queue/src/lib.rs
@@ -43,7 +43,7 @@ pub trait MessageQueue: Debug + Send + Sync + 'static {
         start_offset: StartOffset,
     ) -> Result<Self::ConsumeIterator, Self::Error>;
 
-    async fn delete_up_to(&self, topic_name: &str, offset: Offset) -> Result<(), Self::Error>;
+    async fn delete_to(&self, topic_name: &str, offset: Offset) -> Result<(), Self::Error>;
     // TODO: should design a stream consume method for slave node to fetch wals.
 }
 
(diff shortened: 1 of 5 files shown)


components/message_queue/src/lib.rs
```diff
@@ -15,7 +19,7 @@ pub type Offset = i64;
 
 /// Message queue interface supporting the methods needed in wal module.
 #[async_trait]
-pub trait MessageQueue: Send + Sync + 'static {
+pub trait MessageQueue: Send + Sync + 'static + Debug {
```
ShiKaiWi, line 22: Let's put the `'static` at the end.

ShiKaiWi (approved):
LGTM

PR #377, opened 2022-11-07
fix: add ignore to  test on message queue's kafka implementation.

Rachelint:

Closes #

# Rationale for this change
Tests on message queue's kafka implementation should not be run automatically.

Add `ignore` to message queue's tests.

None.

# How does this change test
None.

### DIFF
--- a/components/message_queue/src/tests/cases.rs
+++ b/components/message_queue/src/tests/cases.rs
@@ -19,6 +19,7 @@ use crate::{
 };
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[ignore]
 async fn test_kafka() {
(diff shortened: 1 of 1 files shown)


components/message_queue/src/tests/cases.rs
```diff
@@ -19,6 +19,7 @@ use crate::{
 };
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[ignore]
```
ShiKaiWi, line 22: It will be apprecaited if you can add reason for the ignore:
```suggestion
#[ignore = "xxxx"]
```

ShiKaiWi (approved):

PR #381, opened 2022-11-08
refactor: build row groups from record batch stream

ShiKaiWi:

Refactor for #370

# Rationale for this change
 Previous implementation for building row groups from record batch is a little bit complex, and this pr will do some refactoring work for that.

- Refactor the `partition_record_batch`;

None.

# How does this change test
Existing ut.

### DIFF
--- a/analytic_engine/src/sst/parquet/builder.rs
+++ b/analytic_engine/src/sst/parquet/builder.rs
@@ -121,8 +121,8 @@ impl RecordBytesReader {
                         self.request_id
                     );
 
-                    // Updated the exhausted `prev_record_batch`, and let next loop to continue fill
-                    // `curr_row_group`.
+                    // Updated the exhausted `prev_record_batch`, and let next loop to continue to
+                    // fill `curr_row_group`.
                     prev_record_batch.replace(v);
                 }
                 None => break,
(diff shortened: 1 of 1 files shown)


analytic_engine/src/sst/parquet/builder.rs
```diff
@@ -67,80 +64,72 @@ impl RecordBytesReader {
...
-        let mut remaining = Vec::with_capacity(pending_record_batch.len());
-        while let Some(batch) = pending_record_batch.pop_front() {
-            remaining.push(batch);
-        }
-        if !remaining.is_empty() {
-            self.partitioned_record_batch.push(remaining);
+        Ok(())
+    }
+
+    /// Fetch an integrate row group from the `self.record_stream`.
```
jiacai2050, line 80: What does `integrate row group` means?

analytic_engine/src/sst/parquet/builder.rs
```diff
@@ -67,80 +64,72 @@ impl RecordBytesReader {
...
+        let mut curr_row_group = vec![];
+        // Used to record the number of remaining rows to fill `curr_row_group`.
+        let mut remaining = self.num_rows_per_row_group;
+
+        // Keep filling `curr_row_group` until `remaining` is zero.
+        while remaining > 0 {
+            // Use the `prev_record_batch` to fill `curr_row_group` if possible.
+            if let Some(v) = prev_record_batch {
+                let total_rows = v.num_rows();
+                if total_rows <= remaining {
```
jiacai2050, line 98: Should be `total_rows < remaining`?

analytic_engine/src/sst/parquet/builder.rs
```diff
@@ -67,80 +64,72 @@ impl RecordBytesReader {
...
+                    // The whole record batch is part of the `curr_row_group`, and let's feed it
+                    // into `curr_row_group`.
+                    curr_row_group.push(prev_record_batch.take().unwrap());
+                    remaining -= total_rows;
+                } else {
+                    // Only part of the record batch fills `curr_row_group`, and let's fill the
```
jiacai2050, line 104: How about
> Only first `remaining` record batch belongs to `curr_row_group`, the rest should be put to `prev_record_batch` for next row_group

jiacai2050:
The general implementation is more clearer than before, good job!

analytic_engine/src/sst/parquet/builder.rs
ShiKaiWi: The word `integrate` may be a wrong usage here. Maybe `integral` or `complete` is better.

Actually, it means the number of the row group is `num_rows_per_group`.

analytic_engine/src/sst/parquet/builder.rs
ShiKaiWi: I guess it should be `<=` here.

analytic_engine/src/sst/parquet/builder.rs
ShiKaiWi: Ok. It's nicer, I will put it like:
> Only first `remaining` rows of the record batch belongs to `curr_row_group`, the rest should be put to `prev_record_batch` for next row group.

jiacai2050 (approved):
LGTM

PR #383, opened 2022-11-09
feat: filter row groups by bloom filter

ShiKaiWi:

Part of #363 

# Rationale for this change
 Utilize the bloom filter stored in the sst meta data to do filtering on the row groups according to the provided predicate.

- Support filter row groups in when reading file.
- Update the `EqPruner`'s interface to make it more reasonable.

None.

# How does this change test
New unit test.

### DIFF
--- a/analytic_engine/src/sst/parquet/row_group_filter.rs
+++ b/analytic_engine/src/sst/parquet/row_group_filter.rs
@@ -17,6 +17,10 @@ use snafu::ensure;
 
 use crate::sst::reader::error::{OtherNoCause, Result};
 
+/// A filter to prune row groups according to the provided predicates.
+///
+/// Currently, two kinds of filters will be applied to such filtering:
+/// min max & bloom filter.
 pub struct RowGroupFilter<'a> {
     schema: &'a SchemaRef,
     row_groups: &'a [RowGroupMetaData],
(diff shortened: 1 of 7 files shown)


analytic_engine/src/sst/parquet/row_group_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+            schema,
+            row_groups,
+            blooms,
+            predicates,
+        })
+    }
+
+    pub fn filter(&self) -> Vec<usize> {
+        let filtered0 = self.filter_by_min_max();
+        let filtered1 = self.filter_by_bloom();
```
jiacai2050, line 52: Why not pass filtered0 to `filter_by_bloom`, it will reduce total row groups to filter

components/parquet_ext/src/prune/equal.rs
```diff
@@ -436,7 +440,7 @@ mod tests {
         // | 1  | 2  | 3  |
         // | 2  | 3  | 4  |
         let row_groups = vec![vec![0, 1, 2], vec![1, 2, 3], vec![2, 3, 4]];
-        let is_equal = |pos: ColumnPosition, val: &ScalarValue, negated: bool| -> bool {
+        let is_equal = |pos: ColumnPosition, val: &ScalarValue, negated: bool| -> Option<bool> {
```
jiacai2050, line 443: Why change to Option here? when will we return None?

ShiKaiWi: When the compare result is unknown.

analytic_engine/src/sst/parquet/row_group_filter.rs
ShiKaiWi: It's a good idea, but current the interface of `EqPruner` seems not flexible to allow us to such improvement. I will leave a TODO here.

jiacai2050 (approved):
LGTM

PR #385, opened 2022-11-09
feat: implement region recovery

### LINKED ISSUE #384: Implement region's recovery
**Description**
After implementing region's `read`, `write`, `mark  delete` in  #367, we need to implement region's recovery now.

**Proposal**
+ implement region meta's recovery.
+ implement region's recovery based on above step.

**Additional context**
Part of #284

Rachelint:

Closes #384

# Rationale for this change
+ see #384.

Implement region's recovery.

None.

# How does this change test
Test by ut.

### DIFF
--- a/wal/src/message_queue_impl/region.rs
+++ b/wal/src/message_queue_impl/region.rs
@@ -182,7 +182,7 @@ impl<M: MessageQueue> Region<M> {
     /// Init the region.
     pub async fn open(namespace: &str, region_id: RegionId, message_queue: Arc<M>) -> Result<Self> {
         info!(
-            "Open region in namespace, namespace:{}, region id:{}",
+            "Begin to open region in namespace, namespace:{}, region id:{}",
             namespace, region_id
         );
 
@@ -222,6 +222,7 @@ impl<M: MessageQueue> Region<M> {
             &meta_encoding,
         )
         .await?;
+
         Self::recover_region_meta_from_log(
             namespace,
             region_id,
@@ -251,6 +252,11 @@ impl<M: MessageQueue> Region<M> {
         ));
         let log_cleaner = LogCleaner::new(region_id, message_queue.clone(), log_topic);
 
+        info!(
+            "Finish to open region in namespace, namespace:{}, region id:{}",
+            namespace, region_id
+        );
+
         Ok(Region {
             inner,
             snapshot_synchronizer,
@@ -266,7 +272,12 @@ impl<M: MessageQueue> Region<M> {
         meta_topic: &str,
         meta_encoding: &MetaEncoding,
     ) -> Result<Offset> {
-        // Fetch earliest offset and high watermark, then check.
+        info!(
+            "Recover region meta from meta, namespace:{}, region id:{}",
+            namespace, region_id
+        );
+
+        // Fetch high watermark and check.
         let high_watermark = message_queue
             .fetch_offset(meta_topic, OffsetType::HighWaterMark)
             .await
@@ -274,10 +285,12 @@ impl<M: MessageQueue> Region<M> {
             .context(OpenWithCause {
                 namespace,
                 region_id,
-                msg: "failed while region_meta_from_meta",
+                msg: "failed while recover from meta",
             })?;
 
+        // Meta topic is empty, it just need to recover from log topic.
         if high_watermark == 0 {
+            debug!("Meta topic is empty, it just need to recover from log topic, namespace:{}, region id:{}", namespace, region_id);
             return Ok(0);
         }
 
@@ -289,7 +302,7 @@ impl<M: MessageQueue> Region<M> {
             .context(OpenWithCause {
                 namespace,
                 region_id,
-                msg: "failed while region_meta_from_meta",
+                msg: "failed while recover from meta",
             })?;
 
         let (latest_message_and_offset, returned_high_watermark) = iter
@@ -299,12 +312,12 @@ impl<M: MessageQueue> Region<M> {
             .context(OpenWithCause {
                 namespace,
                 region_id,
-                msg: "failed while region_meta_from_meta",
+                msg: "failed while recover from meta",
             })?;
 
         // TODO: maybe should assert it?
         ensure!(returned_high_watermark == high_watermark, OpenNoCause { namespace , region_id, msg: format!(
-            "failed while region_meta_from_meta, high watermark shouldn't changed while opening region, 
+            "failed while recover from meta, high watermark shouldn't changed while opening region, 
             origin high watermark:{}, returned high watermark:{}", high_watermark, returned_high_watermark)
         });
 
@@ -315,7 +328,7 @@ impl<M: MessageQueue> Region<M> {
             .with_context(|| OpenNoCause {
                 namespace,
                 region_id,
-                msg: "failed while region_meta_from_meta, key in message shouldn't be None",
+                msg: "failed while recover from meta, key in message shouldn't be None",
             })?;
 
         let raw_value = latest_message_and_offset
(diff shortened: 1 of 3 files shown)


wal/src/message_queue_impl/region.rs
```diff
@@ -200,13 +252,260 @@ impl<M: MessageQueue> Region<M> {
...
+            .map_err(|e| Box::new(e) as _)
+            .context(OpenWithCause {
+                namespace,
+                region_id,
+                msg: "failed while recover from meta",
+            })?;
+
+        // Meta topic is empty, it just need to recover from log topic.
+        if high_watermark == 0 {
+            debug!("Meta topic is empty, it just need to recover from log topic, namespace:{}, region id:{}", namespace, region_id);
```
ShiKaiWi, line 293:
```suggestion
            debug!("Meta topic is empty, it just needs to recover from log topic, namespace:{}, region id:{}", namespace, region_id);
```

wal/src/message_queue_impl/region.rs
```diff
@@ -200,13 +252,260 @@ impl<M: MessageQueue> Region<M> {
...
+            .fetch_offset(meta_topic, OffsetType::HighWaterMark)
+            .await
+            .map_err(|e| Box::new(e) as _)
+            .context(OpenWithCause {
+                namespace,
+                region_id,
+                msg: "failed while recover from meta",
+            })?;
+
+        // Meta topic is empty, it just need to recover from log topic.
```
ShiKaiWi, line 291:
```suggestion
        // Meta topic is empty, it just needs to recover from log topic.
```

wal/src/message_queue_impl/region.rs
```diff
@@ -200,13 +252,260 @@ impl<M: MessageQueue> Region<M> {
...
+            .next_message()
+            .await
+            .map_err(|e| Box::new(e) as _)
+            .context(OpenWithCause {
+                namespace,
+                region_id,
+                msg: "failed while recover from meta",
+            })?;
+
+        // TODO: maybe should assert it?
```
ShiKaiWi, line 318: I guess ensure is enough, because it is better not to assume the remote service always right.

wal/src/message_queue_impl/region.rs
```diff
@@ -163,29 +181,63 @@ pub struct Region<M: MessageQueue> {
...
+            &log_topic,
+            &log_encoding,
+        )
+        .await?;
 
         // Init region inner.
         let inner = RwLock::new(RegionInner::new(
             region_id,
+            region_meta_builder.build(),
             CommonLogEncoding::newest(),
```
ShiKaiWi, line 241: Use `log_encoding` instead?

wal/src/message_queue_impl/region.rs
```diff
@@ -163,29 +181,63 @@ pub struct Region<M: MessageQueue> {
...
+
+        // Build region meta.
+        let mut region_meta_builder = RegionMetaBuilder::default();
+        let high_watermark_in_snapshot = Self::recover_region_meta_from_meta(
+            namespace,
+            region_id,
+            &mut region_meta_builder,
+            message_queue.as_ref(),
+            &meta_topic,
+            &meta_encoding,
```
ShiKaiWi, line 222: Seems no need to pass the `meta_encoding` to the method, and why not just declare it in the method?

wal/src/message_queue_impl/region.rs
```diff
@@ -200,13 +252,260 @@ impl<M: MessageQueue> Region<M> {
...
             inner,
             snapshot_synchronizer,
             log_cleaner,
         })
     }
 
+    async fn recover_region_meta_from_meta(
+        namespace: &str,
+        region_id: RegionId,
+        builder: &mut RegionMetaBuilder,
```
ShiKaiWi, line 270: Put the returned parameter at the last.

wal/src/message_queue_impl/region.rs
```diff
@@ -200,13 +252,260 @@ impl<M: MessageQueue> Region<M> {
...
+                    namespace,
+                    region_id,
+                    msg: format!(
+                        "failed while recover from log, region meta delta:{:?}",
+                        region_meta_delta
+                    ),
+                });
+
+            // Has polled last log in topic, break.
+            if latest_message_and_offset.offset + 1 == high_watermark {
```
ShiKaiWi, line 500: Any possibilities that the `latest_message_and_offset.offset + 1` always larger than `high_watermark`? Shall we do some kind of check for this case?

wal/src/message_queue_impl/region.rs
```diff
@@ -697,12 +1007,10 @@ impl<C: ConsumeIterator> MessageQueueLogIterator<C> {
                 msg: "failed while polling log",
             })?;
 
-        self.previous_value = payload.to_owned();
-
         Ok(Some(LogEntry {
             table_id: log_key.table_id,
             sequence: log_key.sequence_num,
-            payload: self.previous_value.as_slice(),
+            payload: payload.to_owned(),
```
ShiKaiWi, line 1013: Actually, this will lead to another copy. Is it necessary to eliminate the lifetime from the returned value?

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -339,6 +345,118 @@ impl OffsetRange {
...
+#[allow(unused)]
+impl RegionMetaBuilder {
+    pub fn apply_region_meta_snapshot(&mut self, snapshot: RegionMetaSnapshot) -> Result<()> {
+        debug!("Apply region meta snapshot, snapshot:{:?}", snapshot);
+
+        for entry in snapshot.entries.into_iter() {
+            ensure!(self.table_metas.insert(entry.table_id, entry.clone().into()).is_none(),
+                Build { msg: format!("apply snapshot failed, shouldn't exist duplicated entry in snapshot, duplicated entry:{:?}", entry) }
+            );
+        }
```
ShiKaiWi, line 364:
```suggestion
        for entry in snapshot.entries {
            let old_meta = self.table_metas.insert(entry.table_id, entry.clone().into());
            ensure!(old_meta.is_none(),
                Build { msg: format!("apply snapshot failed, shouldn't exist duplicated entry in snapshot, duplicated entry:{:?}", entry) }
            );
        }
```

wal/src/message_queue_impl/region_meta.rs
```diff
@@ -339,6 +345,118 @@ impl OffsetRange {
...
+
+        Ok(())
+    }
+
+    pub fn apply_region_meta_delta(&mut self, delta: RegionMetaDelta) -> Result<()> {
+        debug!("Apply region meta delta, delta:{:?}", delta);
+
+        let mut table_meta_inner = self
```
ShiKaiWi, line 372: Shall we just call this variable as `table_meta`? It is weird to suffix it with `inner`.

wal/src/message_queue_impl/region.rs
Rachelint: It is not necessary, just for adapting to the trait defined in `wal/src/manager.rs`.
It can actually avoid cloning in some implementation such as rocksDB, so I think keeping it is no problem.

wal/src/message_queue_impl/region.rs
Rachelint: There has been check above to ensure `high watermark` equal to `returned high watermark`, and `offset + 1` will always be less than or equal to `returned high watermark`.  I think checking again may be unnecessary?

wal/src/message_queue_impl/region.rs
Rachelint: Because I need to wrap this iterator as a inner in later development, and if return slice here, the borrow checker may complain... The refactor here just for making it happy.

wal/src/message_queue_impl/region.rs
Rachelint: `SnapshotSynchronizer`'s initialization need `MetaEncoding`, too.
So I think define here and pass is better.

wal/src/message_queue_impl/region.rs
Rachelint: Has reverted this refactor.

ShiKaiWi (approved):
LGTM

PR #387, opened 2022-11-10
refactor: wal unit tests.

### LINKED ISSUE #386: Unit test in wal module can be more extensible
### Describe This Problem

Now, the tests code in wal module is so hard code ... It doesn't take full advantage of the `WalBuilder` trait.
So it can be more extensible after refactoring. 

### Proposal

Pass the builder to build tests for different implementation of wal.

### Additional Context

_No response_

Rachelint:

Closes #386 

# Rationale for this change
 
See #386 

Refactor tests in wal.

None.

# How does this change test
They are tests themselves.

### DIFF
--- a/wal/src/tests/read_write.rs
+++ b/wal/src/tests/read_write.rs
@@ -3,7 +3,7 @@
 use std::{ops::Deref, sync::Arc};
 
 use common_types::{
-    table::{Location, DEFAULT_SHARD_ID},
+    table::{Location, TableId, DEFAULT_SHARD_ID},
     SequenceNumber,
 };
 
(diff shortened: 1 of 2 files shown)


wal/src/tests/read_write.rs
```diff
@@ -15,6 +15,130 @@ use crate::{
...
+}
+
+#[test]
+fn test_memory_table_wal_with_ttl() {
+    let builder = MemoryTableWalBuilder::with_ttl("1d");
+
+    test_all(builder);
+}
+
+fn test_all<B: WalBuilder>(builder: B) {
```
ShiKaiWi, line 39: I guess `test_all` is not a best practice for unit test because it is hard to find the problems in such a long procedure when some failure happens.

ShiKaiWi (approved):
LGTM

Issue #391, opened 2022-11-14
Enhance cache for OSS storage


ShiKaiWi:
### Describe This Problem

Currently, we have `DataCache` & `MetaCache` for underlying storage to speed up the query process, and specifically, for OSS storage, a disk cache with size limit is designed for a larger capacity. However, there are still some issues concerning the current cache design, leading to poor performance of query:
- It costs too much CPU and takes long time to decode the custom metadata (including bloom filters, time range of sst) encoded as extended key value pairs in the parquet footer.
- The memory cache has only a limit on the entry number, and no limit on the size, that is to say, such a cache may need much tuning to make it perform best for the user's data pattern and the memory capacity of the machine on which ceresdb is deployed.

### Proposal

To solve the two issues about the current cache design, I propose:
- Add cache for both custom metadata and parquet metadata, to avoid extra decoding procedure for our custom metadata. #392
- Refactor memory data cache to make it featured with size limit, and integrated into the `object_store` module in order to combine it with disk cache to serve as a two-tier cache.

jiacai2050:
- https://github.com/facebook/rocksdb/wiki/Block-Cache

There  are two kinds of cache used in RocksDB, LRU and Clock.

LRU is simple to start with, so I plan to implement one based on https://github.com/jiacai2050/lru-weighted-cache.

Key is `${sst-path}-${range}`, value is bytes, also this cache will support shard to reduce lock contention

jiacai2050:
There will be two kinds of caches: one is memory, the other is disk. 

Both is limited by `capacity` configured when started, and use lru to evict inactive ones.

# Memory

    range = [start_offset, end_offset)
    key = sst_id + range
    value = bytes


# Disk

    key = sst_id + align_range(named blocks)
    value = bytes

`align_range` is used for reduce file number in case when range is too small,
and a get request's range will be split into multiple blocks.

For example, when `align_range` is 4KB, then a get request `(4000, 5000)` will cause
the both `[0, 4096)` and `[4096, 8192)` to be cached,

Rachelint:
Maybe memory cache should be aligned?

jiacai2050:
https://github.com/tikv/rocksdb/blob/0b1b80c4927bccbf0b318cdbd9cde6a1e563ae6f/cache/lru_cache.cc#L356

[...]

PR #392, opened 2022-11-14
feat: meta cache for both custom and parquet metadata

ShiKaiWi:

Part of #391 

# Rationale for this change
See #391 

- Add a new meta cache for both custom and parquet metadata;
- Remove useless code, including metadata cache and cacheable reader in `parquet_ext` module;

Faster query.

# How does this change test
By existing tests.


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -143,72 +145,80 @@ impl<'a> Reader<'a> {
...
-                get_metadata_from_storage().await?
+        }
+        // The metadata can't be found in the cache, and let's fetch it from the
+        // storage.
+
+        let meta_data = {
+            let object_meta = storage.head(path).await.context(ObjectStoreError {})?;
+            let parquet_meta_data =
+                Self::load_meta_data_from_storage(storage, &object_meta).await?;
+            let kv_metas = parquet_meta_data
```
jiacai2050, line 189: It seems this KV is saved twice, is there any possible to reduce this?

ShiKaiWi: It's true. Maybe we can reset the extended kv metas in parquet metadata, because they are already decoded in the custom metadata.

How do you like such reset?

jiacai2050: Yeah, exteneded KV is mainly used by ourselves, so we can erase them before saving to cache.

ShiKaiWi: I've fixed this already. PTAL again.

jiacai2050 (approved):
LGTM

PR #393, opened 2022-11-15
refactor: support multiple writing in a specific table

Rachelint:

Closes #

# Rationale for this change
 Now, we assume the thread model: "one writer to one table" in wal module, that is no reasonable. So, this pr refactor the `RegionMeta` to `RegionContext` for removing above assumption.

+ refactor `RegionMeta` to `RegionContext`.

None.

# How does this change test
Test by ut.


wal/src/message_queue_impl/region_context.rs
```diff
@@ -0,0 +1,660 @@
...
+        self.meta.get_meta_data().await
+    }
+}
+
+/// Wrapper for the [TableMetaInner].
+#[derive(Debug)]
+struct TableMeta {
+    table_id: TableId,
+    /// The race condition may occur between writer thread
+    /// and background flush tread.
```
ShiKaiWi, line 255:
```suggestion
    /// and background flush thread.
```

wal/src/message_queue_impl/region_context.rs
```diff
@@ -0,0 +1,660 @@
...
+
+/// Table meta data, will be updated atomically by mutex.
+#[derive(Debug, Default)]
+struct TableMetaInner {
+    /// Next sequence number for the new log.
+    ///
+    /// It will be updated while having pushed logs successfully.
+    next_sequence_num: SequenceNumber,
+
+    /// The lasted marked deleted sequence number. The log with
```
ShiKaiWi, line 371:
```suggestion
    /// The lastest marked deleted sequence number. The log with
```

wal/src/message_queue_impl/region_context.rs
```diff
@@ -0,0 +1,660 @@
...
+/// NOTICE: Will lock it before writing to make writing to a specific table
+/// sequentially.
+#[derive(Debug)]
+pub struct TableWriter;
+
+impl TableWriter {
+    async fn write<M: MessageQueue>(
+        &self,
+        table_meta: &TableMeta,
+        _ctx: &manager::WriteContext,
```
ShiKaiWi, line 554: Why not remove this parameter if the `ctx` is not needed.

Rachelint: I plan to add necessary `timeout` to every operation later.

ShiKaiWi (approved):

PR #395, opened 2022-11-15
feat: support get_range in aliyun oss

### LINKED ISSUE #365: feat: impl `get_range` func with aliyun oss
### Describe This Problem

I found `get_range` func is not implemented in aliyun oss.
https://github.com/CeresDB/ceresdb/blob/b566c57d9c553f9885d7675859f1b4f3e89716fa/components/object_store/src/aliyun.rs#L115-L120

### Proposal

Refer to [doc](https://help.aliyun.com/document_detail/84825.html) of aliyun oss.

### Additional Context

_No response_

ShiKaiWi:

Closes #365 

# Rationale for this change
It is a basic and common feature of OSS to get range of the object file.

- Support get range of object for aliyun oss

None.

# How does this change test
Add a new test for header making.

### DIFF
--- a/components/object_store/src/aliyun.rs
+++ b/components/object_store/src/aliyun.rs
@@ -49,6 +49,8 @@ pub struct AliyunOSS {
 }
 
 impl AliyunOSS {
+    const RANGE_KEY: &str = "Range";
+
     pub fn new(
         key_id: impl Into<String>,
         key_secret: impl Into<String>,
@@ -62,6 +64,13 @@ impl AliyunOSS {
             bucket.into(),
         );
         Self { oss }
+    }
+
+    fn make_range_header(range: Range<usize>, headers: &mut HashMap<String, String>) {
+        assert!(!range.is_empty());
+        let range_value = format!("bytes={}:{}", range.start, range.end - 1);
+
+        headers.insert(Self::RANGE_KEY.to_string(), range_value);
     }
 }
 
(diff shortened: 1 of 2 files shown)


components/object_store/src/aliyun.rs
```diff
@@ -112,11 +121,23 @@ impl ObjectStore for AliyunOSS {
...
+        let mut headers = HashMap::new();
+        Self::make_range_header(range, &mut headers);
+
+        let bytes = self
+            .oss
+            .get_object(&location.to_string(), Some(headers), None)
+            .await
+            .context(GetObject {
+                path: &location.to_string(),
+            })?;
```
Rachelint, line 138: How about including the `range` to error?

components/object_store/src/aliyun.rs
```diff
@@ -112,11 +121,23 @@ impl ObjectStore for AliyunOSS {
         Ok(GetResult::Stream(stream::once(async { Ok(bytes) }).boxed()))
     }
 
-    async fn get_range(&self, _location: &Path, _range: Range<usize>) -> Result<Bytes> {
-        Err(Error::Unimplemented {
-            op: "get_range".to_string(),
+    async fn get_range(&self, location: &Path, range: Range<usize>) -> Result<Bytes> {
```
jiacai2050, line 124: There is a same name method in `cache.rs`, it currently invoke `get` to fetch data, after this PR, it can invoke `get_range` directly.

ShiKaiWi: Let me fix it in this pr.

ShiKaiWi: I find that `CachedStore` has to cache the whole file, even if only a range of file fetched. I will solve this in #392.

jiacai2050: > has to cache the whole file

Any explain for this? I don't remember it has this requirement.

ShiKaiWi: Actually, the granularity of the `CachedStore` is file level, and the code about the `get_range` of `Cache` is:
```rust
    async fn get_range(&self, location: &Path, range: Range<usize>) -> Result<Bytes> {
        if self.state.lock().await.contains(location) {
            self.local_store.get_range(location, range).await
        } else {
            let remote_obj = self.remote_store.get(location).await?;
            let bytes = remote_obj.bytes().await?;
            let _ = self.try_put_local(location, bytes).await?;
            self.local_store.get_range(location, range).await
        }
    }
```

jiacai2050 (approved):
LGTM

PR #399, opened 2022-11-16
fix: add meta client timeout

ZuLiangWang:
None.

# Rationale for this change
CeresDB interacts with CeresMeta through grpc implemented in `meta_client`, and now the timeout of grpc is not set, this will cause CeresDB to be blocked when the network fails.

* Add timeout config in `meta_client`.

None.

# How does this change test
Local manual test.

### DIFF
--- a/meta_client/src/meta_impl.rs
+++ b/meta_client/src/meta_impl.rs
@@ -61,7 +61,8 @@ impl MetaClientImpl {
                 .map_err(|e| Box::new(e) as _)
                 .context(FailConnect {
                     addr: &config.meta_addr,
-                })?;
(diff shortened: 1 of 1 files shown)


meta_client/src/meta_impl.rs
```diff
@@ -11,6 +11,7 @@ use common_util::config::ReadableDuration;
 use log::{debug, info};
 use serde_derive::Deserialize;
 use snafu::{OptionExt, ResultExt};
+use tokio::time::timeout;
```
ShiKaiWi, line 14: Remove this unused import.

PR #400, opened 2022-11-17
feat: refacor existing wal implementations to support scanning logs of shard

### LINKED ISSUE #389: let exist wal implementations support scanning logs of whole shard
### Describe This Problem

In exist wal implementations, we can just read logs table by table, it lead to duplicated io(rocksdb based), and a huge number of requests(obkv based).
So, I want to introduce the mapping of `shard to region` to them, and make them support shard based reading. 

### Proposal

+ Logs in exist wal implementations are still managed on table level, the local manager doesn't perceive the concept about `shard`(now mapped to `region`).
+ The information about shard will be passed with every necessary function call. 

Cite a case to express the above design. The `shard` information(always `shard id`), is used a bit like `user name`, `tenant`, it is just a pure logic concept to exist wal(in kafka wal, it is a concept about physical unit).

##### Obkv logs cleaning
Now, we can't know what obkv tables logs of tables are stored in:
  + now, `table id` decide logs of ceresdb table will be written to which obkv table. But for supporting `shard based read/write`, it's necessary to use  `shard id` to make this decision.
  + howerver, table may be moved between `shard`s in cluster mode(`shard id` will changed), and there aren't ways to track such moving now.

Above problem make trouble to log cleaning. And we can just relax this problem in following ways:
  + traverse all obkv tables to find logs which should be deleted.
  + just run the logs cleaning routine in current shard.

All of them are not good ways... for simplicity, we implement the second way now.

Rachelint:

Closes #389 

# Rationale for this change
 
See #389.

+ Replace old log format with common log format.
+ Rename `Region` to `TableUnit`.
+ Support scanning in two existing wal implementations.

None.

# How does this change test
Test by ut.

### DIFF
--- a/wal/src/table_kv_impl/namespace.rs
+++ b/wal/src/table_kv_impl/namespace.rs
@@ -507,7 +507,7 @@ impl<T: TableKv> NamespaceInner<T> {
         self.open_table_unit(region_id, table_id).await
     }
 
-    // TODO(yingwen): Provide a close_talbe_unit() method.
+    // TODO(yingwen): Provide a close_table_unit() method.
     async fn open_table_unit(
         &self,
         region_id: RegionId,
--- a/wal/src/table_kv_impl/namespace.rs
+++ b/wal/src/table_kv_impl/namespace.rs
@@ -9,7 +9,10 @@ use std::{
     time::Duration,
 };
 
-use common_types::time::Timestamp;
+use common_types::{
+    table::{Location, TableId},
+    time::Timestamp,
+};
 use common_util::{config::ReadableDuration, define_result, runtime::Runtime};
 use log::{debug, error, info};
 use snafu::{Backtrace, OptionExt, ResultExt, Snafu};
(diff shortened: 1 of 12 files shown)


wal/src/table_kv_impl/namespace.rs
```diff
@@ -465,83 +490,114 @@ impl<T: TableKv> NamespaceInner<T> {
...
         }
 
-        self.open_region(region_id).await
+        self.open_table_unit(region_id, table_id).await
     }
 
-    // TODO(yingwen): Provide a close_region() method.
-    async fn open_region(&self, region_id: RegionId) -> Result<Option<RegionRef>> {
-        let region_meta_table = self.region_meta_table(region_id);
+    // TODO(yingwen): Provide a close_talbe_unit() method.
```
ShiKaiWi, line 510:
```suggestion
    // TODO(yingwen): Provide a close_table_unit() method.
```

ShiKaiWi (commented):
It's a great job. Let's fix some typos and merge it.

ShiKaiWi (approved):
LGTM

PR #404, opened 2022-11-17
feat: Impl wal on mq namespace

### LINKED ISSUE #371: Implement region namespace
**Description**
Implement the region manager for region module which has been implemented(see #367).

**Proposal**
It include two phase:
+ Implement the region's recovery by its meta snapshot.
+ Implement the `WalManager` for `RegionManager`. 

**Additional context**
##### When should we create the new region entry?
The new region is possible to be created in following operation:
+ Write.
+ Read(scan region or table).
+ Mark deleted.

It doesn't matter to do this, we just need to consider the behavior while write/read/mark deleted on a empty region and it is not a complex work.

##### When should we create the new table entry in region?
For table entry, it will just be created and inserted in `write` operation. And while not found it in other operation, error will occur.

Rachelint:

Closes #371 

# Rationale for this change
 
See #371 .

+ implement `Namespace` of wal on mq.
+ implement the `WalManager` interface for it.

You can start ceresDB with wal on kafka now!

# How does this change test
Test by ut.

### DIFF
--- a/wal/src/message_queue_impl/namespace.rs
+++ b/wal/src/message_queue_impl/namespace.rs
@@ -1,0 +1,571 @@
+use std::{collections::HashMap, fmt, sync::Arc};
+
+use common_types::{table::Location, SequenceNumber};
+use common_util::define_result;
+use log::{debug, info};
+use message_queue::{ConsumeIterator, MessageQueue};
+use snafu::{ensure, Backtrace, OptionExt, ResultExt, Snafu};
+use tokio::sync::RwLock;
+
+use super::{
+    encoding::MetaEncoding,
+    region::{self, MessageQueueLogIterator, Region},
+};
+use crate::{
+    kv_encoder::LogEncoding,
+    log_batch::{LogEntry, LogWriteBatch},
+    manager::{ReadContext, ReadRequest, RegionId, ScanContext, ScanRequest, WriteContext},
+};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display(
+        "Open region failed, namespace:{}, location:{:?}, err:{}",
+        namespace,
+        location,
+        source
+    ))]
+    GetSequence {
+        namespace: String,
+        location: Location,
+        source: region::Error,
+    },
+
+    #[snafu(display(
+        "Open region failed, namespace:{}, request:{:?}, err:{}",
+        namespace,
+        request,
+        source
+    ))]
+    ReadWithCause {
+        namespace: String,
+        request: ReadRequest,
+        msg: String,
+        source: region::Error,
+    },
+
+    #[snafu(display(
+        "Open region failed, namespace:{}, request:{:?}, \nBacktrace:\n{}",
+        namespace,
+        request,
+        backtrace,
+    ))]
+    ReadNoCause {
+        namespace: String,
+        request: ReadRequest,
+        msg: String,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display(
+        "Open region failed, namespace:{}, request:{:?}, err:{}",
+        namespace,
+        request,
+        source
+    ))]
+    ScanWithCause {
+        namespace: String,
+        request: ScanRequest,
+        msg: String,
+        source: region::Error,
+    },
+
+    #[snafu(display(
+        "Open region failed, namespace:{}, request:{:?}, \nBacktrace:\n{}",
+        namespace,
+        request,
+        backtrace,
+    ))]
+    ScanNoCause {
+        namespace: String,
+        request: ReadRequest,
+        msg: String,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display(
+        "Open region failed, namespace:{}, location:{:?}, batch_size:{}, err:{}",
+        namespace,
+        location,
+        batch_size,
+        source
+    ))]
+    Write {
+        namespace: String,
+        location: Location,
+        batch_size: usize,
+        source: region::Error,
+    },
+
+    #[snafu(display(
+        "Open region failed, namespace:{}, location:{:?}, sequence_num:{}, err:{}",
+        namespace,
+        location,
+        sequence_num,
+        source
+    ))]
+    MarkDeleteTo {
+        namespace: String,
+        location: Location,
+        sequence_num: SequenceNumber,
+        source: region::Error,
+    },
+}
+
+define_result!(Error);
+
+#[allow(unused)]
+pub struct Namespace<M: MessageQueue> {
+    namespace: String,
+    regions: RwLock<HashMap<RegionId, RegionRef<M>>>,
+    message_queue: Arc<M>,
+    meta_encoding: MetaEncoding,
+    log_encoding: LogEncoding,
+}
+
+// TODO: more information should be included.
+// Need a solution to get information of regions guarded by tokio's `RwLock`.
+// Also implement `Debug` for `MessageQueue`.
+impl<M: MessageQueue> fmt::Debug for Namespace<M> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Namespace")
+            .field("namespace", &self.namespace)
+            .field("message_queue", &self.message_queue)
+            .field("meta_encoding", &self.meta_encoding)
+            .field("log_encoding", &self.log_encoding)
+            .finish()
+    }
+}
+
+#[allow(unused)]
+impl<M: MessageQueue> Namespace<M> {
+    pub fn open(namespace: String, message_queue: Arc<M>) -> Self {
+        Self {
+            namespace,
+            regions: RwLock::new(HashMap::new()),
+            message_queue,
+            meta_encoding: MetaEncoding::newest(),
+            log_encoding: LogEncoding::newest(),
+        }
+    }
+
+    /// Get or open the region of `region_id`.
+    ///
+    /// NOTICE: If the region doesn't exist yet, it will be created no matter
+    /// the high-level caller is `write`, `scan`, `read_batch` or
+    /// `mark_delete_entries_up_to` . So, we should consider situations
+    /// about above operations on an empty region.
+    async fn get_or_open_region(
+        &self,
+        region_id: RegionId,
+    ) -> std::result::Result<RegionRef<M>, region::Error> {
+        {
+            let regions = self.regions.read().await;
+            if let Some(region) = regions.get(&region_id) {
+                debug!(
+                    "Region exists and return it, namespace:{}, region id:{}",
+                    self.namespace, region_id
+                );
+                return Ok(region.clone());
+            }
+        }
+
+        let mut regions = self.regions.write().await;
+        // Multiple tables share one region, so double check here is needed.
+        if let Some(region) = regions.get(&region_id) {
+            debug!(
+                "Region exists and return it, namespace:{}, region id:{}",
+                self.namespace, region_id
+            );
+            return Ok(region.clone());
+        }
+
+        let region =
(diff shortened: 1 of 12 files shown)


wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+        debug!(
+            "Write table logs in namespace, namespace:{}, ctx:{:?}, location:{:?}, batch size:{}",
+            self.namespace,
+            ctx,
+            log_batch.location,
+            log_batch.entries.len()
+        );
+
+        let region = self
+            .get_or_open_region(log_batch.location.shard_id as RegionId)
```
ShiKaiWi, line 413: Actually, I wonder why the `shard_id` is not called `region_id` in the `Location`? `ShardId` should not be the known about the Wal module.

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+            is_terminated: false,
+        }
+    }
+
+    pub async fn next_log_entry(&mut self) -> Result<Option<LogEntry<&'_ [u8]>>> {
+        // If inner iter is `Some`, poll it to get required log entry.
+        if let Some(iter) = self.iter.as_mut() {
+            if self.is_terminated {
+                return Ok(None);
+            }
```
ShiKaiWi, line 517: Maybe we should do this check first?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+    iter: Option<MessageQueueLogIterator<C>>,
+
+    /// Request triggering this scanning
+    request: ScanRequest,
+
+    /// Terminated flag
+    is_terminated: bool,
+
+    /// See the same problem in https://github.com/CeresDB/ceresdb/issues/120
+    previous_value: Vec<u8>,
```
ShiKaiWi, line 488: Shall we call it `current_log_payload`?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+                    request: self.request.clone(),
+                    msg: "failed while polling log",
+                })
+                .map(|log_entry| {
+                    if log_entry.is_none() {
+                        self.is_terminated = true;
+                    }
+
+                    log_entry.map(|log_entry| {
+                        self.previous_value = log_entry.payload.to_owned();
```
ShiKaiWi, line 532: it seems no need to call `to_owned()`?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+                    //  + the table id equals to the one in `request`
+                    //  + the sequence number of log entry is among [start, end], return this log
+                    // entry. Otherwise, just continue return the `log_entry`.
+                    // Otherwise, just continue to poll.
+                    if log_entry.table_id == self.request.location.table_id
+                        && log_entry.sequence >= start_sequence
+                        && log_entry.sequence <= end_sequence
+                    {
+                        self.is_terminated = log_entry.sequence == self.end;
+                        self.previous_value = log_entry.payload.to_owned();
```
ShiKaiWi, line 659: Is it necessary to clone?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+            is_terminated: false,
+            iter: Some(iter),
+            previous_value: Vec::new(),
+        })
+    }
+
+    pub async fn next_log_entry(&mut self) -> Result<Option<LogEntry<&'_ [u8]>>> {
+        // If inner iter is `Some`, poll it to get required log entry.
+        if let Some(iter) = self.iter.as_mut() {
+            if self.is_terminated {
```
ShiKaiWi, line 635: Shall do the check at the beginning?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+    inner: Arc<NamespaceInner<M>>,
+) -> TaskHandle {
+    let name = format!("Log-cleaner-wal-on-mq-{}", inner.namespace);
+    let builder = move || {
+        let inner = inner.clone();
+
+        log_cleaner_routine(inner)
+    };
+
+    TimedTask::start_timed_task(name, runtime, period, builder)
```
ShiKaiWi, line 713: I don't see when to stop such task.

Rachelint: It will be stopped while `Namespace::close` is called. 
I guess we design the method `close_gracefully` in interface is for doing this, but unfortunately I found it is just dead code now. 
I will implement drop to calling make the routine stop temporarily, and see how to make use of `close_gracefully` later.

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+        );
+
+        Ok(region)
+    }
+
+    pub async fn sequence_num(&self, location: Location) -> Result<SequenceNumber> {
+        let region = self
+            .get_or_open_region(location.shard_id as u64)
+            .await
+            .context(GetSequence {
```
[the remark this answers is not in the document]
Rachelint: Yes, I notice it, I plan to do this in a new pr.
Cause it needs to modify the `MessageQueue` interface, too.

wal/src/message_queue_impl/region_context.rs
```diff
@@ -383,6 +388,17 @@ struct TableMetaInner {
     start_sequence_offset_mapping: BTreeMap<SequenceNumber, Offset>,
 }
 
+impl Default for TableMetaInner {
+    fn default() -> Self {
+        Self {
+            next_sequence_num: SequenceNumber::MIN + 1,
```
jiacai2050, line 394: Why this can't be `Default::default()`?

Add some comments explain this.

wal/src/message_queue_impl/wal.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use common_types::{table::Location, SequenceNumber};
+use common_util::runtime::Runtime;
+use message_queue::{ConsumeIterator, MessageQueue};
+use snafu::ResultExt;
+
+use super::{
```
jiacai2050, line 13: remove `super` use

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+        encoding::MetaEncoding,
+        region::{self, MessageQueueLogIterator, Region},
+    },
+    table_kv_impl::timed_task::{TaskHandle, TimedTask},
+};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display(
+        "Open region failed, namespace:{}, location:{:?}, err:{}",
```
jiacai2050, line 29: Other module's error message is like `Failed to ...`, so it's best this module follow this style.

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+pub struct Namespace<M: MessageQueue> {
+    /// Namespace inner
+    inner: Arc<NamespaceInner<M>>,
+
+    /// Handle for cleaner routine
+    cleaner_handle: TaskHandle,
+
+    /// Background runtime for cleaner routine
+    ///
+    /// Keep it here to ensure it won't be drop during the lifetime of
```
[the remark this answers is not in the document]
jiacai2050: Is this fixed?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+        regions.clear();
+
+        self.cleaner_handle.stop_task().await.context(Close {
+            namespace: self.inner.namespace.clone(),
+        })
+    }
+
+    /// Get table's sequence number.
+    ///
+    /// You can define the read range in `ReadRequest`.
```
jiacai2050, line 184: This line is seems useless here. any reason keep here?

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+    message_queue: Arc<M>,
+    meta_encoding: MetaEncoding,
+    log_encoding: LogEncoding,
+}
+
+impl<M: MessageQueue> NamespaceInner<M> {
+    pub fn new(namespace: String, message_queue: Arc<M>) -> Self {
+        Self {
+            namespace,
+            regions: Arc::new(RwLock::new(HashMap::new())),
```
jiacai2050, line 257: `Default:default()` is enough

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+    pub async fn sequence_num(&self, location: Location) -> Result<SequenceNumber> {
+        let region = self
+            .get_or_open_region(location.shard_id as u64)
+            .await
+            .context(GetSequence {
+                namespace: self.namespace.clone(),
+                location,
+            })?;
+
+        Ok(region
```
jiacai2050, line 316: It's a little weird to wrap so many logic inside `Ok(..)`, could you try simplify it?

wal/src/message_queue_impl/namespace.rs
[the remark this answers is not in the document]
jiacai2050: Leave a TODO here if you want to fix this future.

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+    ) -> Result<ReadTableIterator<M::ConsumeIterator>> {
+        debug!(
+            "Read table logs in namespace, namespace:{}, ctx:{:?}, request:{:?}",
+            self.namespace, ctx, request
+        );
+
+        let region = self
+            .get_or_open_region(request.location.shard_id as u64)
+            .await
+            .context(ReadWithCause {
```
jiacai2050, line 341: Same with before

wal/src/message_queue_impl/namespace.rs
```diff
@@ -0,0 +1,714 @@
...
+pub struct ReadTableIterator<C: ConsumeIterator> {
+    /// Namespace name
+    namespace: String,
+
+    /// Log iterator
+    ///
+    /// It will return some logs belonging to other tables or belonging to
+    /// current table but with a out-of-range sequence number, so we need to
+    /// filter before returning.
+    /// It can be `None`, while there isn't any log of the table in the region
```
jiacai2050, line 557: when there ...

jiacai2050 (changes requested):

wal/src/message_queue_impl/namespace.rs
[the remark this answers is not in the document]
Rachelint: It has been used for `drop`.

wal/src/message_queue_impl/namespace.rs
Rachelint: I want to do  this together while fixing the bug in obkv implementation.

wal/src/message_queue_impl/namespace.rs
Rachelint: It needs, the borrow check isn't smart enough to know the &[u8] is actually owed by current iterator ... So in last pr about this, I plan to return `Vec<u8>` instead of `&[u8]` from the inner iterator.

Actually, return &[u8] is just a effective optimization in rocksdb implementation. I found it make trouble in other implements temporarily.

wal/src/message_queue_impl/namespace.rs
Rachelint: Same with above.

jiacai2050 (approved):
LGTM

PR #401, opened 2022-11-17
chore: update obkv-client to latest version

chunshao90:
# Which issue does this PR close?

Closes #

# Rationale for this change
 To fix bug mentioned in https://github.com/oceanbase/obkv-table-client-rs/pull/5.

# What changes are included in this PR?

# Are there any user-facing changes?
No.

# How does this change test

PR #405, opened 2022-11-17
fix: wrong cargo lock

ShiKaiWi:

Closes #

# Rationale for this change
The CI of #401 was skipped, leading to that the update of Cargo.lock is not included.

- Update the Cargo.lock.
- Move the datafusion part to sub separate module in `common_types::time` for ceresdb-client-rs to use.

None.

# How does this change test
Existing UT.


Rachelint (approved):
LGTM.

PR #406, opened 2022-11-18
chore: update the ceresdbproto to latest

ShiKaiWi:
# Which issue does this PR close?

Closes #

# Rationale for this change
 Prepare for remove manual installation of protoc in the ci workflow.

# What changes are included in this PR?
Update the version of ceresdbproto.

# Are there any user-facing changes?
None.

# How does this change test
Existing unit test.

PR #407, opened 2022-11-18
chore: remove manual installation of protoc

ShiKaiWi:

Closes #

# Rationale for this change
 #406 has updated the ceresdbproto to avoid its requirement for protoc, but the `proto` still needs the protoc. Let's introduce the `protoc-bin-vendored` for automatic installation of protoc.

- Remove the installation of protoc.
- Introduce `protoc-bin-vendored` to install protoc by the crate itself.

None.

# How does this change test
Existing test.


jiacai2050 (approved):
LGTM

PR #411, opened 2022-11-18
test: add case sensitive demo

### LINKED ISSUE #220: Create/insert case-sensitive table name success, but not found while select.
### Describe this problem

Create/insert a table named like "DeMo" will success.
But when I select something in this table, error of "table not found" will be returned.

### Steps to reproduce

+ create table
```
CREATE TABLE DeMo (
    `timestamp` timestamp NOT NULL,
    `arch` string TAG,
    `datacenter` string TAG,
    `value` int,
    timestamp KEY (timestamp)) ENGINE=Analytic
WITH(
	 enable_ttl='false'
);
```
+ Insert something:
```
INSERT INTO DeMo
    (`timestamp`, `arch`, `datacenter`, `value`)
VALUES
    (1658304762, 'x86-64', 'china', 100),
```

+ Select something:
```
SELECT * FROM DeMo;
```

### Expected behavior

_No response_

### Additional Information

_No response_

jiacai2050:

Closes #220 

# Rationale for this change
 

- Add case-sensitive testcase
- Update docs explain this
- Format SQL in docs

No

# How does this change test

Test harness

### DIFF
--- a/docs/guides/src/quick_start.md
+++ b/docs/guides/src/quick_start.md
@@ -27,31 +27,52 @@ The easiest to use is HTTP, so sections below will use it for demo. For producti
 ### Create table
 ```shell
 curl --location --request POST 'http://127.0.0.1:5440/sql' \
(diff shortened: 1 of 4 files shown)


Rachelint (approved):
LGTM.

PR #414, opened 2022-11-20
refactor: recovery in standalone mode

### LINKED ISSUE #413: Recovery process in standalone mode may be a bit strange
### Describe This Problem

In standalone mode, we recover table in `Visitor::visit_tables`, as I see, this design is a bit strange. 
I don't think this design will bring any benefits, on the contrary it will make trouble while doing enhancement to recovery, such as recovering table concurrently.
So I propose to refactor it.

### Proposal

+ we just collect the `TableInfo` in `visit_tables`.
+ modify `SchemaImpl::open_table`(table_based) to carry out the actual opening work.
+ we pass the collecting results to above opening method.

### Additional Context

_No response_

Rachelint:

Closes #413 

# Rationale for this change
See #413  

+ modify the `VisitorImpl::visit_tables` to just collect the table infos.
+ move the actual recovery work to `SchemaImpl::open_table`.

None.

# How does this change test
Test by exist ut.

### DIFF
--- a/catalog_impls/src/table_based.rs
+++ b/catalog_impls/src/table_based.rs
@@ -94,8 +94,6 @@ pub struct TableBasedManager {
     catalogs: CatalogMap,
     /// Global schema id generator, Each schema has a unique schema id.
     schema_id_generator: Arc<SchemaIdGenerator>,
-    /// Collect table
-    table_infos: Vec<TableInfo>,
 }
 
 impl Manager for TableBasedManager {
@@ -131,7 +129,6 @@ impl TableBasedManager {
             catalog_table: Arc::new(catalog_table),
             catalogs: HashMap::new(),
             schema_id_generator: Arc::new(SchemaIdGenerator::default()),
-            table_infos: Vec::new(),
         };
 
         manager.init().await?;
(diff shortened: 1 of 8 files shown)


src/setup.rs
```diff
@@ -192,16 +196,39 @@ async fn build_in_standalone_mode<Q: Executor + 'static>(
...
+            .unwrap_or_else(|_| panic!("fail to get schema of table, table_info:{:?}", table_info))
+            .unwrap_or_else(|| panic!("schema of table is not found, table_info:{:?}", table_info));
+
+        let open_request = OpenTableRequest::from(table_info);
+        schema
+            .open_table(open_request.clone(), opts.clone())
+            .await
+            .unwrap_or_else(|_| panic!("fail to open table, open_request:{:?}", open_request))
+            .unwrap_or_else(|| panic!("no table is opened, open_request:{:?}", open_request));
+    }
```
chunshao90, line 230: This part of the code is not suitable here. Define a function like `reopen_table`?

catalog_impls/src/table_based.rs
```diff
@@ -130,18 +130,17 @@ impl TableBasedManager {
...
         manager.init().await?;
 
         Ok(manager)
     }
 
-    #[cfg(test)]
-    pub fn get_engine_proxy(&self) -> TableEngineRef {
-        self.engine_proxy.clone()
+    pub fn get_table_infos(&self) -> Vec<TableInfo> {
+        self.table_infos.clone()
```
ShiKaiWi, line 143: I guess `table_infos` field is not a good idea because the field won't change with following drop/create table operation and that will lead to confusion.

Here, I'm in favor of using `VistorImpl` to fetch table infos dynamically from the backend engine.

Rachelint: Ok, it's actually strange here.

ShiKaiWi (approved):
LGTM

PR #415, opened 2022-11-23
feat: add memory based cache for object store

jiacai2050:

Part of #391

# Rationale for this change

Add CachedStore in object store module, it can
- evict item based on configured memory capacity
- support partition to reduce lock contention

No

# How does this change test

New unit test
- `test_mem_cache_evict`
- `test_mem_cache_partition`

### DIFF
--- a/components/object_store/src/mem_cache.rs
+++ b/components/object_store/src/mem_cache.rs
@@ -202,16 +202,17 @@ mod test {
 
     use super::*;
 
-    async fn prepare_store(mem_cap: usize) -> CachedStore {
+    async fn prepare_store(bits: usize, mem_cap: usize) -> CachedStore {
         let local_path = tempdir().unwrap();
         let local_store = Arc::new(LocalFileSystem::new_with_prefix(local_path.path()).unwrap());
 
-        CachedStore::new(0, mem_cap, local_store)
+        CachedStore::new(bits, mem_cap, local_store)
     }
 
     #[tokio::test]
     async fn test_mem_cache_evict() {
-        let store = prepare_store(13).await;
+        // single partition
+        let store = prepare_store(0, 13).await;
 
         // write date
         let location = Path::from("1.sst");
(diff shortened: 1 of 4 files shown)


components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+            .collect::<Vec<_>>();
+
+        Self {
+            mem_cap,
+            partitions,
+            partition_mask: partition_num - 1,
+        }
+    }
+
+    fn locate_partition(&self, key: &String) -> Arc<Partition> {
```
Rachelint, line 78: Why not make the `hasher` as a trait object and put it in `MemCache`. 
If so, it may be easy to do this  when we want a new hasher.

components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+
+impl Partition {
+    fn new(mem_cap: usize) -> Self {
+        Self {
+            inner: Mutex::new(LruWeightedCache::new(1, mem_cap).expect("invalid params")),
+        }
+    }
+}
+impl Partition {
+    async fn get(&self, key: &String) -> Option<Bytes> {
```
Rachelint, line 43: &String to &str

components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    fn locate_partition(&self, key: &String) -> Arc<Partition> {
+        let mut hasher = DefaultHasher::new();
+        key.hash(&mut hasher);
+        self.partitions[hasher.finish() as usize & self.partition_mask].clone()
+    }
+
+    async fn get(&self, key: &String) -> Option<Bytes> {
```
Rachelint, line 84: &String to &str

components/object_store/src/mem_cache.rs
Rachelint: &String to &str

components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        partition.get(key).await
+    }
+
+    async fn insert(&self, key: String, value: Bytes) {
+        let partition = self.locate_partition(&key);
+        partition.insert(key, value).await;
+    }
+}
+
+impl Display for MemCache {
```
Rachelint, line 95: Why impl `Display` for it rather than `Debug`? 
Do we need to call `to_string` in somewhere? If just debug, maybe impl `Debug` is enough?

components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+            underlying_store,
+        }
+    }
+
+    fn cache_key(location: &Path, range: &Range<usize>) -> String {
+        format!("{}-{}-{}", location, range.start, range.end)
+    }
+}
+
+impl Display for CachedStore {
```
Rachelint, line 128: Same as above.

components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.underlying_store
+            .abort_multipart(location, multipart_id)
+            .await
+    }
+
+    async fn get(&self, location: &Path) -> Result<GetResult> {
+        self.underlying_store.get(location).await
+    }
+
+    async fn get_range(&self, location: &Path, range: Range<usize>) -> Result<Bytes> {
```
Rachelint, line 157: Add a todo for range checking?

components/object_store/src/mem_cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.underlying_store.put_multipart(location).await
+    }
+
+    async fn abort_multipart(&self, location: &Path, multipart_id: &MultipartId) -> Result<()> {
+        self.underlying_store
+            .abort_multipart(location, multipart_id)
+            .await
+    }
+
+    async fn get(&self, location: &Path) -> Result<GetResult> {
```
Rachelint, line 153: How about adding comments to explain why we don't cache it while get all.

components/object_store/src/mem_cache.rs
jiacai2050: It's required for ObjectStore trait
> pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static

components/object_store/src/mem_cache.rs
jiacai2050: > it may be easy to do this when we want a new hasher.

I couldn't think of a situation to switch hasher, if we want, we can just change this here.

I don't want to expose a generic type here.

jiacai2050: > &String to &str

`&String` required by [get](https://github.com/jiacai2050/lru-weighted-cache/blob/master/src/lib.rs#L210)

components/object_store/src/mem_cache.rs
jiacai2050: fixed

components/object_store/src/mem_cache.rs
jiacai2050: fixed

components/object_store/src/mem_cache.rs
Rachelint: Maybe we can leave a comment to explain this.

Rachelint (approved):
LGTM.

PR #416, opened 2022-11-24
feat: Impl wal on mq setup

Rachelint:

Closes #

# Rationale for this change
Wal on mq has been implemented, we just add it to setup now. 

Add wal on mq to setup.

User can setup ceresdb with wal on kafka now.

# How does this change test
Manual check.

### DIFF
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -23,11 +23,15 @@ mod wal_synchronizer;
 #[cfg(any(test, feature = "test"))]
 pub mod tests;
 
+use message_queue::kafka::config::Config as KafkaConfig;
 use meta::details::Options as ManifestOptions;
 use serde_derive::Deserialize;
 use storage_options::{LocalOptions, StorageOptions};
 use table_kv::config::ObkvConfig;
-use wal::table_kv_impl::model::NamespaceConfig;
+use wal::{
+    message_queue_impl::config::Config as MessageQueueWalConfig,
+    table_kv_impl::model::NamespaceConfig,
+};
 
 pub use crate::{compaction::scheduler::SchedulerConfig, table_options::TableOptions};
 
@@ -72,8 +76,13 @@ pub struct Config {
     // Batch size for scan sst
     pub scan_batch_size: usize,
 
-    // Obkv wal config.
-    pub obkv_wal: ObkvWalConfig,
+    /// Wal storage config
+    ///
+    /// Now, following three storages are supported:
+    /// + RocksDB
+    /// + OBKV
+    /// + Kafka
+    pub wal_storage: WalStorageConfig,
 }
 
 impl Default for Config {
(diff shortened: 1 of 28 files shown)


analytic_engine/src/setup.rs
```diff
@@ -64,6 +69,12 @@ pub enum Error {
         path: String,
         source: std::io::Error,
     },
+
+    #[snafu(display("Failed to open wal on kafka err:{}", source))]
+    OpenKafkaWal {
+        // path: String,
```
jiacai2050, line 75: Delete this

analytic_engine/src/setup.rs
```diff
@@ -141,26 +158,38 @@ impl EngineBuilder for RocksEngineBuilder {
...
 #[async_trait]
-impl EngineBuilder for ReplicatedEngineBuilder {
+impl EngineBuilder for ObkvWalEngineBuilder {
     async fn open_wal_and_manifest(
         &self,
         config: Config,
         engine_runtimes: Arc<EngineRuntimes>,
     ) -> Result<(WalManagerRef, ManifestRef)> {
-        assert!(config.obkv_wal.enable);
+        let obkv_wal_config = match &config.wal_storage {
```
jiacai2050, line 170: `ensure!` works better for this case

analytic_engine/src/lib.rs
```diff
@@ -131,3 +138,23 @@ impl Default for ObkvWalConfig {
...
+    /// Wal config
+    pub wal_config: MessageQueueWalConfig,
+}
+
+/// Options for storage backend
+#[derive(Debug, Clone, Deserialize)]
+#[serde(tag = "type")]
+pub enum WalStorageConfig {
+    RocksDB,
+    Obkv(Box<ObkvWalConfig>),
```
jiacai2050, line 158: Why use box here?

analytic_engine/src/setup.rs
```diff
@@ -180,12 +209,77 @@ impl EngineBuilder for MemWalEngineBuilder {
         config: Config,
         engine_runtimes: Arc<EngineRuntimes>,
     ) -> Result<(WalManagerRef, ManifestRef)> {
-        open_wal_and_manifest_with_table_kv(config, engine_runtimes, self.table_kv.clone()).await
+        let obkv_wal_config = match &config.wal_storage {
+            WalStorageConfig::Obkv(config) => config.clone(),
+            _ => panic!(
```
jiacai2050, line 214: It's better to return error here, since this method's return type is `Result`

analytic_engine/src/lib.rs
Rachelint: > Why use box here?

It is recommended by `clippy`. 
As I see, maybe because `enum` is C's `union` actually, and one config here is much larger than another.

analytic_engine/src/lib.rs
```diff
@@ -131,3 +138,23 @@ impl Default for ObkvWalConfig {
...
+#[serde(default)]
+pub struct KafkaWalConfig {
+    /// Kafka client config
+    pub kafka_config: KafkaConfig,
+
+    /// Wal config
+    pub wal_config: MessageQueueWalConfig,
+}
+
+/// Options for storage backend
```
chunshao90, line 153:
```suggestion
/// Options for storage backend.
```

chunshao90 (approved):
LGTM

PR #418, opened 2022-11-24
fix: cleaner bugs of wal on obkv

Rachelint:

Closes #

# Rationale for this change
 There are cleaner bugs in wal on obkv.

 Fix the cleaner bugs in wal on obkv.

None.

# How does this change test
Test manually.


chunshao90 (approved):
LGTM

PR #421, opened 2022-11-25
feat: setup MemCacheStore

jiacai2050:

Closes #

# Rationale for this change
 
After #415, MemCachedStore is implemented, but not used, this PR add it to setup

1. Remove old data cache
2. Add new mem cache options

No

# How does this change test

Start server locally with config below

```
bind_addr = "127.0.0.1"
http_port = 5440
grpc_port = 8831
log_level = "debug"

[analytic.storage]
mem_cache_capacity = '1G'
mem_cache_partition_bits = 0

[analytic.storage.object_store]
type = "Local"
data_path = "/tmp/xx"

[analytic]
wal_path = "/tmp/xx"
sst_meta_cache_cap = 10000
scan_batch_size = 10000

[[static_route.topology.schema_shards]]
schema = 'public'
auto_create_tables = true

[[static_route.topology.schema_shards.shard_views]]
shard_id = 0

[static_route.topology.schema_shards.shard_views.endpoint]
addr = "127.0.0.1"
port = 8831
```


analytic_engine/src/storage_options.rs
```diff
@@ -1,12 +1,20 @@
...
 
+use common_util::config::ReadableSize;
 use object_store::cache::CachedStoreConfig;
 use serde::Deserialize;
 
+#[derive(Debug, Clone, Deserialize)]
 /// Options for storage backend
+pub struct StorageOptions {
+    pub mem_cache_capacity: ReadableSize,
+    pub mem_cache_partition_bits: usize,
```
Rachelint, line 11: Should we expose cache informations here?

analytic_engine/src/setup.rs
```diff
@@ -270,22 +264,34 @@ fn open_storage(
...
                 let remote_store = open_storage(*cache_opts.remote_store).await?;
                 let store = CachedStore::init(local_store, remote_store, cache_opts.cache_opts)
                     .await
                     .context(OpenObjectStore)?;
-                Ok(Arc::new(store) as _)
+                Arc::new(store) as _
             }
+        };
+
+        if opts.mem_cache_capacity.as_bytes() == 0 {
```
Rachelint, line 285: The method name `as_bytes` is confused...

analytic_engine/src/storage_options.rs
jiacai2050: Yeah, this config set how many partition we want in cache.

analytic_engine/src/setup.rs
jiacai2050: Agree, `as_byte` may better match as method like `as_mb()`

jiacai2050: https://github.com/CeresDB/ceresdb/issues/428

Rachelint (approved):
LGTM.

PR #422, opened 2022-11-26
fix: bug in distributed wal(obkv, kafka)

### LINKED ISSUE #441: Serious meta information overwritten bug of distributed wal in cluster mode
### Describe this problem

If a shard (which is mapped to region in wal) is move between nodes like this:
```
A --> B --> ... --> A
```
Then wal module in mode A can't distinguish if the shard has been moved.
This may cause a serious bug: 
+ the shard's wal meta information may be modify in other nodes.
+ when it is moved to A(original node), A doesn't know it and think it is same as before it was moved.
+ A persists the old meta information to storage and overwritten the new one which are persisted in other nodes. 

### Steps to reproduce

+ see `test_move_from_nodes` of `wal/src/tests/read_write.rs` in [pr](https://github.com/CeresDB/ceresdb/pull/422).
+ set `region_version_3` equal to `region_version_1`.

### Expected behavior

Can run normally in cluster mode.

### Additional Information

_No response_

Rachelint:

Closes #441 

# Rationale for this change
If a shard (which is mapped to region in wal) is move between nodes like this:
```
A --> B --> ... --> A
```
Then wal module in mode A can't distinguish if the shard has been moved.
This may cause a serious bug: 
+ the shard's wal meta information may be modify in other nodes.
+ when it is moved to A(original node), A doesn't know it and think it is same as before it was moved.
+ A persists the old meta information to storage and overwritten the new one which are persisted in other nodes. 

#### 1. For rocksdb and obkv impls
before:
```                                                
+----------+         +----------+                     
| table id |-------->|table unit|                     
+----------+         +----------+                     
 ```
 now:                           
 ```                                              
+----------+----------+----------+        +----------+
| region   | region   |  table   | ------>|  table   |
| version  |   id     |   id     |        |   unit   |
+----------+----------+----------+        +----------+
 ```                                                
Before,`Table unit` in `wal manager` is only indexed by `table id`, so `wal manager` know nothing about `shard id` and `shard version` (shard are mapped to region). So where `shard id` and `shard version` of `table` changed, `wal manager` can't perceive. 
Now, `Table unit` is indexed by `region version` + `region id` + `table id`, it can perceive any changes.                                             

#### 2. For kafka impl
before:
```                                                
+----------+         +----------+                     
|region id |-------> |  region  |                     
+----------+         +----------+                     
 ```
 now:                           
 ```                                              
+----------+----------+          +----------+
| region   | region   |--------->|  region  |
| version  |   id     |          |          |
+----------+----------+          +----------+
 ``` 
Similar as above, `region` in kafka impl can perceive `shard version` now.

None.

# How does this change test
Test by ut.

### DIFF
--- a/analytic_engine/src/instance/alter.rs
+++ b/analytic_engine/src/instance/alter.rs
@@ -122,19 +122,19 @@ impl Instance {
         let payload = WritePayload::AlterSchema(&alter_schema_pb);
 
         // Encode payloads
-        let region_id = table_data.location();
+        let wal_location = table_data.wal_location();
         let log_batch_encoder =
             self.space_store
                 .wal_manager
-                .encoder(region_id)
+                .encoder(wal_location)
                 .context(GetLogBatchEncoder {
                     table: &table_data.name,
-                    table_location: table_data.location(),
+                    wal_location: table_data.wal_location(),
                 })?;
 
         let log_batch = log_batch_encoder.encode(&payload).context(EncodePayloads {
             table: &table_data.name,
-            table_location: table_data.location(),
+            wal_location: table_data.wal_location(),
         })?;
 
         // Write log batch
@@ -159,7 +159,7 @@ impl Instance {
         let update = MetaUpdate::AlterSchema(manifest_update);
         self.space_store
             .manifest
-            .store_update(MetaUpdateRequest::new(table_data.location(), update))
+            .store_update(MetaUpdateRequest::new(table_data.wal_location(), update))
             .await
             .context(WriteManifest {
                 space_id: table_data.space_id,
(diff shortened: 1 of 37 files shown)


chunshao90:
Please describe the solution in `What changes are included in this PR?`.

chunshao90:
I don't think the current tests are sufficient to cover this bug.

chunshao90:
The amount of code modification is large. Please tell me where is the core modified code.

Rachelint:
> Please describe the solution in `What changes are included in this PR?`.

Done.

Rachelint:
> I don't think the current tests are sufficient to cover this bug.

Done.

Rachelint:
> The amount of code modification is large. Please tell me where is the core modified code.

The codes in `wal/src` is main.

jiacai2050 (approved):
LGTM

PR #423, opened 2022-11-26
feat: replay from flushed sequence

### LINKED ISSUE #282: Replay WAL from the flushed sequence number
### Describe This Problem

Currently, all the existing log entries will be read and replayed when replaying wal, and this is correct now for the `mark_delete_entries_up_to` implementation of RocksDB WAL is sync deletion. However, the method `mark_delete_entries_up_to` is not ensured sync operation, especially for Oceanbase / Kafka implementations.

### Proposal

When replaying wal, we should omit the log entries whose sequence number is smaller than the flushed sequence number persisted in the `Manifest`.

### Additional Context

_No response_

Rachelint:

Closes #282 

# Rationale for this change
 See #282 .

See #282 .

None.

# How does this change test
Test by existing ut.

### DIFF
--- a/analytic_engine/src/instance/open.rs
+++ b/analytic_engine/src/instance/open.rs
@@ -281,7 +281,7 @@ impl Instance {
     ) -> Result<()> {
         let read_req = ReadRequest {
             location: table_data.location(),
-            start: ReadBoundary::Min,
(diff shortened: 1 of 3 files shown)


chunshao90 (approved):
LGTM

PR #431, opened 2022-11-28
fix: fix grpc service metrics

chunshao90:

Closes #

# Rationale for this change
 The grpc service executing time duration metric is incorrectly.

Fix grpc service executing time duration metric.

No.

# How does this change test
No need.


jiacai2050 (approved):
LGTM

PR #434, opened 2022-11-29
fix: load last sequence in obkv wal

Rachelint:

Closes #

# Rationale for this change
 Now, load last sequence step in obkv wal has a logic error, which leads to a serious bug about load a mistaken last sequence.

Fix above bug.

None.

# How does this change test
Test manually.

### DIFF
--- a/wal/src/table_kv_impl/table_unit.rs
+++ b/wal/src/table_kv_impl/table_unit.rs
@@ -395,8 +395,7 @@ impl TableUnit {
         table_id: TableId,
         buckets: &[BucketRef],
     ) -> Result<SequenceNumber> {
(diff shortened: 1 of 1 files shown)


wal/src/table_kv_impl/table_unit.rs
```diff
@@ -407,11 +406,11 @@ impl TableUnit {
                 region_id,
                 table_id,
             )? {
-                last_sequence = seq;
+                return Ok(seq);
```
chunshao90, line 409: Add a comment to describe why return here.

Rachelint: There have been comments originally.

jiacai2050 (approved):
LGTM

PR #437, opened 2022-11-29
refactor: manifest wal config

Rachelint:

Closes #

# Rationale for this change
Background of this pr:
+ now, our `manifest` is `wal` based, and in `wal` on obkv, there is a `ttl` based logs clean strategy.
But data in `manifest` should be permanent and shouldn't be cleaned due to `ttl`. So we should hide the `ttl` of `wal` on obkv, while it is set for `manifest`.

+ the `ttl` and `permanent` strategies are incompatible, so we should check it while opening `wal` (on obkv).

See above background.

Obkv users should switch to new config.

# How does this change test
Test manually.

### DIFF
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -131,12 +131,18 @@ pub struct ObkvWalConfig {
     pub manifest: ManifestNamespaceConfig,
 }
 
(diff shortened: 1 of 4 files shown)


chunshao90 (approved):
LGTM

PR #438, opened 2022-12-01
chore: add logs about replaying.

Rachelint:

Closes #

# Rationale for this change
 When I debug about replaying, I found mostly no logs in the related modules, that made me mad. I add logs to it gradually and after a long long time, I found the bug finally. I think the logs added in my debugging process may be useful. 

+ add logs to `analytic_engine/src/instance/open.rs`.
+ add logs to `analytic_engine/src/meta/details.rs`.

Users won't become mad while debugging related module now.

# How does this change test
Not need.


jiacai2050 (approved):
LGTM

PR #439, opened 2022-12-01
chore: return schema not found when schmea not exists

chunshao90:

Closes #

# Rationale for this change
 When query a not existed schema, the server will return `table not found` response which is confusing.
 

Return schema not found when schmea not exists.

# How does this change test

### DIFF
--- a/sql/src/provider.rs
+++ b/sql/src/provider.rs
@@ -34,6 +34,9 @@ pub enum Error {
         source: catalog::Error,
     },
 
+    #[snafu(display("Failed to find schema, name:{}", name))]
+    SchemaNotFound { name: String },
+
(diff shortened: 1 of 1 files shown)


sql/src/provider.rs
```diff
@@ -111,7 +114,12 @@ impl<'a> MetaProvider for CatalogMetaProvider<'a> {
                 name: resolved.schema,
             })? {
             Some(s) => s,
-            None => return Ok(None),
+            None => {
+                return SchemaNotFound {
+                    name: resolved.schema,
+                }
+                .fail()
```
Rachelint, line 121:
```suggestion
            None => {
                return SchemaNotFound {
                    name: resolved.schema,
                }
                .fail();
```

PR #443, opened 2022-12-01
fix: object_store config

jiacai2050:

Closes #

# Rationale for this change
 
After #421, config of object store is changed, but minimal config is still using old format, and it will cause error below
```
thread 'main' panicked at 'Failed to parse config.: ParseToml { path: "/etc/ceresdb/ceresdb.toml", source: Error { inner: ErrorInner { kind: Custom, l
ine: Some(21), col: 0, at: Some(381), message: "missing field `mem_cache_capacity`", key: ["analytic", "storage"] } }, backtrace: Backtrace(   0: <sna
```

No 
# How does this change test

CI

### DIFF
--- a/.github/workflows/docker-build-image.yml
+++ b/.github/workflows/docker-build-image.yml
@@ -7,6 +7,7 @@ on:
       - '.github/workflows/docker-build-image.yml'
       - 'Dockerfile'
       - 'docker/**'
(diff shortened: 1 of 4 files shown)


ZuLiangWang (approved):
LGTM

Issue #442, opened 2022-12-01
Identifiers should be case-sensitive for all protocols by default


jiacai2050:
### Describe This Problem

In current implementation, we delegate SQL execution to DataFusion, it will automatically convert identifiers to lower case if not quoted, this behavior is what other DBMS does.

But for SQL besides SELECT, identifiers are unchanged no matter whether they are quoted, this cause conflict when 
- CREATE/INSERT table with uppercase letter, such as `Test`
- Then SELECT like `select * from Test` will cause error since no table named `test` exists

### Proposal

<del>Convert identifiers without quoted to lower case for all SQL </del>

After some discussion, proposal above doesn't make much sense in practical. The reason is that there are other protocols support(gRPC) in CeresDB, and they are case-sensitive by default, if we want to keep them consistent with SQL layer, then we need to convert ALL identifiers without quote to lower-case, which is hard to maintain, and may introduce bug when new interface is added without normalized identifiers, so I suggest

> All protocols in CeresDB are case-sensitive by default(with or without quote)

In this way, we only need to modify statements passed to DataFusion, all other interface stay the same with before.

- This may not consistent with other DBMS, but I think it will cause less confusion for CeresDB users.
### Additional Context

- https://github.com/apache/arrow-datafusion/issues/1746
- #220

PR #444, opened 2022-12-02
feat: convert table name to lowercase when not quoted

jiacai2050:

Part of #442

# Rationale for this change
 

Change table name to lowercase when not quoted

Yes, table name will be converted to lowercase when not quoted
# How does this change test

Test harness


Rachelint (approved):
LGTM.

PR #446, opened 2022-12-02
refactor: convert metrics to lowercase in router

Rachelint:

Part of #442 

# Rationale for this change
Unquoted table name should be converted to lowercase.
We should design to solve quoted name later, but it need to refactor sdk, too.

As mentioned in title.

When user uses sdk to interact with ceresdb, the unquoted table name will be converted to lowercase now.

# How does this change test
Test manually.


server/src/grpc/storage_service/route.rs
```diff
@@ -13,11 +18,61 @@ pub async fn handle_route<Q>(
     ctx: &HandlerContext<'_, Q>,
     req: RouteRequest,
 ) -> Result<RouteResponse> {
+    // TODO: the case sensitive mode with quoted is not supported now, all table
```
ShiKaiWi, line 21: Maybe we should file a new issue to track the changes about the case sensitivity. As for this PR, maybe we need a configuration option called `case-insensitive = true` to enable such conversion. 

BTW, in the future changes, I am in favor of that CeresDB may be case-sensitive by default.

PR #447, opened 2022-12-05
feat: add disk cache value crc

jiacai2050:

Closes #

# Rationale for this change
 
Followup PR of #420, add CRC for cached bytes

No

# How does this change test

UT


Rachelint (approved):
LGTM.

PR #449, opened 2022-12-05
feat: impl parallel get_byte_ranges for ObjectStoreReader

chunshao90:

Closes #

# Rationale for this change
 

# How does this change test


ShiKaiWi:
This seems duplicate with #450

PR #451, opened 2022-12-06
refactor: make http max body size configurable and modify default value

Rachelint:

Closes #

# Rationale for this change
Now,  http max body size can't be configured and the default value is too small.

See title.

None.

# How does this change test
Test manually.


jiacai2050 (approved):
LGTM

PR #453, opened 2022-12-06
fix: wrong primary key when define tsid and timestamp key as primary key

ShiKaiWi:

Closes #

# Rationale for this change
 After #340 merged, it can't define (tsid, timestamp_key_name) as primary key. And this pr will fix this problem.

- Refactor the generation of create table plan;
- We can define a primary key with tsid;

The tsid column will be the last column in the schema after this pr, compared with that tsid column just follows the timestamp key column.

# How does this change test
New sql test in the harness.

### DIFF
--- a/sql/src/planner.rs
+++ b/sql/src/planner.rs
@@ -356,31 +356,25 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
             .auto_increment_column_id(true)
             .enable_tsid_primary_key(enable_tsid_primary_key);
 
-        let mut primary_key_column_idxs = Vec::with_capacity(primary_key_columns.len());
-        for column in primary_key_columns {
-            if let Some(idx) = column_idxs_by_name.get(&*column.value) {
-                primary_key_column_idxs.push(*idx);
-            } else {
-                UndefinedColumnInPrimaryKey {
-                    name: &column.value,
-                }
-                .fail()?;
-            }
+        // Collect the key columns.
+        for key_col in primary_key_columns {
+            let col_name = key_col.value.as_str();
+            let col = columns_by_name
+                .remove(col_name)
+                .context(UndefinedColumnInPrimaryKey { name: col_name })?;
+            schema_builder = schema_builder
+                .add_key_column(col)
+                .context(BuildTableSchema)?;
         }
 
-        // The key columns have been consumed.
-        for (idx, col) in columns.iter().enumerate() {
-            let col_name = col.value.as_str();
+        // Collect the normal columns.
+        for normal_col in columns {
+            let col_name = normal_col.value.as_str();
+            // Only normal columns are kept in the `columns_by_name`.
             if let Some(col) = columns_by_name.remove(col_name) {
-                if primary_key_column_idxs.contains(&idx) {
-                    schema_builder = schema_builder
-                        .add_key_column(col)
-                        .context(BuildTableSchema)?;
-                } else {
-                    schema_builder = schema_builder
-                        .add_normal_column(col)
-                        .context(BuildTableSchema)?;
-                }
+                schema_builder = schema_builder
+                    .add_normal_column(col)
+                    .context(BuildTableSchema)?;
             }
         }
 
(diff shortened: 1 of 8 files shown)


sql/src/planner.rs
```diff
@@ -322,29 +343,133 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
...
-            schema::Builder::with_capacity(stmt.columns.len()).auto_increment_column_id(true);
+        debug!("Create table to plan, stmt:{:?}", stmt);
 
         // Build all column schemas.
-        let mut name_column_map = stmt
+        let mut columns_by_name = stmt
             .columns
             .iter()
             .map(|col| Ok((col.name.value.as_str(), parse_column(col)?)))
             .collect::<Result<BTreeMap<_, _>>>()?;
```
jiacai2050, line 470: Both `columns_by_name` and `column_idxs_by_name` require no ordering, HashMap is enough.

sql/src/planner.rs
```diff
@@ -353,112 +478,67 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
...
+            Self::find_and_ensure_timestamp_column(&columns_by_name, &stmt.constraints)?;
+        let tsid_column = Ident::with_quote(DEFAULT_QUOTE_CHAR, TSID_COLUMN);
+        let mut columns: Vec<_> = stmt.columns.iter().map(|col| col.name.clone()).collect();
+        let mut enable_tsid_primary_key = false;
+
+        let mut add_tsid_column = || {
+            columns_by_name.insert(TSID_COLUMN, Self::tsid_column_schema()?);
+            column_idxs_by_name.insert(TSID_COLUMN, columns.len());
+            columns.push(tsid_column.clone());
+            enable_tsid_primary_key = true;
```
jiacai2050, line 497: I dont remember why this field is required, maybe can be removed?

tests/cases/local/05_ddl/create_tables.sql
```diff
@@ -69,6 +69,16 @@ CREATE TABLE `05_create_tables_t9`(c1 int, c2 bigint default 0, c3 uint32 defaul
 show create table `05_create_tables_t9`;
 drop table `05_create_tables_t9`;
 
+-- Explicit primary key with tsid
+CREATE TABLE `05_create_tables_t10`(c1 int, t1 timestamp NOT NULL TIMESTAMP KEY, PRIMARY KEY(tsid, t1)) ENGINE = Analytic;
+show create table `05_create_tables_t10`;
+drop table `05_create_tables_t10`;
+
+-- Explicit primary key with tsid
+CREATE TABLE `05_create_tables_t11`(c1 int, t1 timestamp NOT NULL TIMESTAMP KEY, PRIMARY KEY(t1, tsid)) ENGINE = Analytic;
```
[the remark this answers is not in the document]
ShiKaiWi: The cases above in the file cover that.

sql/src/planner.rs
```diff
@@ -353,112 +478,67 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
...
-            } = constraint
-            {
-                if *is_primary {
-                    // Build primary key, the builder will check timestamp column is in primary key.
-                    for column in columns {
-                        if let Some(idx) = name_column_index_map.get(&*column.value) {
-                            primary_key_column_idxs.push(*idx);
+        let timestamp_column =
+            Self::find_and_ensure_timestamp_column(&columns_by_name, &stmt.constraints)?;
+        let tsid_column = Ident::with_quote(DEFAULT_QUOTE_CHAR, TSID_COLUMN);
```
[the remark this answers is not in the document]
ShiKaiWi: Make all the idents are case-sensitive.

sql/src/planner.rs
ShiKaiWi: Let's remove it in the next PR because this needs updating of ceresdbproto.

jiacai2050 (approved):
LGTM

PR #454, opened 2022-12-06
feat: convert table name to quoted style

Rachelint:

Closes #

# Rationale for this change
For #442 .
Now we convert table name to quoted style to prevent `datafusion` from converting it to lowercase. 

All changes are in `sql/src/parser.rs`.
Main steps:
+ extract the table name from `select ast`.
+ convert it to quoted style.
+ rebuild the `select ast`.

Table name in ceresdb will be case-sensitive.

# How does this change test
Test by ut and integrated test.

### DIFF
--- a/sql/src/parser.rs
+++ b/sql/src/parser.rs
@@ -180,7 +180,6 @@ impl<'a> Parser<'a> {
                         self.parser.next_token();
                         self.parse_exists()
                     }
-                    // Keyword::Q
                     _ => {
                         // use the native parser
                         Ok(Statement::Standard(Box::new(maybe_normalize_table_name(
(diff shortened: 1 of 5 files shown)


sql/src/parser.rs
```diff
@@ -540,6 +543,130 @@ fn build_timestamp_key_constraint(col_defs: &[ColumnDef], constraints: &mut Vec<
...
+/// Add quotes in table name (for example: convert table to `table`).
+///
+/// It is used to process table name in `SELECT`, for preventing `datafusion`
+/// converting the table name to lowercase, because `CeresDB` only support
+/// case-sensitive in sql.
+// TODO: maybe other items(such as: alias, column name) need to be normalized,
+// too.
+pub fn maybe_normalize_table_name(statement: SqlStatement) -> SqlStatement {
+    let original_statement = statement.clone();
+    if let SqlStatement::Query(query) = statement {
```
ShiKaiWi, line 555: Maybe we can modify the statement directly to avoid verbose struct unwrapping.

Rachelint: > Maybe we can modify the statement directly to avoid verbose struct unwrapping.

Done.

ShiKaiWi (approved):
LGTM

PR #459, opened 2022-12-07
refactor: remove `enable_tsid_primary_key`

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, the field in the schema `enable_tsid_primary_key` is useless. Let's remove it.

Remove `enable_tsid_primary_key` field in the schema.

This will lead to breaking change for that the field `enable_tsid_primary_key` is part of the format of stored schema.

# How does this change test
Existing tests.


jiacai2050 (approved):
LGTM

PR #460, opened 2022-12-07
feat: make rocksdb wal compatible with other distributed implementation

Rachelint:

Closes #

# Rationale for this change
Now, If you write log with a not default region id in key, you will be unable to recover it when you restart the database(use a rocksdb based wal). It make no sense, so we should adjust the recover process in rocksdb based wal.

Adjust the recover process in rocksdb based wal.

None.

# How does this change test
Test by ut.

### DIFF
--- a/wal/src/manager.rs
+++ b/wal/src/manager.rs
@@ -119,6 +119,9 @@ pub mod error {
 
         #[snafu(display("Failed to execute in runtime, err:{}", source))]
         RuntimeExec { source: common_util::runtime::Error },
+
+        #[snafu(display("Encountered Unknown error, msg:{}.\nBacktrace:\n{}", msg, backtrace))]
+        Unknown { msg: String, backtrace: Backtrace },
     }
 
     define_result!(Error);
--- a/wal/src/manager.rs
+++ b/wal/src/manager.rs
@@ -120,7 +120,7 @@ pub mod error {
         #[snafu(display("Failed to execute in runtime, err:{}", source))]
         RuntimeExec { source: common_util::runtime::Error },
 
-        #[snafu(display("Encountered Unknown error, msg:{}.\nBacktrace:\n{}", msg, backtrace))]
+        #[snafu(display("Encountered unknown error, msg:{}.\nBacktrace:\n{}", msg, backtrace))]
         Unknown { msg: String, backtrace: Backtrace },
     }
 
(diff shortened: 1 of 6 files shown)


wal/src/manager.rs
```diff
@@ -119,6 +119,9 @@ pub mod error {
 
         #[snafu(display("Failed to execute in runtime, err:{}", source))]
         RuntimeExec { source: common_util::runtime::Error },
+
+        #[snafu(display("Encountered Unknown error, msg:{}.\nBacktrace:\n{}", msg, backtrace))]
```
ShiKaiWi:
```suggestion
        #[snafu(display("Encounter unknown error, msg:{}.\nBacktrace:\n{}", msg, backtrace))]
```

wal/src/rocks_impl/manager.rs
```diff
@@ -270,40 +270,114 @@ impl RocksImpl {
...
         }
 
         Ok(())
     }
 
     fn find_table_seqs_from_table_log(
+        &self,
+        table_max_seqs: &mut HashMap<TableId, SequenceNumber>,
+    ) -> Result<()> {
+        self.find_table_seqs_from_table_log_by_region_id(table_max_seqs)
```
ShiKaiWi, line 283: How about avoiding defining the method `find_table_seqs_from_table_log_by_region_id` because it looks redundant.

wal/src/rocks_impl/manager.rs
```diff
@@ -270,40 +270,114 @@ impl RocksImpl {
...
+                break;
+            }
+
+            let log_key = self
+                .log_encoding
+                .decode_key(iter.key())
+                .map_err(|e| Box::new(e) as _)
+                .context(Decoding)?;
+
+            // Found a valid log key by `current_region_id`, search table sequences by table
```
ShiKaiWi, line 336: Actually, the log key may not belong to `current_region_id`, let's put the comment as:
```suggestion
            // The max valid log key in a region is found, search table sequences by table then.
```

wal/src/rocks_impl/manager.rs
```diff
@@ -315,22 +389,38 @@ impl RocksImpl {
...
+            debug!(
+                "RocksImpl has found log key by table id, log key:{:?}",
+                log_key
+            );
+
+            if log_key.region_id != region_id {
+                debug!("RocksImpl find table unit pairs stop scanning by table id, because has encountered next region id,
+                    searching table id:{}
+                    current region id:{},
+                    next region id:{}", current_table_id, region_id, log_key.region_id);
```
ShiKaiWi, line 413: The format for these ids seems weird, how about letting the `cargo fmt` take over the formatting work instead of the manual formatting.

ShiKaiWi (approved):
LGTM

PR #461, opened 2022-12-07
fix: kill ceresdb-server if run harness failed

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, the `ceresdb-server` won't exit if the harness test fails.

Kill the `ceresdb-server` whether the harness test succeeds or not.

No confusion when running harness test locally.

# How does this change test
Test by manually.


Rachelint (approved):
LGTM

PR #462, opened 2022-12-07
chore: kill test server in drop

jiacai2050:

Closes #

# Rationale for this change
 
When a test cause panic, the test server will not exit.

No

# How does this change test

Test harness and this https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ead33da929c60c86d4d5ce9730028bd2


ShiKaiWi:
#461 has fixed this. 😅

PR #465, opened 2022-12-08
fix: wrong path in the result from StoreWithPrefix

ShiKaiWi:

Closes #

# Rationale for this change
 The real path from the `StoreWithPrefix` contains the extra `prefix` which should be stripped.

- Remove the extra `prefix` from the result of `StoreWithPrefix`;

None.

# How does this change test
New tests.

### DIFF
--- a/components/object_store/src/prefix.rs
+++ b/components/object_store/src/prefix.rs
@@ -193,10 +193,172 @@ impl ObjectStore for StoreWithPrefix {
 mod tests {
     use std::sync::Arc;
 
+    use chrono::{DateTime, Utc};
+    use futures::stream;
     use tempfile::tempdir;
     use upstream::local::LocalFileSystem;
 
     use super::*;
+
+    #[derive(Debug, Clone)]
+    struct PathPrefixChecker {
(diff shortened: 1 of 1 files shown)


components/object_store/src/prefix.rs
```diff
@@ -130,11 +193,171 @@ impl ObjectStore for StoreWithPrefix {
...
+        let mock_store = Arc::new(mock_store);
+        let prefix_store = StoreWithPrefix::new(test_prefix.to_string(), mock_store).unwrap();
+
+        let test_filepath = Path::from("0/100");
+
+        let _ = prefix_store
+            .put(&test_filepath, Bytes::from_static(b"1111"))
+            .await;
+
+        let _ = prefix_store.get(&test_filepath).await;
```
jiacai2050, line 335: If you ignore those result, how can you ensure expected behavior?

ShiKaiWi: I have added some comments to explain this.

jiacai2050 (approved):
LGTM

PR #474, opened 2022-12-12
feat: scan row groups in one sst file parallelly

Rachelint:

Closes #

# Rationale for this change
 Now, the concurrent level while pulling sst data is on file level, we can enhance it to row group level.

# How does this change test
Test by ut.

### DIFF
--- a/analytic_engine/Cargo.toml
+++ b/analytic_engine/Cargo.toml
@@ -51,5 +51,5 @@ wal = { workspace = true }
 common_types = { workspace = true, features = ["test"] }
 common_util = { workspace = true, features = ["test"] }
 env_logger = { workspace = true }
+rand = { workspace = true }
 wal = { workspace = true, features = ["test"] }
-rand = "0.8.5"
--- a/analytic_engine/Cargo.toml
+++ b/analytic_engine/Cargo.toml
@@ -52,3 +52,4 @@ common_types = { workspace = true, features = ["test"] }
 common_util = { workspace = true, features = ["test"] }
 env_logger = { workspace = true }
 wal = { workspace = true, features = ["test"] }
+rand = "0.8.5"
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -111,7 +111,7 @@ impl Default for Config {
             /// it.
             db_write_buffer_size: 0,
             scan_batch_size: 500,
-            sst_background_read_parallelism: 1,
+            sst_background_read_parallelism: 4,
             wal_storage: WalStorageConfig::RocksDB,
         }
     }
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -111,7 +111,7 @@ impl Default for Config {
             /// it.
             db_write_buffer_size: 0,
             scan_batch_size: 500,
-            sst_background_read_parallelism: 4,
+            sst_background_read_parallelism: 8,
             wal_storage: WalStorageConfig::RocksDB,
         }
     }
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -73,10 +73,13 @@ pub struct Config {
     pub space_write_buffer_size: usize,
     /// The maximum size of all Write Buffers across all spaces.
     pub db_write_buffer_size: usize,
-    // End of global write buffer options.
+    /// End of global write buffer options.
 
-    // Batch size for scan sst
+    // Iterator scanning options
+    /// Batch size for iterator
     pub scan_batch_size: usize,
+    /// Sst background reading parallelism
+    pub sst_background_read_parallelism: usize,
 
     /// Wal storage config
     ///
@@ -108,6 +111,7 @@ impl Default for Config {
             /// it.
             db_write_buffer_size: 0,
             scan_batch_size: 500,
+            sst_background_read_parallelism: 1,
             wal_storage: WalStorageConfig::RocksDB,
         }
     }
--- a/analytic_engine/src/sst/parquet/async_reader.rs
+++ b/analytic_engine/src/sst/parquet/async_reader.rs
@@ -20,7 +20,7 @@ use common_types::{
 use common_util::{runtime::Runtime, time::InstantExt};
 use datafusion::datasource::file_format;
 use futures::{future::BoxFuture, FutureExt, Stream, StreamExt, TryFutureExt};
-use log::{error, info};
+use log::{debug, error, info, warn};
 use object_store::{ObjectMeta, ObjectStoreRef, Path};
 use parquet::{
     arrow::{async_reader::AsyncFileReader, ParquetRecordBatchStreamBuilder, ProjectionMask},
@@ -59,11 +59,17 @@ pub struct Reader<'a> {
     /// Init those fields in `init_if_necessary`
     meta_data: Option<MetaData>,
     row_projector: Option<RowProjector>,
+
+    /// Options for `read_parallelly`
+    parallelism_options: ParallelismOptions,
 }
 
 impl<'a> Reader<'a> {
     pub fn new(path: &'a Path, storage: &'a ObjectStoreRef, options: &SstReaderOptions) -> Self {
         let batch_size = options.read_batch_row_num;
+        let parallelism_options =
+            ParallelismOptions::new(options.read_batch_row_num, options.num_rows_per_row_group);
+
         Self {
             path,
             storage,
@@ -73,6 +79,7 @@ impl<'a> Reader<'a> {
             batch_size,
             meta_data: None,
             row_projector: None,
+            parallelism_options,
         }
     }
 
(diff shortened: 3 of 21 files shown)


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -371,39 +434,36 @@ impl<'a> SstReader for Reader<'a> {
...
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        self.as_mut().rx.poll_recv(cx)
+        let cur_rx_idx = self.cur_rx_idx;
+        // `cur_rx_idx` is impossible to be out-of-range, because it is got by round
+        // robin.
+        let cur_rx = self.rx_group.get_mut(cur_rx_idx).unwrap();
+        let poll_result = cur_rx.poll_recv(cx);
+
+        match poll_result {
```
jiacai2050, line 460: Add comments why only advance idx in Ready

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -105,29 +142,54 @@ impl<'a> Reader<'a> {
...
-            "fetch_record_batch row_groups total:{}, after filter:{}",
+        info!(
+            "fetch_record_batch row_groups total:{}, after filter:{}, row_group_num_per_reader:{}",
             meta_data.parquet().num_row_groups(),
-            filtered_row_groups.len()
+            filtered_row_groups.len(),
+            read_parallelism,
         );
 
+        // TODO: now `batch_size` is equal to `num_rows_per_row_group`, other situations
```
jiacai2050, line 152: I  can't understand this TODO, what issue are you trying to solve?

jiacai2050 (changes requested):

analytic_engine/src/sst/parquet/async_reader.rs
Rachelint: Done.

analytic_engine/src/sst/parquet/async_reader.rs
Rachelint: Has written an new TODO.

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -371,39 +523,56 @@ impl<'a> SstReader for Reader<'a> {
...
+                "cur_rx_idx is impossible to be out-of-range, cur_rx_idx:{}, rx_group len:{}",
+                cur_rx_idx, rx_group_len
+            )
+        });
+        let poll_result = cur_rx.poll_recv(cx);
+
+        match poll_result {
+            Poll::Ready(result) => {
+                // If found `Poll::Pending`, we need to keep polling current rx
+                // util found `Poll::Ready` for ensuring the order of record batches,
```
jiacai2050, line 562: until

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -198,12 +297,65 @@ impl<'a> Reader<'a> {
     }
 }
 
-#[derive(Debug)]
+/// Options for `read_parallely` in [Reader]
+#[derive(Debug, Clone, Copy)]
+struct ParallelismOptions {
```
jiacai2050, line 302: How about this
```
enum ParallelismOptions {
  Disable,
  Enable(max_row_groups_in_batch)
}
```

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -453,9 +630,200 @@ impl<'a> SstReader for ThreadedReader<'a> {
...
+        // Get underlying sst readers and channels.
+        let sub_readers = self.inner.read_parallelly(self.read_parallelism).await?;
+        if sub_readers.is_empty() {
+            return Ok(Box::new(RecordBatchReceiver {
+                rx_group: Vec::new(),
+                cur_rx_idx: 0,
+            }) as _);
+        }
+
+        let read_parallelism = if self.read_parallelism >= sub_readers.len() {
```
jiacai2050, line 642: Isn't the same with 

```
let read_parallelism =  sub_readers.len()
```

analytic_engine/src/sst/parquet/async_reader.rs
Rachelint: I have removed `max_row_groups_in_batch` and revert the logic.
Because if we set the batch_size to integral multiple `num_rows_per_row_group`, a serious error will happen in the new logic due to the implementation of parquet reader.
So, we shouldn't support it now and should make it a todo.

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -92,42 +148,85 @@ impl<'a> Reader<'a> {
...
+            filtered_row_group_len,
+            max_row_groups_in_batch,
+            row_group_batches.len(),
+            read_parallelism,
         );
 
+        // Partition the batches by `read_parallelism`.
+        let read_parallelism = std::cmp::min(row_group_batches.len(), read_parallelism);
+        let chunks_num = read_parallelism;
+        let chunk_size = filtered_row_group_len / read_parallelism + 1;
```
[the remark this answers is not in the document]
Rachelint: Impossible, it has an assert in `read_parallelly`.

analytic_engine/Cargo.toml
```diff
@@ -52,3 +52,4 @@ common_types = { workspace = true, features = ["test"] }
 common_util = { workspace = true, features = ["test"] }
 env_logger = { workspace = true }
 wal = { workspace = true, features = ["test"] }
+rand = "0.8.5"
```
chunshao90, line 55: Useless.

chunshao90: rand = { workspace = true }

Rachelint: Done.

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -73,9 +79,54 @@ impl<'a> Reader<'a> {
             batch_size,
             meta_data: None,
             row_projector: None,
+            parallelism_options,
         }
     }
 
+    async fn read_parallelly(
```
jiacai2050, line 86: How about name this:
```
try_parallel_read
```

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -198,12 +288,47 @@ impl<'a> Reader<'a> {
...
 }
 
-#[derive(Debug)]
+/// Options for `read_parallely` in [Reader]
+#[derive(Debug, Clone, Copy)]
+struct ParallelismOptions {
+    /// Whether allow parallelly reading.
+    ///
+    /// NOTICE: now we only allow `read_parallelly` when
+    /// `read_batch_row_num` % `num_rows_per_row_group` == 0
```
jiacai2050, line 297: This comments is not consistent with code below
```
if read_batch_row_num != num_rows_per_row_group
```

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -419,28 +561,36 @@ pub struct ThreadedReader<'a> {
...
         Self {
             inner: reader,
             runtime,
             channel_cap: DEFAULT_CHANNEL_CAP,
+            read_parallelism,
         }
     }
 
-    async fn read_record_batches(&mut self, tx: Sender<Result<RecordBatchWithKey>>) -> Result<()> {
-        let mut stream = self.inner.read().await?;
```
jiacai2050, line 434: Better to add some comments to inner read is not used now. such as

```
This method is not used in favor of `try_read_parallel`
```

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -56,14 +56,20 @@ pub struct Reader<'a> {
     predicate: PredicateRef,
     batch_size: usize,
 
-    /// init those fields in `init_if_necessary`
+    /// Init those fields in `init_if_necessary`
     meta_data: Option<MetaData>,
     row_projector: Option<RowProjector>,
+
+    /// Options for `read_parallelly`
+    parallelism_options: ParallelismOptions,
```
jiacai2050, line 64: This fields is unnecessary  for `Reader`, it's only used in `read_parallel`, we can just a argument there.

analytic_engine/src/sst/parquet/async_reader.rs
chunshao90: Why modify `read_parallelly` to `try_parallel_read`? The return value is `Result` cannot decide to use the `try` prefix.

analytic_engine/src/sst/parquet/async_reader.rs
Rachelint: Maybe it is suitable to place it here?
+ `Reader` has the parallel read ability(by `read_parallelly`).
+ caller can pass `read_parallelism` as a suggest parallelism.
+ `Reader` check the suggest parallelism according to its `ParallelismOptions` generated in `new`.

But the name `ParallelismOptions` may be not a good name to express this?

analytic_engine/src/sst/parquet/async_reader.rs
Rachelint: How about `suggest_to_read_parallelly`? 
Try prefix seems to emphasize the `Result` returned value?

analytic_engine/src/lib.rs
```diff
@@ -108,6 +111,7 @@ impl Default for Config {
             /// it.
             db_write_buffer_size: 0,
             scan_batch_size: 500,
+            sst_background_read_parallelism: 4,
```
jiacai2050, line 114: default to 8

jiacai2050 (approved):
LGTM

PR #475, opened 2022-12-12
chore: bump oss sdk

ZuLiangWang:

Closes #

# Rationale for this change
Upgrade oss client to the latest version.

* Update oss client and adapt to the new configuration.
* Add update log level return value.

None.

# How does this change test

### DIFF
--- a/analytic_engine/src/storage_options.rs
+++ b/analytic_engine/src/storage_options.rs
@@ -1,5 +1,7 @@
 // Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
 
+use std::time::Duration;
+
 use common_util::config::{ReadableDuration, ReadableSize};
 use serde::Deserialize;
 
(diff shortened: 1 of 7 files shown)


server/src/http.rs
```diff
@@ -252,7 +252,7 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
                         .set_level_by_str(log_level.as_str())
                         .map_err(|e| Error::HandleUpdateLogLevel { msg: e });
                     match result {
-                        Ok(()) => Ok(reply::reply()),
+                        Ok(()) => Ok(reply::json(&"ok")),
```
jiacai2050, line 255: Return `log_level`

analytic_engine/src/storage_options.rs
```diff
@@ -54,4 +54,6 @@ pub struct AliyunOptions {
     pub endpoint: String,
     pub bucket: String,
     pub prefix: String,
+    pub pool_max_idle_per_host: usize,
```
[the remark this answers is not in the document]
jiacai2050: This default to 0 now, which is not what we want.
```
    #[serde(default = 1024)]
```
- https://serde.rs/attr-default.html

jiacai2050 (changes requested):

analytic_engine/src/storage_options.rs
```diff
@@ -54,4 +54,8 @@ pub struct AliyunOptions {
     pub endpoint: String,
     pub bucket: String,
     pub prefix: String,
+    #[serde(default)]
+    pub pool_max_idle_per_host: usize,
+    #[serde(default)]
+    pub timeout: ReadableDuration,
```
jiacai2050, line 60: Same with before.

jiacai2050 (approved):
LGTM

PR #476, opened 2022-12-12
feat: support memory usage limit on background compaction

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, there is no limit on the memory usage by the background compaction task, leading to frequent OOM on the prod environment.

- Add memory usage limit for compaction scheduler, so the task will be skipped and put back to the request queue if the memory threshold is reached.

None.

# How does this change test
New unit tests.


Rachelint (approved):
LGTM

Issue #468, opened 2022-12-09
More metrics for observability of important modules


ShiKaiWi:
### Describe This Problem

Currently, the existing metrics are not enough for troubleshooting performance issues. We need more metrics for better observability on important modules.

### Proposal

Currently, these modules need metrics:
- [x] Object store: query range distribution, read/write latency, read/write throughput, counters for rpc call. by #478 
- [x] Write: time cost for critical path
- [ ] Query: time cost for critical path

PR #478, opened 2022-12-12
chore: add metrics oss

chunshao90:

Closes #

# Rationale for this change
 Part of  #468 

* Use `static_metric` in prometheus to reduce the specific overhead of retrieval.
* Impl StoreWithMetrics.

No.

# How does this change test


components/object_store/src/metrics.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#[async_trait]
+impl ObjectStore for StoreWithMetrics {
+    async fn put(&self, location: &Path, bytes: Bytes) -> Result<()> {
+        let _timer = OBJECT_STORE_DURATION_HISTOGRAM.put.start_timer();
+        OBJECT_STORE_THROUGHPUT_HISTOGRAM
+            .put
+            .observe(bytes.len() as f64);
+        self.store.put(location, bytes).await
```
ShiKaiWi, line 95: How about adding a newline between the metrics and the call on the real store.

ShiKaiWi: Maybe we can fix this in another PR. After all, it is just a trivial.

ShiKaiWi (approved):

PR #479, opened 2022-12-12
feat: support bloom filter in hybrid format

Rachelint:

Closes #

# Rationale for this change
Now, parquet builder has support manual flushing api, we can try to support bloom filter in hybrid format. 

Manual flushing after writing a hybrid row group.

None.

# How does this change test
Test by ut.


chunshao90 (approved):
LGTM

PR #482, opened 2022-12-14
fix: compact table in background scheduler

jiacai2050:

Closes #

# Rationale for this change
 
In current implementation, when a table has no write any more, its data will never be compacted.

- Update scheduler, do table compact at fixed interval.

No

# How does this change test

Manullay.

### DIFF
--- a/analytic_engine/src/compaction/scheduler.rs
+++ b/analytic_engine/src/compaction/scheduler.rs
@@ -614,13 +614,14 @@ impl ScheduleWorker {
         self.space_store.list_all_tables(&mut tables_buf);
 
         let request_id = RequestId::next_id();
-        // Spawn a background job to purge ssts and avoid schedule thread blocked.
         for table_data in tables_buf {
             info!(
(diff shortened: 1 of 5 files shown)


analytic_engine/src/compaction/scheduler.rs
```diff
@@ -600,67 +605,28 @@ impl ScheduleWorker {
     }
 
     async fn schedule(&mut self) {
-        self.purge_tables();
+        self.purge_tables().await;
         self.flush_tables().await;
     }
 
-    fn purge_tables(&mut self) {
+    async fn purge_tables(&mut self) {
```
ShiKaiWi, line 612: It is weird that the purge procedure seems removed? I guess we should add a procedure for periodical compaction rather than replacing the purge procedure with the new compaction procedure.

jiacai2050: The purge logic is already in `compact_table`, I just reuse  `handle_table_compaction_request` function to do the compaction.

ShiKaiWi (approved):
LGTM

PR #483, opened 2022-12-15
feat: add max_input_sstable_size

### LINKED ISSUE #408: Support limit the size of sst generated by compaction
### Describe This Problem

Currently, it seems there is no limit on the sst size generated by the compaction procedure, that is to say, some huge sst files may occur. It is ok for the local file system, but it will cause some problems, including higher possibility of failure of downloading or uploading when using the remote object store as the underlying storage layer.

### Proposal

Firstly, we should make the sst size limit configurable. However, this is not enough to handle the case where the size of the compaction output is indeed larger than the limit. So I propose the compaction procedure should support multiple output files. After that, the very large output sst will be split to multiple small ssts.

### Additional Context

_No response_

jiacai2050:

Part of #408

# Rationale for this change
 
The target of #408 plan to limit output sst file, but there is no easy way to tell how large the output file will be, so
this PR use another strategy: limit total input size.

When pick files for compaction, add a new limit: `max_input_sstable_size`,
it will control total input file size to compact.

For now, we can custom it via environment variables, such as
```
export CERESDB_MAX_INPUT_SSTABLE_SIZE=1G
```

No
# How does this change test

New unit test `test_size_tiered_picker`

### DIFF
--- a/analytic_engine/src/compaction/mod.rs
+++ b/analytic_engine/src/compaction/mod.rs
@@ -115,7 +115,6 @@ const MIN_THRESHOLD_KEY: &str = "compaction_min_threshold";
 const MAX_THRESHOLD_KEY: &str = "compaction_max_threshold";
 const MIN_SSTABLE_SIZE_KEY: &str = "compaction_min_sstable_size";
 const TIMESTAMP_RESOLUTION_KEY: &str = "compaction_timestamp_resolution";
-const MAX_INPUT_SSTABLE_SIZE_KEY: &str = "compaction_max_input_sstable_size";
 const DEFAULT_STRATEGY: &str = "default";
 const STC_STRATEGY: &str = "size_tiered";
 const TWC_STRATEGY: &str = "time_window";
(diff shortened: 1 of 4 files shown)


analytic_engine/src/compaction/mod.rs
```diff
@@ -113,6 +115,7 @@ const MIN_THRESHOLD_KEY: &str = "compaction_min_threshold";
 const MAX_THRESHOLD_KEY: &str = "compaction_max_threshold";
 const MIN_SSTABLE_SIZE_KEY: &str = "compaction_min_sstable_size";
 const TIMESTAMP_RESOLUTION_KEY: &str = "compaction_timestamp_resolution";
+const MAX_INPUT_SSTABLE_SIZE_KEY: &str = "compaction_max_input_sstable_size";
```
ShiKaiWi, line 118: I guess `compaction_max_input_sstable_size` is just a temporary method to limit the sst file size, so should we put it as one of the table options (which may be persisted as table meta data)?

jiacai2050: Removed

ShiKaiWi (approved):
LGTM

PR #485, opened 2022-12-15
feat: support create partition table

chunshao90:

Closes #

# Rationale for this change
 Job part of #492 

 * Store partition info in manifest.
 * Transfer partition info to ceresmeta when `createTable`.
 * `show create table` display partition info.

No.

# How does this change test
* Old tests cover and manual test.
* Add `test_render_partition_info` test.


table_engine/src/partition/mod.rs
```diff
@@ -2,26 +2,84 @@
...
+    pub name: String,
+    pub origin_name: Option<String>,
 }
 
-/// Size type of partition num
-pub type PartitionNum = u16;
+#[derive(Clone, Debug, PartialEq)]
+pub struct HashPartitionInfo {
+    pub definitions: Vec<Definition>,
+    pub columns: Vec<String>,
```
Rachelint, line 32: I think store `Expr` is enough.
We can call `accept` to get column name in `Expr`.

table_engine/src/partition/rule.rs
```diff
@@ -18,91 +17,55 @@ pub enum Error {
...
-        let mut column_index = SmallVec::with_capacity(col_name_list.size_hint().0);
-        for col_name in col_name_list {
-            for (i, v) in schema.columns().iter().enumerate() {
-                if col_name == v.name {
-                    column_index.push(i);
-                    break;
-                }
-            }
-        }
+impl PartitionRule for HashPartitionRule {
```
Rachelint, line 51: `PartitionRule` should be place in a new pr, and the methods in it should be considered more.

proto/protos/meta_update.proto
```diff
@@ -13,6 +13,18 @@ message AddSpaceMeta {
...
+message Definition {
+  string name = 1;
+  oneof origin_name{
+    string origin = 2;
+  }
+}
+
+message HashPartitionInfo {
+  repeated Definition definitions = 1;
+  repeated string columns = 2;
```
Rachelint, line 25: Modify 
```suggestion
 bytes expr = 2;
 bool linear = 3;
```

catalog/src/schema.rs
```diff
@@ -219,6 +220,8 @@ pub struct CreateTableRequest {
     pub shard_id: ShardId,
     /// Cluster version of shard
     pub cluster_version: ClusterVersion,
+    /// Partition info if this is a partitioned table
```
ZuLiangWang, line 223:
```suggestion
    /// Partition info if this is a partitioned table.
```

catalog/src/schema.rs
```diff
@@ -219,6 +220,8 @@ pub struct CreateTableRequest {
     pub shard_id: ShardId,
     /// Cluster version of shard
```
ZuLiangWang, line 221:
```suggestion
    /// Cluster version of shard.
```

ZuLiangWang (changes requested):

PR #487, opened 2022-12-17
feat: support parsing partition table creating

Rachelint:

Closes #

# Rationale for this change
We need to support table partition now, this pr is for supporting parsing the related sql to statement.
The syntax is same as [Mysql](https://dev.mysql.com/doc/refman/8.0/en/partitioning-types.html) .
But now, we only support the hash strategy. 

Add sql syntax support for creating partition table(only hash strategy).
Mainly modify `sql/src/parser.rs`.

Ceresdb can accept related sql syntax now(but related function is not supported yet).

# How does this change test
Test by ut.

### DIFF
--- a/sql/src/ast.rs
+++ b/sql/src/ast.rs
@@ -13,7 +13,7 @@ pub enum Statement {
     Standard(Box<SqlStatement>),
     // Other extensions
     /// CREATE TABLE
-    Create(CreateTable),
+    Create(Box<CreateTable>),
     /// Drop TABLE
     Drop(DropTable),
     Describe(DescribeTable),
@@ -69,6 +69,22 @@ pub struct CreateTable {
     pub constraints: Vec<TableConstraint>,
     /// Table options in `WITH`.
     pub options: Vec<SqlOption>,
+    pub partition: Option<Partition>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Partition {
+    Hash(HashPartition),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct HashPartition {
+    /// Decide to use which hash algorithm
+    ///
+    /// You can see: https://dev.mysql.com/doc/refman/8.0/en/partitioning-hash.html
+    pub linear: bool,
+    pub partition_num: u64,
+    pub expr: sqlparser::ast::Expr,
 }
 
 #[derive(Debug, PartialEq, Eq)]
--- a/sql/src/parser.rs
+++ b/sql/src/parser.rs
@@ -8,7 +8,7 @@ use log::debug;
 use paste::paste;
 use sqlparser::{
     ast::{
-        ColumnDef, ColumnOption, ColumnOptionDef, Expr, Ident, ObjectName, SetExpr,
+        ColumnDef, ColumnOption, ColumnOptionDef, DataType, Expr, Ident, ObjectName, SetExpr,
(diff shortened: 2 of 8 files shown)


sql/src/parser.rs
```diff
@@ -818,6 +914,94 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_create_partition_table_with_hash() {
```
chunshao90, line 918: This test is too large, spilt in some same tests?

sql/src/ast.rs
```diff
@@ -69,6 +69,19 @@ pub struct CreateTable {
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Partition {
+    Hash(HashPartition),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct HashPartition {
+    pub linear: bool,
```
ZuLiangWang, line 82: Add some comments  to explain the use of `linear`.

ZuLiangWang (changes requested):

sql/src/ast.rs
Rachelint: Done.

sql/src/parser.rs
Rachelint: Done.

sql/src/parser.rs
```diff
@@ -510,6 +518,82 @@ impl<'a> Parser<'a> {
...
+        }
+
+        // TODO: support all valid exprs not only column expr.
+        let expr = self.parse_valid_expr_in_hash()?;
+
+        // Parse second part: "PARTITIONS num" (if not set, num will use 1 as default).
+        let partition_num = if self.parser.parse_keyword(Keyword::PARTITIONS) {
+            match self.parser.parse_number_value()? {
+                sqlparser::ast::Value::Number(v, _) => v.parse::<u64>().map_err(|e| {
+                    ParserError::ParserError(format!(
```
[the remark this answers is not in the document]
Rachelint: Done.

ZuLiangWang (approved):
LGTM

sql/src/parser.rs
```diff
@@ -518,6 +622,51 @@ impl<'a> Parser<'a> {
...
+}
+
+// Valid column expr in hash should meet following conditions:
+// 1. column must be a tag, tsid + timestamp can be seen as the combined unique
+// key, and partition key must be the subset of it(for supporting overwritten
+// insert mode). About the reason, maybe you can see: https://docs.pingcap.com/zh/tidb/stable/partitioned-table
+//
+// 2. column's datatype must be integer(int, tiny int, small int, big int ...).
+//
+// TODO: we should consider the situation: tag column not exist.
```
chunshao90:
```suggestion
// TODO: we should consider the situation: no tag column is set.
```

sql/src/partition.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub fn parse_hash(hash_stmt: HashPartition) -> Result<HashPartitionInfo> {
+        let HashPartition {
+            linear,
+            partition_num,
+            expr,
+        } = hash_stmt;
+
+        let definitions = parse_to_definition(partition_num);
+        // TODO: we should not manually convert it.
+        let expr = Expr::Column(Column::from_name(expr.to_string()));
```
chunshao90, line 34: Use match to convert `sql expr` to `datafusion expr`.

PR #488, opened 2022-12-19
feat: make the bloom filter optional in sst meta

ShiKaiWi:

Prepare for #489 

# Rationale for this change
 Currently, bloom filter index is necessary in the sst reader. However, this is actually optional.

- Make the bloom filter index in the sst reader optional.

None.

# How does this change test
Current tests.


jiacai2050 (approved):
LGTM

Issue #489, opened 2022-12-19
Reduce the memory consumption required by Compaction


ShiKaiWi:
### Describe This Problem

Current compaction procedure consumes much more memory than expected. The theoretical usage of memory should be a little more than the total size of all the input ssts, but now the actual consumption is 4-10 times theoretical value.

### Proposal

The reasons for the high memory consumption, I can come up with these:
- #486
- The input data which will be invalidated later should be put back into the cache;
- Some indexes (e.g. bloom filter) are not necessary to be loaded because predicates won't be used by compaction.

jiacai2050:
Fixed in main.

PR #490, opened 2022-12-19
feat: support store with readonly cache

ShiKaiWi:

Prepare #489 

# Rationale for this change
 Mentioned by #489, some data from the object store is not necessary to do caching so an object store with readonly cache are needed.

- Add object store with readonly cache for the instance.
- Object store with readonly cache is used for reading input data in compaction procedure.

None.

# How does this change test
Existing tests.

### DIFF
--- a/components/object_store/src/mem_cache.rs
+++ b/components/object_store/src/mem_cache.rs
@@ -158,6 +158,7 @@ pub struct MemCacheStore {
 }
 
 impl MemCacheStore {
+    /// Create a default [`MemCacheStore`].
     pub fn new(cache: MemCacheRef, underlying_store: ObjectStoreRef) -> Self {
         Self {
             cache,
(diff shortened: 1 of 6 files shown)


components/object_store/src/mem_cache.rs
```diff
@@ -209,6 +230,10 @@ impl ObjectStore for MemCacheStore {
         // TODO(chenxiang): What if two threads reach here? It's better to
         // pend one thread, and only let one to fetch data from underlying store.
         let bytes = self.underlying_store.get_range(location, range).await;
+        if self.read_only_cache {
```
jiacai2050, line 233: I think of one issue with current implementation.

Although this read only cache will ensure mem isn't wasted, it will make LRU useless, it may better to call `pop` method for this case.
- https://docs.rs/clru/0.6.1/clru/struct.CLruCache.html#method.peek

It will not update LRU position.

ShiKaiWi: 👌

jiacai2050 (approved):
LGTM

PR #491, opened 2022-12-19
feat: support ignore bloomfilter when compaction

ShiKaiWi:

Part #489 

# Rationale for this change
 After #488, bloom filter has been optional, but the compaction procedure has no way to tell the SST reader to skip loading bloom filter. This PR will implement it.

- Add `just_once` field to SST reader to decide whether to skip updating cache;
- SST reader will skip loading bloom filter if `just_once` is set and predicates are not provided;
- Remove sync sst reader;

None.

# How does this change test
Existing tests.


analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -914,6 +914,7 @@ impl SpaceStore {
                 sst_type: table_data.sst_type,
                 read_batch_row_num: table_options.num_rows_per_row_group,
                 reverse: false,
+                just_once: true,
```
jiacai2050, line 917: just_once seems awkward, how about introduce a `ReadType`
```
enum ReadType {
   Compact,
   Select,
   ...
}
impl ReadType {
  fn cache_sst(self) -> bool {

  }
  fn require_bloom_filter(&self) -> bool {
  }
}
```

ShiKaiWi: Replacing the boolean value with an enum sounds good, but it may not be a good pattern in terms of the two types of that. The sst reader is the dependency of`Query` and `Compaction` module, which means it should have no knowledge of its callers. Otherwise, the more read types are needed when more dependencies are introduced.

Maybe we should replace the `ReadType` with `ReadFrequency`:
```
enum ReadFrequency {
    Once,
    Repetitive,
}
```

jiacai2050: Sounds good.

jiacai2050 (approved):
LGTM

PR #493, opened 2022-12-19
fix: remove EncodingWriter

### LINKED ISSUE #53: try to replace parquet with parquet2?
**Description**

replace parquet with [parquet2](https://github.com/jorgecarleitao/parquet2)

The five main differentiators in comparison with `parquet` are:
* it uses `#![forbid(unsafe_code)]`
* delegates parallelism downstream
* decouples reading (IO intensive) from computing (CPU intensive)
* it is faster (10-20x when reading to arrow format)
* supports `async` read and write.
* It is integration-tested against pyarrow and (py)spark 3

**Proposal**

**Additional context**

jiacai2050:

Closes #53.

We don't have plan to switch to parquet2 any more.

# Rationale for this change
 
In old version of parquet, there is no way to get inner writer.

Upstream have added this method, so don't required this workaround.

No 
# How does this change test

Old unit test

### DIFF
--- a/analytic_engine/src/sst/parquet/encoding.rs
+++ b/analytic_engine/src/sst/parquet/encoding.rs
@@ -1,10 +1,6 @@
 // Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
 
-use std::{
-    convert::TryFrom,
-    io::Write,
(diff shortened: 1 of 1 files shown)


ShiKaiWi (approved):
LGTM

PR #494, opened 2022-12-20
feat: support query limit by rule

### LINKED ISSUE #470: Restriction on bad SQL for server stability
### Describe This Problem

Currently, one SQL with no predicates querying on a very big table may cause large consumption of CPU and memory, leading to instability of server.

### Proposal

We may need a module to do such things:
- Based provided rules, prevent the execution of any SQL which may lead to instability of server;
- Such rules are configurable, e.g. a rule for resolving the problem above may be called `disable_query_without_predicate`;

### Additional Context

_No response_

ShiKaiWi:

Closes #470 

# Rationale for this change
Refer to #470.

- Support limit query by `BlockRule`;
- Add a `BlockRule` called `QueryWithoutPredicate` to block the query like `select * from t`;
- Expose the operation of the block rules to http api service; 

Support block query with block rules, especially, `QueryWithoutPredicate`.

# How does this change test
New unit tests.

### DIFF
--- a/server/src/grpc/storage_service/write.rs
+++ b/server/src/grpc/storage_service/write.rs
@@ -56,13 +56,14 @@ pub(crate) async fn handle_write<Q: QueryExecutor + 'static>(
         );
         let plan = Plan::Insert(insert_plan);
 
-        if ctx.instance.limiter.should_limit(&plan) {
-            ErrNoCause {
-                code: StatusCode::TOO_MANY_REQUESTS,
-                msg: "Insert limited by reject list",
-            }
-            .fail()?;
-        }
+        ctx.instance
+            .limiter
+            .should_limit(&plan)
+            .map_err(|e| Box::new(e) as _)
+            .context(ErrWithCause {
+                code: StatusCode::FORBIDDEN,
+                msg: "Insert is blocked",
+            })?;
 
         let interpreter_ctx = InterpreterContext::builder(request_id)
             // Use current ctx's catalog and tenant as default catalog and tenant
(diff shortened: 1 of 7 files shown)


server/src/grpc/storage_service/write.rs
```diff
@@ -207,13 +208,14 @@ async fn create_table<Q: QueryExecutor + 'static>(
...
-    if instance.limiter.should_limit(&plan) {
-        ErrNoCause {
-            code: StatusCode::TOO_MANY_REQUESTS,
-            msg: "Create table limited by reject list",
-        }
-        .fail()?;
-    }
+    instance
+        .limiter
+        .try_limit(&plan)
```
chunshao90, line 213: There is no `blockRule` of `Create table is blocked`?

server/src/limiter.rs
```diff
@@ -2,20 +2,86 @@
...
+    BlockedByRule {
+        rule: BlockRule,
+        backtrace: Backtrace,
+    },
+}
+
+define_result!(Error);
+
+#[derive(Clone, Copy, Deserialize, Debug, PartialEq, Eq, Hash, Serialize, PartialOrd, Ord)]
+pub enum BlockRule {
```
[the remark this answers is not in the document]
ShiKaiWi: It sounds great! However, this PR may not do such refactoring. Let's refactor this in another PR.

server/src/grpc/storage_service/write.rs
ShiKaiWi: Actually, there is no such rule to forbid creating tables. However, it may be added in the future.

chunshao90 (approved):
LGTM

Issue #492, opened 2022-12-19
Support table partition


chunshao90:
### Describe This Problem

For very large tables, such as a single table of 10 billion+ rows, the single node cannot support such a large amount query and write workload. We need to support table partition.

### Proposal

Some jobs are needed:
1. Support create partition table.
     * [x] ceresmeta support create partition table and open sub table.  https://github.com/CeresDB/ceresmeta/pull/123
     * [x] ceresdb-server support parse sql which contains partition info. #487 
     * [x] ceresdb-server store table partition info, and create partition table. #485 
2. [x] Support query partition table. #512 
3. [x] Support write partition table. #512 
4. [x] Impl a ceresdb inner sdk, to communicate with each other ceresdb-server, transfer insert and query task to specific ceresdb-server. #509 
5. [x] Define new insert and query grpc service for inner sdk. #508 

### Additional Context

TODO:
+ Support `alter` on partition table.
+ Support `drop` partition table.
+ More complete partition rule implementation
  + support column type checking
  + support linear hash
  + support  `in`, `and`, `or` expressions

PR #498, opened 2022-12-21
feat: add implementation of key partition 

ZuLiangWang:
https://github.com/CeresDB/ceresdb/issues/492

# Rationale for this change
Support the use of partition key to create partition tables.
```
CREATE TABLE `demo` (`name` string TAG, `value` double NOT NULL, `t` timestamp NOT NULL, TIMESTAMP KEY(t)) PARTITION BY KEY(name) PARTITIONS 2 ENGINE=Analytic with (enable_ttl="false")
```

* Add key partition parser.
* Add pb definition of `KeyPartition` to persist partition info.

Users could create partition table by key partition sql.

# How does this change test

### DIFF
--- a/proto/protos/meta_update.proto
+++ b/proto/protos/meta_update.proto
@@ -28,7 +28,7 @@ message HashPartitionInfo {
 
 message KeyPartitionInfo {
   repeated Definition definitions = 1;
-  bytes partition_key = 2;
+  repeated string columns = 2;
   bool linear = 3;
 }
 
--- a/proto/protos/meta_update.proto
+++ b/proto/protos/meta_update.proto
@@ -26,6 +26,12 @@ message HashPartitionInfo {
   bool linear = 3;
 }
 
+message KeyPartitionInfo {
+  repeated Definition definitions = 1;
+  bytes partition_key = 2;
+  bool linear = 3;
+}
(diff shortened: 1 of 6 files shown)


sql/src/parser.rs
```diff
@@ -567,30 +570,100 @@ impl<'a> Parser<'a> {
...
+        }
+
+        let column_names = self.parser.parse_parenthesized_column_list(Mandatory)?;
+
+        // Partition key length must be 1, we do not supported default key now.
+        if column_names.len() != 1 {
+            return parser_err!(format!(
+                "except only one partition key, found key length:{:?}",
+                column_names.len()
+            ));
```
Rachelint, line 609: return parser_err!("Default partition key is unsupported now");

proto/protos/meta_update.proto
```diff
@@ -26,6 +26,12 @@ message HashPartitionInfo {
   bool linear = 3;
 }
 
+message KeyPartitionInfo {
+  repeated Definition definitions = 1;
+  bytes partition_key = 2;
```
Rachelint, line 31: repeated string columns = 2;

ShiKaiWi:
In favor of #506

Issue #495, opened 2022-12-20
Impose a custom format of SST


ShiKaiWi:
### Describe This Problem

Currently, the format of sst is just parquet, which may lead to some possible issues in the future upgrading, e.g. introducing some new format like TSM to organize the data.

### Proposal

Impose the new custom format of SST, which may look like:
```
┌───────────────┐
│               │
│               │
│    Payload    │
│               │
│               │
├───────────────┤
│    Footer     │
└───────────────┘
```

And the footer should contain these fields:
```
/// The version number of the format for upgrading compatibility.
version: u16
/// Format of payload, e.g. Parquet, TSM
payload_format: u16
/// A number denotes that this is our custom format.
magical number: u32
```

So that the footer will be a fixed size 64B.

### Additional Context

Although this custom format brings us some great benefits, the sst can't be used by the third-party applications which can process the parquet files.

### Related work
Currently, the sst type is used to decide which type of sst reader/builder to create, but this option is configured for table level, that is to say, one table can only use one sst type. I guess this is not reasonable, and it is more flexible and elegant to allow the data of one table being encoded in different sst formats. So I guess we make this option only works when deciding the type of sst builder, and add a new sst type called `Auto` for the flexibility of choosing a better sst format.

ShiKaiWi:
Here is the [TSM format](https://docs.influxdata.com/influxdb/v1.8/concepts/storage_engine/#tsm-files):
```
+--------+------------------------------------+-------------+--------------+
| Header |               Blocks               |    Index    |    Footer    |
|5 bytes |              N bytes               |   N bytes   |   4 bytes    |
+--------+------------------------------------+-------------+--------------+
+-------------------+
|      Header       |
+-------------------+
|  Magic  │ Version |
| 4 bytes │ 1 byte  |
+-------------------+
+---------+
│ Footer  │
+---------+
│Index Ofs│
│ 8 bytes │
+---------+
```

ShiKaiWi:
Here is rocksdb's [BlockBasedTable Format](https://github.com/facebook/rocksdb/wiki/Rocksdb-BlockBasedTable-Format):
```
<beginning_of_file>
[data block 1]
[data block 2]
...
[data block N]
[meta block 1: filter block]                  (see section: "filter" Meta Block)
[meta block 2: index block]
[meta block 3: compression dictionary block]  (see section: "compression dictionary" Meta Block)
[meta block 4: range deletion block]          (see section: "range deletion" Meta Block)
[...]

PR #499, opened 2022-12-21
chore: adjust usage of sst type

ShiKaiWi:

Prepare for #495 

# Rationale for this change
 Refer to #495:
> Currently, the sst type is used to decide which type of sst reader/builder to create, but this option is configured for table level, that is to say, one table can only use one sst type. I guess this is not reasonable, and it is more flexible and elegant to allow the data of one table being encoded in different sst formats. So I guess we make this option only works when deciding the type of sst builder, and add a new sst type called Auto for the flexibility of choosing a better sst format.

- Remove sst type option when reading sst;
- Add a new sst type called `Auto`;

None.

# How does this change test
Existing tests.


chunshao90 (approved):
LGTM

PR #504, opened 2022-12-22
fix: bugs when table migrate between shards

Rachelint:

Closes #

# Rationale for this change
While tables moved from shards, following bugs will be occurred:
+ manifest will drop, because `shard id` has changed, because now we locate manifest log by table id + shard id.
+ may loading a error last sequence, because we scan the wal data for this now. And while table just be moved to an new shard, no data will be scanned.
 

Fix above bugs.

None.

# How does this change test
Test by ut and manually.

### DIFF
--- a/analytic_engine/src/instance/mod.rs
+++ b/analytic_engine/src/instance/mod.rs
@@ -22,12 +22,13 @@ use std::{
     sync::{Arc, RwLock},
 };
 
+use common_types::table::TableId;
 use common_util::{define_result, runtime::Runtime};
 use log::info;
 use mem_collector::MemUsageCollector;
 use snafu::{ResultExt, Snafu};
 use table_engine::{engine::EngineRuntimes, remote::RemoteEngineRef};
-use wal::manager::WalManagerRef;
+use wal::manager::{WalLocation, WalManagerRef};
 
 use crate::{
     compaction::scheduler::CompactionSchedulerRef,
@@ -39,7 +40,7 @@ use crate::{
         file::FilePurger,
         meta_data::cache::MetaCacheRef,
     },
-    table::data::TableDataRef,
+    table::data::{TableDataRef, TableShardInfo},
     wal_synchronizer::WalSynchronizer,
     TableOptions,
 };
@@ -239,3 +240,12 @@ impl Instance {
 
 /// Instance reference
 pub type InstanceRef = Arc<Instance>;
+
+#[inline]
+pub(crate) fn create_wal_location(table_id: TableId, shard_info: TableShardInfo) -> WalLocation {
+    WalLocation::new(
+        shard_info.shard_id as u64,
+        shard_info.cluster_version,
+        table_id,
+    )
+}
--- a/analytic_engine/src/instance/write.rs
+++ b/analytic_engine/src/instance/write.rs
@@ -19,6 +19,7 @@ use tokio::sync::oneshot;
 use wal::manager::{SequenceNumber, WalLocation, WriteContext};
 
 use crate::{
+    instance,
     instance::{
         flush_compaction::TableFlushOptions,
         write_worker,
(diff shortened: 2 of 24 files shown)


analytic_engine/src/meta/meta_update.rs
```diff
@@ -457,7 +460,19 @@ pub struct MetaUpdateRequest {
 }
 
 impl MetaUpdateRequest {
-    pub fn new(location: WalLocation, meta_update: MetaUpdate) -> Self {
+    pub fn new(table_ident: TableIdentifier, meta_update: MetaUpdate) -> Self {
+        // Region id in manifest shouldn't change following by its moving from shards,
+        // so it should be mapped to `table_id`.
+        let versioned_region_id = VersionedRegionId {
+            version: table_ident.shard_info.cluster_version,
```
ShiKaiWi, line 467: Actually, I can't understand the necessity of the cluster version for `WalLocation`. Is it introduced to solve the problem caused by shard id reuse?

Rachelint: The detail about this, you can see the pr introducing it #422

analytic_engine/src/meta/meta_update.rs
```diff
@@ -457,7 +460,19 @@ pub struct MetaUpdateRequest {
...
+            id: table_ident.id,
+        };
+
+        let location = WalLocation {
```
jiacai2050, line 471: Any different with before? 

Isn't this location is the same with location passed in args?

jiacai2050 (changes requested):

jiacai2050:
@Rachelint How about rename this PR's title to 
```
fix: wrong last seq when table migrate between shards
```

analytic_engine/src/meta/meta_update.rs
Rachelint: It call `wal_location` to get it before.
But this returned `location` will pick the `shard_id` as `region_id`. But the `region_id` should not be mapped to `shard_id` in manifest, if so, the manifest info will lose after table migrating to new shard.

analytic_engine/src/instance/mod.rs
```diff
@@ -239,3 +239,12 @@ impl Instance {
 
 /// Instance reference
 pub type InstanceRef = Arc<Instance>;
+
+#[inline]
+pub(crate) fn create_wal_location(table_location: TableLocation) -> WalLocation {
```
jiacai2050, line 244:
```
Impl From<TableLocaltion> for WalLocation
```

analytic_engine/src/instance/write.rs
```diff
@@ -401,7 +401,12 @@ impl Instance {
 
         // Encode payload
         let payload = WritePayload::Write(&write_req_pb);
-        let wal_location = table_data.wal_location();
+        let table_location = table_data.table_location();
+        let wal_location = WalLocation::new(
```
jiacai2050, line 405:
```
WalLocation::from(..)
```

wal/src/table_kv_impl/table_unit.rs
```diff
@@ -405,38 +429,60 @@ impl TableUnit {
...
                 region_id,
                 table_id,
             )? {
-                return Ok(seq);
+                ensure!(
+                    sequence + 1 >= start_sequence,
+                    LoadLastSequence {
+                        table_id,
+                        msg: format!(
+                            "found last sequence, but next sequence after last
```
jiacai2050, line 461: It seems indent is wrong. How about

```
Found last sequence, but it's smaller than start sequece, last:{}, start:{}
```

Rachelint: in fact, it should be "last sequence + 1 < start sequence".
I found `start sequence` is equal to "flushed sequence + 1"...

jiacai2050 (approved):
LGTM

PR #505, opened 2022-12-22
chore: define remote_engine grpc service

chunshao90:

Closes #

# Rationale for this change
In the process of implementing the partition table, remote calls between ceresdb-server are involved.
So `remote_engine` grpc service is needed.

 Define remote_engine grpc service.

No.

# How does this change test
No need.

### DIFF
--- a/proto/protos/remote_engine.proto
+++ b/proto/protos/remote_engine.proto
@@ -1,0 +1,82 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+syntax = "proto3";
+package remote_engine;
+
+import "common.proto";
+
+message ResponseHeader {
+  uint32 code = 1;
+  string error = 2;
+}
+
+service RemoteEngineService {
+  rpc Read(ReadRequest) returns (stream ReadResponse) {}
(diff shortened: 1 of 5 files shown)


proto/protos/remote_engine.proto
```diff
@@ -0,0 +1,70 @@
...
+}
+
+message WriteRequest {
+  TableIdentifier table = 1;
+  RowGroup row_group = 2;
+}
+
+message WriteResponse {
+  uint64 affected_rows = 1;
+}
```
Rachelint, line 70: I think `\n` may be needed here?

proto/protos/remote_engine.proto
```diff
@@ -0,0 +1,70 @@
...
+  ReadOrder order = 5;
+}
+
+message ReadResponse {
+  repeated bytes rows = 1;
+}
+
+message RowGroup {
+  common.TableSchema table_schema = 1;
+  repeated bytes rows = 2;
```
Rachelint, line 58: Why using bytes to represent one row here? I feel a bit confused.

Rachelint (approved):
LGTM

PR #506, opened 2022-12-23
feat: support parse key partition

ShiKaiWi:

Closes #

# Rationale for this change
 In order to support the key partition feature, it is necessary to support parse the statement.

- Support parse key partition;

Now user can use key partition statement to create table.

# How does this change test
TODO

### DIFF
--- a/interpreters/src/show_create.rs
+++ b/interpreters/src/show_create.rs
@@ -139,6 +139,7 @@ impl ShowCreateInterpreter {
                     .as_str()
                 }
             }
+            PartitionInfo::Key(_) => {}
         }
 
         // TODO: update datafusion to remove `#`.
--- a/interpreters/src/show_create.rs
+++ b/interpreters/src/show_create.rs
@@ -140,24 +140,18 @@ impl ShowCreateInterpreter {
                 }
             }
             PartitionInfo::Key(v) => {
-                let partition_key = match Expr::from_bytes(&v.partition_key) {
-                    Ok(expr) => expr,
-                    Err(e) => {
-                        error!("show create table parse partition info failed, err:{}", e);
-                        return res;
-                    }
-                };
+                let rendered_partition_key = v.partition_key.join(",");
                 if v.linear {
                     res += format!(
                         " PARTITION BY LINEAR KEY({}) PARTITIONS {}",
-                        partition_key,
+                        rendered_partition_key,
                         v.definitions.len()
                     )
                     .as_str()
                 } else {
                     res += format!(
                         " PARTITION BY KEY({}) PARTITIONS {}",
-                        partition_key,
+                        rendered_partition_key,
                         v.definitions.len()
                     )
                     .as_str()
@@ -223,7 +217,7 @@ mod test {
 
     #[test]
     fn test_render_key_partition_info() {
-        let expr = col("col1");
+        let partition_key_col_name = "col1";
         let partition_info = PartitionInfo::Key(KeyPartitionInfo {
             definitions: vec![
                 Definition {
@@ -235,7 +229,7 @@ mod test {
                     origin_name: None,
                 },
             ],
-            partition_key: expr.to_bytes().unwrap(),
+            partition_key: vec![partition_key_col_name.to_string()],
             linear: false,
         });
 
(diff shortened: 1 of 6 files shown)


interpreters/src/show_create.rs
```diff
@@ -139,6 +139,7 @@ impl ShowCreateInterpreter {
                     .as_str()
                 }
             }
+            PartitionInfo::Key(_) => {}
```
ShiKaiWi, line 142: Maybe this should be implemented too. Let me fix it.

ShiKaiWi: @ZuLiangWang has added in his repo, and I will cherry-pick it here.

sql/src/parser.rs
```diff
@@ -593,6 +579,78 @@ impl<'a> Parser<'a> {
...
+            let tag_option = col_def.options.iter().find(|opt| {
+                opt.option == ColumnOption::DialectSpecific(vec![Token::make_keyword(TAG)])
+            });
+            if tag_option.is_none() {
+                return parser_err!(format!(
+                    "partition key must be tag, key name:{:?}",
+                    key_col.value
+                ));
+            }
+        }
```
Rachelint, line 626: The check logic here is same as hash partition, maybe extract a common function is better?

table_engine/src/partition/mod.rs
```diff
@@ -11,12 +11,14 @@ use proto::meta_update as meta_pb;
...
 pub enum PartitionInfo {
     Hash(HashPartitionInfo),
+    Key(KeyPartitionInfo),
 }
 
 impl PartitionInfo {
     pub fn get_definitions(&self) -> Vec<Definition> {
         match self {
             Self::Hash(v) => v.definitions.clone(),
+            PartitionInfo::Key(v) => v.definitions.clone(),
```
Rachelint, line 21: Unify to `Self` or `PartitionInfo`?

sql/src/parser.rs
```diff
@@ -593,6 +579,78 @@ impl<'a> Parser<'a> {
...
+        } else if self.consume_tokens(&["LINEAR", "KEY"]) {
+            true
+        } else {
+            return Ok(None);
+        };
+
+        let key_columns = self.parser.parse_parenthesized_column_list(Mandatory)?;
+
+        // Ensure at least one column for partition key.
+        if key_columns.is_empty() {
```
ZuLiangWang, line 597: In fact, the logic here is useless at present, because when the SQL is' KEY () ', `parser.parse_parenthesized_column_list` will directly throw an error. If you want to tell the user the exact cause of the error, you need to modify the parse logic.

ShiKaiWi: From the source code, it is yes. However, it can't be concluded from the signature from the `parse_parenthesized_column_list`. So I guess we should keep the check here.

ZuLiangWang: Maybe we should leave a todo here to record this problem. Later, we can consider using our own `parse_parenthesized_column_list` to replace the native method of `sqlparser`.

ShiKaiWi: I see. What you are afraid of is the strange error reported to user. I will add some customized error here.

ZuLiangWang (approved):
LGTM

sql/src/parser.rs
```diff
@@ -1289,4 +1344,52 @@ mod tests {
...
+                _ => panic!("failed"),
+            }
+        }
+
+        fn default_key_partition() {
+            let sql = r#"CREATE TABLE `demo` (`name` string TAG, `value` double NOT NULL, `t` timestamp NOT NULL, TIMESTAMP KEY(t)) PARTITION BY KEY() PARTITIONS 2 ENGINE=Analytic with (enable_ttl="false")"#;
+            let stmt = Parser::parse_sql(sql);
+            assert_eq!(
+                stmt.err().unwrap(),
+                ParserError("Expected identifier, found: )".to_string())
```
ZuLiangWang, line 1381: Because the error information returned above has been modified, the unit test here also needs to be modified.
``ParserError::ParserError(format!("Fail to parse partition key, err:{}", e))``

ShiKaiWi: Understood.

PR #511, opened 2022-12-26
feat: auto forward query for grpc

### LINKED ISSUE #147: Auto-redirect traffic to specific instances
**Description**
In some scenarios where CeresDB servers are not accessible for client nodes because of the network isolation, the current end-to-end communications won't work any more.

**Proposal**
Make the CeresDB server able to auto-redirect the traffic to the right instances according to the routing inforamtion.

**Additional context**
We can only make gRPC service work first.

ShiKaiWi:

Part of #147 

# Rationale for this change
 Check #147 

- Add `forward` module to do traffic auto redirecting for grpc query;

Support forward config.

# How does this change test
New unit test.

### DIFF
--- a/server/src/grpc/storage_service/mod.rs
+++ b/server/src/grpc/storage_service/mod.rs
@@ -36,6 +36,7 @@ use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::metadata::{KeyAndValueRef, MetadataMap};
 
+use super::forward::ForwarderRef;
 use crate::{
     consts,
     grpc::{
@@ -98,6 +99,7 @@ pub struct HandlerContext<'a, Q> {
     catalog: String,
     schema: String,
     schema_config: Option<&'a SchemaConfig>,
+    forwarder: ForwarderRef,
 }
 
 impl<'a, Q> HandlerContext<'a, Q> {
@@ -106,6 +108,7 @@ impl<'a, Q> HandlerContext<'a, Q> {
         router: Arc<dyn Router + Sync + Send>,
         instance: InstanceRef<Q>,
         schema_config_provider: &'a SchemaConfigProviderRef,
+        forwarder: ForwarderRef,
     ) -> Result<Self> {
         let default_catalog = instance.catalog_manager.default_catalog_name();
         let default_schema = instance.catalog_manager.default_schema_name();
@@ -147,6 +150,7 @@ impl<'a, Q> HandlerContext<'a, Q> {
             catalog,
             schema,
             schema_config,
+            forwarder,
         })
     }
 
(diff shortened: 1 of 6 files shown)


server/src/grpc/storage_service/mod.rs
```diff
@@ -36,6 +36,7 @@ use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::metadata::{KeyAndValueRef, MetadataMap};
 
+use super::forward::ForwarderRef;
```
jiacai2050, line 39: No relative import.

server/src/grpc/storage_service/query.rs
```diff
@@ -43,10 +51,66 @@ fn empty_ok_resp() -> QueryResponse {
...
+
+        Box::new(query) as _
+    };
+
+    match forwarder.forward(forward_req, do_query).await {
+        Ok(forward_res) => match forward_res {
+            ForwardResult::Forwarded(v) => Some(v),
+            ForwardResult::Original => None,
+        },
+        Err(e) => {
```
jiacai2050, line 98: why not return error directly when this fail?

server/src/server.rs
```diff
@@ -319,6 +319,9 @@ impl<Q: QueryExecutor + 'static> Builder<Q> {
             .context(MissingSchemaConfigProvider)?;
         let rpc_services = grpc::Builder::new()
             .endpoint(Endpoint::new(self.config.bind_addr, self.config.grpc_port).to_string())
+            .local_endpoint(
+                Endpoint::new(self.config.cluster.node.addr, self.config.grpc_port).to_string(),
```
jiacai2050, line 323: Does this mean we have to config every node ip in order to avoid forward to itself?

This seems not easy to configure.

jiacai2050 (changes requested):

server/src/grpc/storage_service/query.rs
ShiKaiWi: For error caused by `Forward` itself, I guess we should give it a try to the local service.

server/src/server.rs
ShiKaiWi: > Does this mean we have to config every node ip in order to avoid forward to itself?

Yes.

> This seems not easy to configure.

Actually, it is easy. Firstly, the forwarding is disabled by default. Secondly, the forwarding makes sense only when run ceresdb in cluster mode, and the `config.cluster.node.addr` must be provided which will be used as the forwarder's local endpoint if we start the ceresdb in cluster mode. In conclusion, the local endpoint of forwarder won't introduce extra configurations besides its own configurations.

jiacai2050: Correct me if I misunderstand something

The current request flow is:
```
sdk -> route -> server
```
After your patches, the request flow will be this?
```
sdk -> route -> server -> forwarded server
```
I couldn't quite understand why this forward is required.

ShiKaiWi: The reason has been explained in https://github.com/CeresDB/ceresdb/issues/147:
> In some scenarios where CeresDB servers are not accessible for client nodes because of the network isolation, the current end-to-end communications won't work any more.

jiacai2050: Then it sound like this?
```
sdk -> server(served as Querier) -> route -> forwarded server
```

ShiKaiWi: Yes. I will update the comment by appending the diagram to it.

jiacai2050 (approved):
LGTM

PR #513, opened 2022-12-27
fix: remove size from sst metadata

### LINKED ISSUE #321: SstMetaData's size is 0 after flush
### Describe this problem

When test #312, I find size of SstMetaData is 0 after flush memtable, this cause following error when restart(reading sst)

```
2022-10-19 11:49:32.342 ERRO [common_util/src/panic.rs:42] thread 'main' panicked 'Failed to create catalog manager: VisitSysCatalog { source: ReadStream { source: ErrWithSource { msg: "Read record batch", source: ReadFromSubIter { source: PullRecordBat
ch { source: DecodeRecordBatch { source: ExternalError(ParquetError(General("CachableParquetFileReader::get_metadata error: Execution error: file size of 0 is less than footer"))) } } } } } }' at "src/setup.rs:187"
   0: common_util::panic::set_panic_hook::{{closure}}
```

### Steps to reproduce

1. Insert some data
```sql
CREATE TABLE `cpu2` (
`my_ts` timestamp NOT NULL, 
 `t1` string tag,  
 `t2` string tag,  
 `value` double,
  TIMESTAMP KEY(my_ts)) 
  with (
enable_ttl = 'false'
);

insert into cpu2  (my_ts, t1,t2, `value`) 
values
(123, 'web', 'hz', 100),
(124, 'web','hz', 101),
(123, 'ios', 'sh', 100),
(124, 'ios', 'sh', 101);
```
2. flush
```bash
curl --location --request POST 'localhost:5440/flush_memtable'
```

3. restart service

### Expected behavior

No panic

### Additional Information

 memtable is write to sst in
https://github.com/CeresDB/ceresdb/blob/c557007e6c56390d12debe2d69478dcd73ae4796/analytic_engine/src/instance/flush_compaction.rs#L641

but we update size after this, in https://github.com/CeresDB/ceresdb/blob/c557007e6c56390d12debe2d69478dcd73ae4796/analytic_engine/src/instance/flush_compaction.rs#L657
which is too late, since it's already persisted.

jiacai2050:

Closes #321

# Rationale for this change
 

1. Remove size, row_num from sst metadata. Those two fields cannot be known before write sst, so we have to store them in other place(manifest now).
2. When read sst, pass one more arg: file_size, which is stored inside FileHandle.

No 

# How does this change test

Existing UT

### DIFF
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -692,8 +692,8 @@ impl Instance {
         }
         batch_record_senders.clear();
 
-        let ret = try_join_all(sst_handlers).await;
-        for (idx, info_and_meta) in ret.context(RuntimeJoin)?.into_iter().enumerate() {
+        let info_and_metas = try_join_all(sst_handlers).await.context(RuntimeJoin)?;
+        for (idx, info_and_meta) in info_and_metas.into_iter().enumerate() {
             let (sst_info, sst_meta) = info_and_meta?;
             files_to_level0.push(AddFile {
                 level: 0,
@@ -975,12 +975,14 @@ impl SpaceStore {
                 path: sst_file_path.to_string(),
             })?;
 
+        let sst_file_size = sst_info.file_size as u64;
+        let sst_row_num = sst_info.row_num as u64;
         table_data
             .metrics
-            .compaction_observe_output_sst_size(sst_info.file_size as u64);
+            .compaction_observe_output_sst_size(sst_file_size);
         table_data
             .metrics
-            .compaction_observe_output_sst_row_num(sst_info.row_num as u64);
+            .compaction_observe_output_sst_row_num(sst_row_num);
 
         info!(
             "Instance files compacted, table:{}, table_id:{}, request_id:{}, output_path:{}, input_files:{:?}, sst_meta:{:?}",
@@ -1006,8 +1008,8 @@ impl SpaceStore {
             level: input.output_level,
             file: FileMeta {
                 id: file_id,
-                size: sst_info.file_size as u64,
-                row_num: sst_info.row_num as u64,
+                size: sst_file_size,
+                row_num: sst_row_num,
                 meta: sst_meta,
             },
         });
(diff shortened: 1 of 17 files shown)


analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -698,12 +693,15 @@ impl Instance {
         batch_record_senders.clear();
 
         let ret = try_join_all(sst_handlers).await;
```
Rachelint, line 695:
```suggestion
        let info_and_metas = try_join_all(sst_handlers).await.context(RuntimeJoin)?;
```

analytic_engine/src/sst/factory.rs
```diff
@@ -169,9 +148,10 @@ impl Factory for FactoryImpl {
         options: &SstReaderOptions,
         path: &'a Path,
         store_picker: &'a ObjectStorePickerRef,
+        file_size: u64,
```
Rachelint, line 151: How about defining a struct(for example, `FileContext`?) to wrap `path` and `file_size` ?

analytic_engine/src/sst/file.rs
```diff
@@ -424,6 +424,10 @@ impl FileHandleSet {
 pub struct FileMeta {
     /// Id of the sst file
     pub id: FileId,
+    /// file size in bytes
+    pub size: u64,
+    // total row number
```
Rachelint, line 429: /// total row number

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -979,16 +975,12 @@ impl SpaceStore {
...
-        sst_meta.size = sst_info.file_size as u64;
-
         table_data
             .metrics
-            .compaction_observe_output_sst_size(sst_meta.size);
+            .compaction_observe_output_sst_size(sst_info.file_size as u64);
         table_data
             .metrics
-            .compaction_observe_output_sst_row_num(sst_meta.row_num);
+            .compaction_observe_output_sst_row_num(sst_info.row_num as u64);
```
Rachelint, line 983: How about
```
let sst_file_size  = sst_info.file_size as u64;
let sst_row_num = sst_info.row_num as u64;
```
And reuse it here and follows?

jiacai2050: fixed

analytic_engine/src/sst/factory.rs
jiacai2050: Can't think of any advantage to wrap them together.

Rachelint (approved):
LGTM

PR #519, opened 2022-12-29
fix: fix conversion by avro

Rachelint:

Closes #
Part of #492 

# Rationale for this change
We convert all `int` related types to `int32` and `int64`, because only these two int types in `avro` now.
So when converting `Value`(`avro` datum) to `Datum`, we should convert it back to the right type in schema.
  

Modify `avro_value_to_datum` in `avro.rs` for converting back to the right type.

None.

# How does this change test
Test by ut.


chunshao90 (approved):
LGTM

PR #521, opened 2022-12-29
fix: fix convert record batch

ZuLiangWang:

Closes #

# Rationale for this change
Now  `convert_records` will intercept the `record_ batches` only uses the first element, which will result in multiple `records_batch` in the query result is truncated.

* Traverse `record_batch` to return all data.

None.

# How does this change test
Pass existing unit test.


chunshao90 (approved):
LGTM

PR #522, opened 2022-12-30
fix: fix partition table affected_rows

chunshao90:

Closes #

# Rationale for this change
 When insert into a partition table, the affected_rows result is wrong.

* Fix affected_rows.

No.

# How does this change test
Manual testing.


Rachelint (approved):
LGTM

PR #523, opened 2022-12-30
chore: add more context for query log

jiacai2050:

Closes #

# Rationale for this change
 
Some log are too verbose, so PR try to normalize them

Update log when reading sst

No

# How does this change test

No need

### DIFF
--- a/analytic_engine/src/sst/parquet/async_reader.rs
+++ b/analytic_engine/src/sst/parquet/async_reader.rs
@@ -3,6 +3,7 @@
 //! Sst reader implementation based on parquet.
 
 use std::{
+    fmt,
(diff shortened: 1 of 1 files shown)


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -395,13 +391,19 @@ impl ParallelismOptions {
     }
 }
 
-#[derive(Debug, Clone)]
 struct ReaderMetrics {
```
jiacai2050, line 394: This should not be cloneable, it will duplicate the counter

Rachelint (approved):
LGTM

PR #526, opened 2022-12-30
fix: rename NumKeyColumns to PrimaryKeyIndexes

### LINKED ISSUE #419: whether to need to adjust the `ArrowSchemaMetaKey: : NumKeyColumns`
### Describe This Problem

after the modification of #340 . 
The concept of `NumKeyColumns` is no longer important, so does it need to be changed when building arrow schema meta data here?
https://github.com/CeresDB/ceresdb/blob/c0f498bca3cf1efc36936ff3b591b7f22621ddbd/common_types/src/schema.rs#L1129-L1134

### Proposal

Maybe save the `primary_key_indexes`?

### Additional Context

This may run into serialization problems

dust1:

Closes #419 

# Rationale for this change
 
NumKeyColumns is no longer useful anymore, and should be removed.

I add a new struct `Indexes`  to hold the primary key indexes.

none

# How does this change test

use unit test. in `test_build_from_arrow_schema`


jiacai2050:
@dust1 Please update PR's description to tell us how you test this, and if there are any user changes.

common_types/src/schema.rs
```diff
@@ -250,15 +281,15 @@ impl TryFrom<&HashMap<String, String>> for ArrowSchemaMeta {
...
+    PrimaryKeyIndexes,
     TimestampIndex,
     Version,
 }
 
 impl ArrowSchemaMetaKey {
     fn as_str(&self) -> &str {
         match self {
-            Self::NumKeyColumns => "schema:num_key_columns",
+            Self::PrimaryKeyIndexes => "schema:primary_key_indexes",
```
jiacai2050, line 292: `schema::primary_key_indexes`

common_types/src/schema.rs
```diff
@@ -1106,8 +1133,12 @@ impl Builder {
     fn build_arrow_schema_meta(&self) -> HashMap<String, String> {
         [
             (
-                ArrowSchemaMetaKey::NumKeyColumns.to_string(),
-                self.primary_key_indexes.len().to_string(),
+                ArrowSchemaMetaKey::PrimaryKeyIndexes.to_string(),
+                self.primary_key_indexes
```
jiacai2050, line 1137: Isn't this duplicated with ToString method?

common_types/src/schema.rs
```diff
@@ -206,11 +207,41 @@ pub enum CompatError {
...
+    }
+}
+
+impl FromStr for Indexes {
+    type Err = ParseIntError;
+
+    fn from_str(s: &str) -> std::result::Result<Indexes, ParseIntError> {
+        let list = s
+            .split(',')
+            .filter_map(|n| match n.parse::<usize>() {
```
jiacai2050, line 236: It should return Error when parse failed.

jiacai2050:
@dust1 I realize this is a breaking change, you need update target branch to https://github.com/CeresDB/ceresdb/tree/breaking-changes.

Main doesn't accept breaking change now.

dust1:
ok

common_types/src/schema.rs
```diff
@@ -206,11 +207,48 @@ pub enum CompatError {
...
 #[derive(Default)]
 pub struct ArrowSchemaMeta {
-    num_key_columns: usize,
+    primary_key_indexes: Indexes,
     timestamp_index: usize,
     version: u32,
 }
 
+#[derive(Debug, Default, PartialEq)]
+struct Indexes {
```
jiacai2050, line 216:
```
struct Indexes(Vec<usize>);
```

dust1:
Oops. I can't find the Settings in this pr to modify the target branch, I may need to recreate a pr

common_types/src/schema.rs
```diff
@@ -206,11 +207,48 @@ pub enum CompatError {
...
+    pub indexes: Vec<usize>,
+}
+
+impl ToString for Indexes {
+    fn to_string(&self) -> String {
+        format!("{:?}", self.indexes)
```
jiacai2050, line 222: We shouldn't rely on `debug format` to format indexes, what if it changed between different Rust version?
```
self.indexes.join(',')
```

This is enough.

jiacai2050 (changes requested):

common_types/src/schema.rs
```diff
@@ -1106,8 +1140,8 @@ impl Builder {
     fn build_arrow_schema_meta(&self) -> HashMap<String, String> {
         [
             (
-                ArrowSchemaMetaKey::NumKeyColumns.to_string(),
-                self.primary_key_indexes.len().to_string(),
+                ArrowSchemaMetaKey::PrimaryKeyIndexes.to_string(),
+                format!("{:?}", self.primary_key_indexes),
```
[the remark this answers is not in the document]
dust1: `self.primary_key_indexes` is `Vec<usize>`, it not implement `ToString`, so I use `join` to generate the value in the `HashMap`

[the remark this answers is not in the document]
jiacai2050: The problem is we need to make sure to_string/parse_string is on pair.

I think you can just wrap it like this
```
Indexes(self.primary_key_indexes).to_string()
```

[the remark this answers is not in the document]
dust1: got it!

jiacai2050:
I think this PR is ready for merge, I will push some minor changes.

Thanks @dust1

jiacai2050 (approved):
LGTM

PR #528, opened 2022-12-31
feat: add pre commit hooks

jiacai2050:

Closes #

# Rationale for this change
 
There are some convention in our coding rules, it's annoying to manually fix them.

So PR add git commit hooks to automatically fix them.

Add pre-commit config, it will
- Ensure validation of toml/yaml files
- Ensure Cargo.toml is sorted, depend on https://github.com/DevinR528/cargo-sort
- Ensure license header

Developers can install those hooks via
```
pre-commit install
```
This command only need to execute once, it will work when `git commit` next time.

# How does this change test

Manually

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,24 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: check-toml
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.3.1
+    hooks:
+      - id: insert-license
+        files: \.rs$
+        args:
+          - --license-filepath
+          - scripts/license-header.txt
+          - --comment-style
+          - //
+  - repo: local
+    hooks:
+      - id: cargo-sort
+        name: Check Cargo.toml is sorted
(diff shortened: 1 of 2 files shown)


scripts/license-header.txt
```diff
@@ -0,0 +1 @@
+Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
```
Rachelint, line 1: How about placing the license like [iox](https://github.com/influxdata/influxdb_iox)?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+        args:
+          - --license-filepath
+          - scripts/license-header.txt
+          - --comment-style
+          - //
+  - repo: local
+    hooks:
+      - id: cargo-sort
+        name: Check Cargo.toml is sorted
+        entry: cargo-sort
```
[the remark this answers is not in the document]
jiacai2050: Not required, cargo.toml will be passed as args directly to cargo-sort

scripts/license-header.txt
jiacai2050: There is already an license file in root directory, but it's different with header inserted in source code, so I create this new file.

Rachelint (approved):
LGTM

PR #530, opened 2023-01-03
feat: support querying partition tables through select in

ZuLiangWang:

Closes #

# Rationale for this change
We implemented table partition in the last release, but querying partitioned tables through `select in` will lead to the failure to identify the correct partitions, resulting in full table scanning. We need to support select in to improve query performance.

* Implement parsing of `select in` in `filterExtractor`

None.

# How does this change test
Pass unit test.

### DIFF
--- a/table_engine/src/partition/rule/df_adapter/extractor.rs
+++ b/table_engine/src/partition/rule/df_adapter/extractor.rs
@@ -63,10 +63,39 @@ impl FilterExtractor for KeyExtractor {
                     (Expr::Column(col), Operator::Eq, Expr::Literal(val))
                     | (Expr::Literal(val), Operator::Eq, Expr::Column(col)) => {
                         let datum_opt = Datum::from_scalar_value(&val);
-                        datum_opt.map(|d| PartitionFilter::new(col.name, PartitionCondition::Eq(d)))
+                        datum_opt.map(|datum| {
+                            PartitionFilter::new(col.name, PartitionCondition::Eq(datum))
+                        })
                     }
                     _ => None,
                 },
+                Expr::InList {
+                    expr,
+                    list,
+                    negated,
+                } => {
+                    if let (Expr::Column(col), list, false) = (*expr, list, negated) {
+                        let mut datums = Vec::with_capacity(list.len());
+                        for entry in list {
+                            if let Expr::Literal(val) = entry {
+                                let datum_opt = Datum::from_scalar_value(&val);
+                                if let Some(datum) = datum_opt {
+                                    datums.push(datum)
+                                }
+                            }
+                        }
+                        if datums.is_empty() {
+                            None
+                        } else {
+                            Some(PartitionFilter::new(
+                                col.name,
+                                PartitionCondition::In(datums),
+                            ))
+                        }
+                    } else {
+                        None
+                    }
+                }
                 _ => None,
             };
 
@@ -84,7 +113,7 @@ pub type FilterExtractorRef = Box<dyn FilterExtractor>;
 #[cfg(test)]
 mod tests {
     use datafusion::scalar::ScalarValue;
-    use datafusion_expr::col;
+    use datafusion_expr::{col, Expr::Literal};
 
     use super::{FilterExtractor, *};
 
@@ -106,5 +135,41 @@ mod tests {
         let rejected_expr = col("col1").gt(Expr::Literal(ScalarValue::Int32(Some(42))));
         let partition_filter = extractor.extract(&[rejected_expr], &columns);
         assert!(partition_filter.is_empty());
+    }
+
+    #[test]
+    fn test_key_extractor_in_list_filter() {
+        let extractor = KeyExtractor;
+
+        let columns = vec!["col1".to_string()];
+        let accepted_expr = col("col1").in_list(
+            vec![
+                Literal(ScalarValue::Int32(Some(42))),
+                Literal(ScalarValue::Int32(Some(38))),
+            ],
+            false,
+        );
+        let partition_filter = extractor.extract(&[accepted_expr], &columns);
+        let expected = PartitionFilter {
+            column: "col1".to_string(),
+            condition: PartitionCondition::In(vec![Datum::Int32(42), Datum::Int32(38)]),
+        };
+        assert_eq!(partition_filter.get(0).unwrap(), &expected);
+    }
+
+    #[test]
+    fn test_key_extractor_in_list_filter_with_negated() {
+        let extractor = KeyExtractor;
+
+        let columns = vec!["col1".to_string()];
+        let accepted_expr = col("col1").in_list(
+            vec![
+                Literal(ScalarValue::Int32(Some(42))),
+                Literal(ScalarValue::Int32(Some(38))),
+            ],
+            true,
+        );
+        let partition_filter = extractor.extract(&[accepted_expr], &columns);
+        assert!(partition_filter.is_empty())
     }
 }


table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
                     }
                     _ => None,
                 },
+                Expr::InList {
+                    expr: _,
```
Rachelint, line 71: We should check the expr, we only process  Expr::Column(col).

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
...
+                    list,
+                    negated,
+                } => {
+                    // The 'not in' syntax is not currently supported.
+                    if negated {
+                        continue;
+                    }
```
Rachelint, line 78: We can use tuple match to check here and above, for example:
`if let (Expr::Column(col), true) = (*expr, negated) `

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
...
+                    let mut v = Vec::with_capacity(list.len());
+                    for e in list {
```
Rachelint, line 80: Maybe e -> entry

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
...
+                        if let Expr::Literal(val) = e {
+                            let datum_opt = Datum::from_scalar_value(&val);
+                            if let Some(d) = datum_opt {
```
Rachelint, line 83: d -> datum

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
...
+                                v.push(d)
+                            }
+                        }
+                    }
+                    let column_name = columns_in_filter.iter().next().unwrap();
```
Rachelint, line 88: We should use the column name which get from `expr`.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
...
+                Expr::InList {
+                    expr: _,
+                    list,
+                    negated,
+                } => {
+                    // The 'not in' syntax is not currently supported.
+                    if negated {
+                        continue;
+                    }
+                    let mut v = Vec::with_capacity(list.len());
```
Rachelint, line 79: v -> datums

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -67,6 +67,30 @@ impl FilterExtractor for KeyExtractor {
...
+                            if let Some(d) = datum_opt {
+                                v.push(d)
+                            }
+                        }
+                    }
+                    let column_name = columns_in_filter.iter().next().unwrap();
+                    Some(PartitionFilter::new(
+                        column_name.clone(),
+                        PartitionCondition::In(v),
+                    ))
```
Rachelint, line 92: Check datums, if it is empty, we should return None.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -107,4 +131,24 @@ mod tests {
...
+    fn test_key_extractor_select_in() {
+        let extractor = KeyExtractor;
+
+        let columns = vec!["col1".to_string()];
+        let accepted_expr = col("col1").in_list(
+            vec![
+                Literal(ScalarValue::Int32(Some(42))),
+                Literal(ScalarValue::Int32(Some(38))),
+            ],
+            false,
```
Rachelint, line 145: Negated should be tested.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -107,4 +131,24 @@ mod tests {
         let partition_filter = extractor.extract(&[rejected_expr], &columns);
         assert!(partition_filter.is_empty());
     }
+
+    #[test]
+    fn test_key_extractor_select_in() {
```
Rachelint, line 136: Maybe select_in -> in_list_fitler.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -63,7 +63,35 @@ impl FilterExtractor for KeyExtractor {
...
+                    }
+                    _ => None,
+                },
+                Expr::InList {
+                    expr,
+                    list,
+                    negated,
+                } => match (*expr, list, negated) {
+                    // The 'not in' syntax is not currently supported.
+                    (Expr::Column(col), list, false) => {
```
Rachelint, line 78: `list` is unneccessary to match.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -63,7 +63,35 @@ impl FilterExtractor for KeyExtractor {
...
+                        })
+                    }
+                    _ => None,
+                },
+                Expr::InList {
+                    expr,
+                    list,
+                    negated,
+                } => match (*expr, list, negated) {
+                    // The 'not in' syntax is not currently supported.
```
Rachelint, line 77: Can not support actually, maybe this comment can be removed.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -63,7 +63,35 @@ impl FilterExtractor for KeyExtractor {
...
+                                }
+                            }
+                        }
+                        match datums.is_empty() {
+                            false => Some(PartitionFilter::new(
+                                col.name,
+                                PartitionCondition::In(datums),
+                            )),
+                            true => None,
+                        }
```
Rachelint, line 94: For bool, usually use `if else`.

table_engine/src/partition/rule/df_adapter/extractor.rs
```diff
@@ -107,4 +135,40 @@ mod tests {
...
+        let columns = vec!["col1".to_string()];
+        let accepted_expr = col("col1").in_list(
+            vec![
+                Literal(ScalarValue::Int32(Some(42))),
+                Literal(ScalarValue::Int32(Some(38))),
+            ],
+            true,
+        );
+        let partition_filter = extractor.extract(&[accepted_expr], &columns);
+        assert_eq!(partition_filter.len(), 0);
```
Rachelint, line 172: Usually check `is_empty`.

table_engine/src/partition/rule/df_adapter/extractor.rs
ZuLiangWang: `list` is moved in  `match (*expr, list, negated)`, so we must declare it again.

Rachelint: It seems strange?
How about:
```suggestion
{
      if  let (Expr::Column(col), false) = (Expr::Column(col), false) {
         ...
      } else {
           None
      }
 }
```

Rachelint:
LGTM

Rachelint (approved):
LGTM

PR #532, opened 2023-01-03
feat: support drop partition table

chunshao90:

Closes #

# Rationale for this change
 Support drop partition table.

Add `PartitionInfo` in `DropTablePlan` to support drop partition table.

Support drop partition table.

# How does this change test
Manual testing with `ceresmeta`.

### DIFF
--- a/interpreters/src/table_manipulator/meta_based.rs
+++ b/interpreters/src/table_manipulator/meta_based.rs
@@ -105,9 +105,19 @@ impl TableManipulator for TableManipulatorImpl {
         plan: DropTablePlan,
         _table_engine: TableEngineRef,
     ) -> Result<Output> {
+        let sub_table_names = plan.partition_info.clone().map(|v| {
+            v.get_definitions()
+                .iter()
+                .map(|def| format_sub_partition_table_name(&plan.table, &def.name))
+                .collect::<Vec<String>>()
+        });
+        let partition_table_info =
+            sub_table_names.map(|v| PartitionTableInfo { sub_table_names: v });
+
         let req = DropTableRequest {
             schema_name: ctx.default_schema().to_string(),
             name: plan.table,
+            partition_table_info,
         };
 
         let resp = self
--- a/interpreters/src/table_manipulator/meta_based.rs
+++ b/interpreters/src/table_manipulator/meta_based.rs
@@ -48,8 +48,7 @@ impl TableManipulator for TableManipulatorImpl {
                 ),
             })?;
 
-        let partition_table_info =
-            convert_to_partition_table_info(&plan.table, &plan.partition_info);
+        let partition_table_info = create_partition_table_info(&plan.table, &plan.partition_info);
 
         let encoder = PartitionInfoEncoder::default();
 
@@ -101,8 +100,7 @@ impl TableManipulator for TableManipulatorImpl {
         plan: DropTablePlan,
         _table_engine: TableEngineRef,
     ) -> Result<Output> {
-        let partition_table_info =
-            convert_to_partition_table_info(&plan.table, &plan.partition_info);
+        let partition_table_info = create_partition_table_info(&plan.table, &plan.partition_info);
 
         let req = DropTableRequest {
             schema_name: ctx.default_schema().to_string(),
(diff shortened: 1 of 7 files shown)


sql/src/planner.rs
```diff
@@ -564,16 +564,26 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
     }
 
     fn drop_table_to_plan(&self, stmt: DropTable) -> Result<Plan> {
-        let table = if stmt.if_exists {
-            stmt.table_name.to_string()
-        } else {
-            self.find_table(stmt.table_name)?.name().to_string()
-        };
+        let (table_name, partition_info) =
+            if let Some(table) = self.find_option_table(stmt.table_name.clone())? {
```
ZuLiangWang, line 568: The error returned by `find_option_table` has not been handled. Is the exception information lost here?

sql/src/planner.rs
```diff
@@ -564,16 +564,26 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
     }
 
     fn drop_table_to_plan(&self, stmt: DropTable) -> Result<Plan> {
```
ZuLiangWang, line 566: Maybe we should add unit test for this function.

interpreters/src/table_manipulator/meta_based.rs
```diff
@@ -127,3 +127,17 @@ impl TableManipulator for TableManipulatorImpl {
         Ok(Output::AffectedRows(0))
     }
 }
+
+fn convert_to_partition_table_info(
```
Rachelint, line 131: I think passing `None` into the convert function is a bit strange.

sql/src/planner.rs
```diff
@@ -735,6 +745,15 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
             .context(MetaProviderFindTable)?
             .with_context(|| TableNotFound { name: table_name })
     }
+
+    fn find_option_table(&self, table_name: TableName) -> Result<Option<TableRef>> {
```
Rachelint, line 749: Maybe naming it `find_table` is better?

interpreters/src/table_manipulator/meta_based.rs
```diff
@@ -105,17 +101,21 @@ impl TableManipulator for TableManipulatorImpl {
         plan: DropTablePlan,
         _table_engine: TableEngineRef,
     ) -> Result<Output> {
+        let partition_table_info =
```
Rachelint, line 104: create_to_partition_table_info?

sql/src/planner.rs
Rachelint: Should we pass &TableName or &str?
Clone here is strange.

sql/src/planner.rs
chunshao90: Return the error with `?`.

sql/src/planner.rs
chunshao90: `find_table` already exists.

interpreters/src/table_manipulator/meta_based.rs
chunshao90: I rename it to `create_partition_table_info`.

sql/src/planner.rs
chunshao90: `test_drop_statement_to_plan` already covers this `fn`.

sql/src/planner.rs
chunshao90: I remove this function, and modify `find_table`.

Rachelint (approved):
LGTM

sql/src/planner.rs
```diff
@@ -690,7 +714,10 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
     }
 
     fn alter_add_column_to_plan(&self, stmt: AlterAddColumn) -> Result<Plan> {
-        let table = self.find_table(stmt.table_name)?;
+        let table_name = stmt.table_name.to_string();
+        let table = self
+            .find_table(table_name.as_str())?
```
Rachelint, line 719: Unify to `&table_name` ?

sql/src/planner.rs
```diff
@@ -592,7 +606,13 @@ impl<'a, P: MetaProvider> PlannerDelegate<'a, P> {
                 source,
                 ..
             } => {
-                let table = self.find_table(ObjectName(table_name.0).into())?;
+                let table_name_string = TableName::from(table_name).to_string();
```
Rachelint, line 609: Maybe `table_name` is better?

PR #503, opened 2022-12-21
refactor: separate object store from parquet sst async reader

ShiKaiWi:
# Which issue does this PR close?

Prepare for #495 

# Rationale for this change
 In order to control how to decode our custom sst file, the object store must be separated from the Parquet SST async reader because the api of object store has no flexibility to introduce another footer after the original file content.

# What changes are included in this PR?
- Separate the ObjectStore from the parquet sst async reader.

# Are there any user-facing changes?
None.

# How does this change test
Existing tests.

PR #536, opened 2023-01-04
Revert "refactor: separate object store from parquet sst async reader…

ShiKaiWi:
This reverts commit 436d4f076f83046e0d313b602d9939df06033f86. (#503)

Closes #

# Rationale for this change
 After discussion, We find that there is no need to define a custom file format to wrap the real sst file encoded in different kinds of storage format. The real storage format will be told by the manifest module.

Revert #503 

None

# How does this change test
No need.


jiacai2050 (approved):
LGTM

PR #537, opened 2023-01-04
chore: add version in PartitionInfo

chunshao90:

Closes #

# Rationale for this change
 Add version field in `PartitionInfo` for future extension.

* Add version field in `PartitionInfo` for future extension.
* Rename `Definition` to `PartitionDefinition`.

No.

# How does this change test
Old tests have covered.

### DIFF
--- a/analytic_engine/src/meta/details.rs
+++ b/analytic_engine/src/meta/details.rs
@@ -1188,8 +1188,9 @@ mod tests {
 
         runtime.block_on(async move {
             let table_id = ctx.alloc_table_id();
+            let default_version = 0;
             let partition_info = Some(PartitionInfo::Hash(HashPartitionInfo {
-                version: 0,
+                version: default_version,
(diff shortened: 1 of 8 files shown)


analytic_engine/src/meta/details.rs
```diff
@@ -1189,7 +1189,8 @@ mod tests {
         runtime.block_on(async move {
             let table_id = ctx.alloc_table_id();
             let partition_info = Some(PartitionInfo::Hash(HashPartitionInfo {
-                definitions: vec![Definition {
+                version: 0,
```
ZuLiangWang, line 1192: Use `DEFAULT_PARTITION_VERSION` instead of magic number.

chunshao90: Here is the test, I set a variable `default_version`.

Rachelint (approved):
LGTM

PR #538, opened 2023-01-05
refactor: simpify sst meta data in manifest

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, the meta data of sst file in the manifest shares the same information with the one encoded in the sst file. And this is not necessary and may lead to big entry in the manifest log.

Use separate meta data to describe the sst file in the manifest log, excluding the min/max key, and table schema.

None.

# How does this change test
Test by existing unit tests.


jiacai2050 (approved):
LGTM 🚀

PR #539, opened 2023-01-05
feat: add timeout for query

### LINKED ISSUE #165: Add timeout parameter to request
**Description**

For now, there is no way for server to control how long a request will be, although a client can cancel a request at any time, the server thread  will still be running, which wastes server resources.

**Proposal**

Add a timeout parameter for every request, then server can simply reject running when exceed this limit, besides it's best that the server can overwrite this value in case of client's misuse or administration.

**Additional context**

The proposal above don't deal with client disconnect, it's best the server can detect the disconnection, and cancel request afterwards.

jiacai2050:

Closes #165

# Rationale for this change
 
See issue above

This PR only support set timeout in config file, will support set timeout in request level in next PR.

In principle, if all operations are async/await, we can just timeout at the beginning of query, but it's not that ideal in real case, there maybe exists 
- spawned background tasks, similar leak happens in https://github.com/apache/arrow-datafusion/pull/1121
- stream made of iterators, which is a sync operation and cannot be cancelled from outside.

So we have to add timeout detect in those cases beside top-level timeout. AKAIK, there are two places need special handling:
- Memtable scan, include sync operation
- SST scan, include spawned tasks

When implement this feature, I also find TableProvider is wrongly cached, please see comments in 
- https://github.com/CeresDB/ceresdb/pull/539/files#r1063260688
- https://github.com/CeresDB/ceresdb/pull/539/files#r1063262263

No

# How does this change test

Existing UT, and manually test a large table with `select count(1) from cpu;`, it will return following error after 5s timeout
```
$ time curl --location --request POST 'http://0:5446/sql' --data-raw 'select count(1) from cpu;'

{"code":500,"message":"Failed to handle request, err:Query timeout, query:select count(1) from cpu;, err:deadline has elapsed"}
real    0m5.013s
user    0m0.008s
sys     0m0.000s
```

### DIFF
--- a/analytic_engine/src/memtable/skiplist/iter.rs
+++ b/analytic_engine/src/memtable/skiplist/iter.rs
@@ -165,7 +165,7 @@ impl<A: Arena<Stats = BasicStats> + Clone + Sync + Send> ColumnarIterImpl<A> {
 
         if num_rows > 0 {
             if let Some(deadline) = self.deadline {
-                if deadline.saturating_elapsed().is_zero() {
+                if deadline.check_deadline() {
                     return IterTimeout {}.fail();
                 }
             }
--- a/analytic_engine/src/memtable/skiplist/iter.rs
+++ b/analytic_engine/src/memtable/skiplist/iter.rs
@@ -13,7 +13,7 @@ use common_types::{
     schema::Schema,
     SequenceNumber,
 };
-use common_util::codec::row;
+use common_util::{codec::row, time::InstantExt};
 use log::trace;
 use skiplist::{ArenaSlice, IterRef, Skiplist};
 use snafu::ResultExt;
@@ -165,7 +165,7 @@ impl<A: Arena<Stats = BasicStats> + Clone + Sync + Send> ColumnarIterImpl<A> {
 
         if num_rows > 0 {
             if let Some(deadline) = self.deadline {
-                if Instant::now() >= deadline {
+                if deadline.saturating_elapsed().is_zero() {
                     return IterTimeout {}.fail();
                 }
             }
--- a/analytic_engine/src/memtable/skiplist/iter.rs
+++ b/analytic_engine/src/memtable/skiplist/iter.rs
@@ -50,7 +50,7 @@ pub struct ColumnarIterImpl<A: Arena<Stats = BasicStats> + Clone + Sync + Send> 
 
     // Options related:
     batch_size: usize,
-    deadline: Instant,
+    deadline: Option<Instant>,
 
     start_user_key: Bound<Bytes>,
     end_user_key: Bound<Bytes>,
@@ -164,8 +164,10 @@ impl<A: Arena<Stats = BasicStats> + Clone + Sync + Send> ColumnarIterImpl<A> {
         }
 
         if num_rows > 0 {
-            if self.deadline >= Instant::now() {
-                return IterTimeout {}.fail();
+            if let Some(deadline) = self.deadline {
+                if Instant::now() >= deadline {
+                    return IterTimeout {}.fail();
+                }
             }
 
             let batch = builder.build().context(BuildRecordBatch)?;
--- a/analytic_engine/src/memtable/skiplist/iter.rs
+++ b/analytic_engine/src/memtable/skiplist/iter.rs
@@ -2,7 +2,7 @@
 
 //! Skiplist memtable iterator
 
-use std::{cmp::Ordering, iter::Rev, ops::Bound};
+use std::{cmp::Ordering, iter::Rev, ops::Bound, time::Instant};
 
 use arena::{Arena, BasicStats};
 use common_types::{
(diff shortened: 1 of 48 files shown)


sql/src/provider.rs
```diff
@@ -218,27 +223,15 @@ impl<'a, P: MetaProvider> ContextProvider for ContextProviderAdapter<'a, P> {
...
-        if let Some(p) = self.table_cache.borrow().get(name) {
-            return Ok(p);
+        if let Some(table_ref) = self.table_cache.borrow().get(name) {
+            return Ok(self.table_source(table_ref));
         }
 
         // Find in meta provider
         match self.meta_provider.table(name) {
             Ok(Some(table)) => {
-                let table_adapter = Arc::new(TableProviderAdapter::new(
```
jiacai2050, line 228: We shouldn't cache `TableProviderAdapter` directly, since it contains request_id, which is different between different request.

If we cache them, request_id will be same across different requests.

table_engine/src/provider.rs
```diff
@@ -72,14 +75,21 @@ impl TableProviderAdapter {
         limit: Option<usize>,
         read_order: ReadOrder,
     ) -> Result<Arc<dyn ExecutionPlan>> {
+        let request_id = RequestId::from(state.config.config_options().get_u64(CERESDB_REQUEST_ID));
```
jiacai2050, line 78: We can only set user-defined variables via SessionConfig, since `TableProvider` is returned from a trait defined in DataFusion.

sql/src/provider.rs
ShiKaiWi: I find that the request id has been removed from the `TableProviderAdapter`, and if so, the `TableProviderAdapter` seems able to be cached directly?

table_engine/src/table.rs
```diff
@@ -318,13 +319,17 @@ pub struct ReadOptions {
...
+    pub deadline: Instant,
 }
 
 impl Default for ReadOptions {
     fn default() -> Self {
         Self {
             batch_size: 10000,
             read_parallelism: DEFAULT_READ_PARALLELISM,
+            // TODO: read from config
+            deadline: Instant::now() + Duration::from_secs(60),
```
ShiKaiWi, line 332: Maybe `None` is better for the default deadline denoting no timeout.

analytic_engine/src/memtable/skiplist/iter.rs
```diff
@@ -162,6 +164,12 @@ impl<A: Arena<Stats = BasicStats> + Clone + Sync + Send> ColumnarIterImpl<A> {
         }
 
         if num_rows > 0 {
+            if let Some(deadline) = self.deadline {
+                if Instant::now() >= deadline {
+                    return IterTimeout {}.fail();
+                }
+            }
```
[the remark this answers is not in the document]
jiacai2050: Use `deadline.saturating_elapsed().is_zero()` instead.

interpreters/src/context.rs
```diff
@@ -19,14 +19,16 @@ define_result!(Error);
 #[derive(Debug, Clone)]
 pub struct Context {
     request_id: RequestId,
+    deadline: Instant,
```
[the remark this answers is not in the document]
jiacai2050: fixed

proto/protos/remote_engine.proto
```diff
@@ -24,6 +24,8 @@ message TableIdentifier {
 message ReadOptions {
   uint64 batch_size = 1;
   uint64 read_parallelism = 2;
+  // request timeout in millis seconds
+  uint64 timeout = 3;
```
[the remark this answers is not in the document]
jiacai2050: fixed

server/src/mysql/service.rs
```diff
@@ -95,7 +100,7 @@ impl<Q: QueryExecutor + 'static> MysqlService<Q> {
 
                     let rt = runtimes.read_runtime.clone();
                     rt.spawn(AsyncMysqlIntermediary::run_on(
-                        MysqlWorker::new(instance, runtimes),
+                        MysqlWorker::new(instance, runtimes,timeout),
```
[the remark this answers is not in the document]
jiacai2050: fixed

table_engine/src/table.rs
jiacai2050: fixed

proto/protos/remote_engine.proto
```diff
@@ -24,6 +24,7 @@ message TableIdentifier {
 message ReadOptions {
   uint64 batch_size = 1;
   uint64 read_parallelism = 2;
+  common.U64Wrapper timeout_ms = 3;
```
ShiKaiWi, line 27: Maybe it is more elegant to use int64 as the type `timeout_ms` so that we can use a negative timeout to express no deadline.

analytic_engine/src/memtable/skiplist/iter.rs
```diff
@@ -162,6 +164,12 @@ impl<A: Arena<Stats = BasicStats> + Clone + Sync + Send> ColumnarIterImpl<A> {
         }
 
         if num_rows > 0 {
+            if let Some(deadline) = self.deadline {
+                if deadline.saturating_elapsed().is_zero() {
+                    return IterTimeout {}.fail();
+                }
+            }
```
ShiKaiWi, line 171: Maybe it is more readable to define a method called `check_deadline` for this code block.

jiacai2050: Fixed

proto/protos/remote_engine.proto
jiacai2050: Fixed

ShiKaiWi (approved):
LGTM

PR #541, opened 2023-01-06
feat: add permission check for sub tables in table partition

Rachelint:

Closes #

# Rationale for this change
 Common user shouldn't have right to operate the sub tables in table partition directly, and only admin can do this.
So this pr add a permission check for reaching this target. 

Add permission check in interpreter creating stage.

None.

# How does this change test
Test by ut.


interpreters/src/tests.rs
```diff
@@ -252,6 +293,31 @@ where
...
+            .contains("only admin can process sub tables in table partition directly"));
+        let res1 = self.insert_table_and_check("____test_table", false).await;
+        assert!(format!("{:?}", res1)
+            .contains("only admin can process sub tables in table partition directly"));
+        let res2 = self.select_table_and_check("____test_table", false).await;
+        assert!(format!("{:?}", res2)
+            .contains("only admin can process sub tables in table partition directly"));
+
+        // Admin, operations above will success.
+        self.create_table_and_check("____test_table", true)
```
ShiKaiWi, line 311: The test may need updating after #566 merged.

interpreters/src/validator.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub fn validate(&self, plan: &Plan) -> Result<()> {
+        self.validate_admin_permission(plan)?;
+
+        Ok(())
+    }
+
+    fn validate_admin_permission(&self, plan: &Plan) -> Result<()> {
+        // Only admin can operate the sub tables(table partition) directly.
+        if Validator::contains_sub_tables(plan) && !self.ctx.admin {
```
ShiKaiWi, line 34: Maybe we should change the order of the boolean expression because in the most of the cases, admin is `true`:
```
if !self.ctx.admin && Validator::contains_sub_tables(plan)
```

interpreters/src/validator.rs
```diff
@@ -0,0 +1,95 @@
...
+
+            Plan::Show(show_plan) => {
+                if let ShowPlan::ShowCreatePlan(show_create_plan) = show_plan {
+                    is_sub_table!(show_create_plan.table.name())
+                } else {
+                    false
+                }
+            }
+
+            Plan::Exists(_) => false,
```
ShiKaiWi, line 87: Why to avoid the table name check for `Exists` plan?

Rachelint: > Why to avoid the table name check for `Exists` plan?

Not table name in `Exist`...
```
pub struct ExistsTablePlan {
    pub exists: bool,
}
```

table_engine/src/partition/mod.rs
```diff
@@ -367,3 +367,7 @@ pub fn format_sub_partition_table_name(table_name: &str, partition_name: &str) -
         PARTITION_TABLE_PREFIX, table_name, partition_name
     )
 }
+
+pub fn is_sub_partition_table(table_name: &str) -> bool {
+    table_name.starts_with("__")
```
ShiKaiWi, line 372:
```suggestion
    table_name.starts_with("PARTITION_TABLE_PREFIX")
```

ShiKaiWi (approved):
LGTM

PR #544, opened 2023-01-06
support schema header for http/grpc service

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, it is unreasonable to use tenant as the schema when accessing grpc/http service.

- Add new header `x-ceresdb-schema`;
- For compatibility, `tenant` is still used as schema if `schema` is not set.

User now can use `x-ceresdb-schema` header to set schema instead of `x-ceresdb-access-tenant`.

# How does this change test
Existing tests.

### DIFF
--- a/server/src/http.rs
+++ b/server/src/http.rs
@@ -278,23 +278,29 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
         let timeout = self.config.timeout;
 
         header::optional::<String>(consts::CATALOG_HEADER)
+            .and(header::optional::<String>(consts::SCHEMA_HEADER))
             .and(header::optional::<String>(consts::TENANT_HEADER))
-            .and_then(move |catalog: Option<_>, tenant: Option<_>| {
-                // Clone the captured variables
-                let default_catalog = default_catalog.clone();
-                let default_schema = default_schema.clone();
-                let runtime = runtime.clone();
-                async move {
-                    RequestContext::builder()
-                        .catalog(catalog.unwrap_or(default_catalog))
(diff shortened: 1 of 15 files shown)


server/src/http.rs
```diff
@@ -276,22 +276,28 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
...
-            })
+            .and_then(
+                move |catalog: Option<_>, schema: Option<_>, tenant: Option<_>| {
+                    // Clone the captured variables
+                    let default_catalog = default_catalog.clone();
+                    let default_schema = default_schema.clone();
+                    let runtime = runtime.clone();
+                    // FIXME: for compatibility, we may use tenant as the schema if schema is
+                    // missing.
+                    let schema = schema.or(tenant).unwrap_or(default_schema);
```
jiacai2050, line 289: Should gRPC/MySQL service also have this compatible handling?

ShiKaiWi: gRPC service has already handled this. As for MYSQL service, it doesn't include this part.

jiacai2050: I prefer to add a common method to handle this compatibility, also we may need a flag to control this.

ShiKaiWi: Actually, the ways to extract the schema and tenant header value differs from each other for different service implementation. I guess such a common method can't be defined to satisfy the two service (http & gRPC).

jiacai2050: Then how about adding a flag to control this behavior?

ShiKaiWi: OK. I will add such flag.

jiacai2050 (approved):
LGTM

PR #547, opened 2023-01-09
feat: add cached router for remote engine client

Rachelint:

Closes #
Part of #492 

# Rationale for this change
 Now, we do route work in each request, that make performance too bad.
So, this pr add a cached router in remote engine client. 

Add cached router for remote engine client.

None.

# How does this change test
Test manually.

### DIFF
--- a/remote_engine_client/src/cached_router.rs
+++ b/remote_engine_client/src/cached_router.rs
@@ -6,10 +6,10 @@ use std::num::NonZeroUsize;
 
 use ceresdbproto::storage;
 use clru::CLruCache;
+use common_util::partitioned_lock::PartitionedMutex;
 use router::RouterRef;
 use snafu::{OptionExt, ResultExt};
 use table_engine::remote::model::TableIdentifier;
-use tokio::sync::Mutex;
 use tonic::transport::Channel;
 
 use crate::{channel::ChannelPool, config::Config, error::*};
@@ -21,7 +21,7 @@ pub struct CachedRouter {
     /// Cache mapping table to channel of its endpoint
     // TODO: should be replaced with a cache(like "moka")
     // or partition the lock.
-    cache: Mutex<CLruCache<TableIdentifier, Channel>>,
+    cache: PartitionedMutex<CLruCache<TableIdentifier, Channel>>,
 
     /// Channel pool
     channel_pool: ChannelPool,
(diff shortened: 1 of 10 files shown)


remote_engine_client/src/cached_router.rs
```diff
@@ -0,0 +1,110 @@
...
+            // If not found, do real route and put it into cache, and return then.
+            let mut cache = self.cache.lock().await;
+
+            // Double check here, it may have been put by others.
+            let channel_opt = cache.get(table_ident).cloned();
+            if let Some(channel) = channel_opt {
+                channel
+            } else {
+                self.do_route(table_ident).await?
+            }
```
ShiKaiWi, line 61: I guess it is not a good way to process heavy stuff (`do_route`) with the cache lock held. Some duplicate routing requests are tolerant, so just do route without holding the cache lock.

remote_engine_client/src/cached_router.rs
```diff
@@ -0,0 +1,110 @@
...
+        } else {
+            // If not found, do real route and put it into cache, and return then.
+            let mut cache = self.cache.lock().await;
+
+            // Double check here, it may have been put by others.
+            let channel_opt = cache.get(table_ident).cloned();
+            if let Some(channel) = channel_opt {
+                channel
+            } else {
+                self.do_route(table_ident).await?
```
ShiKaiWi, line 60: It seems the new channel has not been updated into the cache?

Rachelint: Fix.

remote_engine_client/src/cached_router.rs
Rachelint: Yes, `duplicate io` or `io with lock` ...
Mainly, I don't know which is more tolerant.
In fact, I think a mechanism similar to `group commit` may be the good way to solve it.

ShiKaiWi (approved):
LGTM

remote_engine_client/src/cached_router.rs
```diff
@@ -0,0 +1,115 @@
...
+            .fail();
+        }
+
+        // Get channel from pool.
+        let endpoint = route_infos
+            .first()
+            .unwrap()
+            .endpoint
+            .clone()
+            .context(RouteNoCause {
```
chunshao90, line 105:
```suggestion
            .with_context(||RouteNoCause {
```

remote_engine_client/src/client.rs
```diff
@@ -57,24 +50,34 @@ impl Client {
             .await
             .context(Rpc {
```
chunshao90, line 51:
```suggestion
            .with_context(||Rpc {
```

remote_engine_client/src/client.rs
```diff
@@ -87,10 +90,21 @@ impl Client {
...
```
chunshao90, line 91:
```suggestion
            .with_context(||Rpc {
```

chunshao90 (approved):
LGTM

PR #549, opened 2023-01-09
Revert "feat: support bloom filter in hybrid format (#479)"

jiacai2050:

Closes #

# Rationale for this change
After #479, sst file size is larger than before, so revert this commit

No

# How does this change test

Existing UT


chunshao90 (approved):
LGTM

PR #550, opened 2023-01-09
feat: disable do_snapshot when recover table data

chunshao90:

Closes #

# Rationale for this change
 Since the read-only partition table and table(follower role) need to be opened in multi servers, the server cannot do snapshot of the manifest when the table is opened.

* Disable `do_snapshot` when recover table data.

No.

# How does this change test
No need.


ShiKaiWi (approved):
LGTM

PR #551, opened 2023-01-09
avoid routing for non-existent table

ShiKaiWi:

Closes #

# Rationale for this change
 Auto picking a random node for non-existing table may leading to confusion to the router's user.

- Remove the logic for auto picking a random node for non-existent table;

Empty routing result will be returned for non-existent table.

# How does this change test
Existing tests.


chunshao90 (approved):
LGTM

Issue #402, opened 2022-11-17
Potential API/storage format breaking change


jiacai2050:
### Describe This Problem

When release 1.0, breaking change is not acceptable, so we need to 
1. Remove bad design 
2. Add compatible method to old format 

This issue is mainly used to collect potential breaking changes

### Proposal

As above

### Additional Context

Some examples I can think of
- #321, we need to record sst file size in manifest;
- #340;
- #495 Define the custom format for SST;

ShiKaiWi:
What I can come up with now includes:
- Define the custom format for SST (now it is totally a parquet file) for the possible brand-new format in the future;
- Refactor the manifest implementation because current one is complicated and may not work well with OBKV;

ShiKaiWi:
One more thing deserved mentioning here, currently the protocol maps the tenant to the schema, which is actually reasonable actually.

jiacai2050:
- rename NumKeyColumns to PrimaryKeyIndexes https://github.com/CeresDB/ceresdb/pull/526

ShiKaiWi:
https://github.com/CeresDB/ceresdb/pull/538

ShiKaiWi:
All the breaking changes we know are all finished.

PR #556, opened 2023-01-10
refactor: make storage format more elegant

ShiKaiWi:

Part of #402 

# Rationale for this change
 Currently, the storage format has been misused for many places. This PR will address this problem.

- Remove `SsType` from the `TableData`;
- Add `StorageFormatHint` in the table options;
- Remove `StorageFormatOptions` from the `SstMetaData`;

`storage_format` has been replaced with `storage_format_hint` in the table options.

# How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/sst/factory.rs
+++ b/analytic_engine/src/sst/factory.rs
@@ -12,7 +12,7 @@ use table_engine::predicate::PredicateRef;
 use crate::{
     sst::{
         builder::SstBuilder,
-        meta_cache::MetaCacheRef,
+        meta_data::cache::MetaCacheRef,
         parquet::{builder::ParquetSstBuilder, AsyncParquetReader, ThreadedReader},
         reader::SstReader,
     },
--- a/analytic_engine/src/sst/factory.rs
+++ b/analytic_engine/src/sst/factory.rs
@@ -16,7 +16,7 @@ use crate::{
         parquet::{builder::ParquetSstBuilder, AsyncParquetReader, ThreadedReader},
         reader::SstReader,
     },
-    table_options::Compression,
+    table_options::{Compression, StorageFormat, StorageFormatHint},
 };
 
 /// Pick suitable object store for different scenes.
@@ -47,6 +47,7 @@ pub trait Factory: Send + Sync + Debug {
         &self,
         options: &SstReaderOptions,
         path: &'a Path,
+        storage_format: StorageFormat,
         store_picker: &'a ObjectStorePickerRef,
     ) -> Option<Box<dyn SstReader + Send + 'a>>;
 
@@ -56,12 +57,6 @@ pub trait Factory: Send + Sync + Debug {
         path: &'a Path,
         store_picker: &'a ObjectStorePickerRef,
     ) -> Option<Box<dyn SstBuilder + Send + 'a>>;
-}
-
-#[derive(Debug, Copy, Clone)]
-pub enum SstType {
-    Auto,
-    Parquet,
 }
 
 /// The frequency of query execution may decide some behavior in the sst reader,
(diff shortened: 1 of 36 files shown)


proto/protos/sst.proto
```diff
@@ -15,6 +15,7 @@ message SstBloomFilter {
   repeated RowGroupFilter row_group_filters = 1;
 }
 
+/// Used by ssts encoded by parquet, incuding columar&hybrid storage formats.
 message SstMetaData {
```
jiacai2050, line 19: How about add one more level abstration:
```
message SstMetaData {
  oneof MetaInfo {
    ParquetMetaData data = 1;
  }
}

message ParquetMetaData {
  // Min key in the sst
  bytes min_key = 1;
  // Max key in the sst
  bytes max_key = 2;
  // Max sequence number in the sst
  uint64 max_sequence = 3;
  // The time range of the sst
  common.TimeRange time_range = 4;
  common.TableSchema schema = 5;
  uint64 size = 6;
  uint64 row_num = 7;
  analytic_common.StorageFormatOptions storage_format_opts = 8;
  SstBloomFilter bloom_filter = 9;
}

```

analytic_engine/src/sst/factory.rs
```diff
@@ -104,11 +99,13 @@ impl Factory for FactoryImpl {
...
         options: &SstReaderOptions,
         path: &'a Path,
+        storage_format: StorageFormat,
         store_picker: &'a ObjectStorePickerRef,
     ) -> Option<Box<dyn SstReader + Send + 'a>> {
         // TODO: Currently, we only have one sst format, and we have to choose right
         // reader for sst according to its real format in the future.
-        let reader = AsyncParquetReader::new(path, store_picker, options);
+        let hybrid_encoding = matches!(storage_format, StorageFormat::Hybrid);
+        let reader = AsyncParquetReader::new(path, hybrid_encoding, store_picker, options);
```
jiacai2050, line 108: why not pass StorageFormat into AsyncParquetReader?

What if we add a third format besides columnar/hybrid, but they are all based on parquet

proto/protos/analytic_common.proto
```diff
@@ -20,17 +20,20 @@ message TableOptions {
...
   Overwrite = 0;
   Append = 1;
 }
 
-message StorageFormatOptions {
-  StorageFormat format = 1;
-  repeated uint32 collapsible_cols_idx = 2;
+message StorageFormatHint {
+  oneof hint {
+    // The value of the auto is meaningless.
```
jiacai2050, line 33:
```
// Auto means the storage format is automatically determined by CeresDB
// it's value have no specific meaning.
```

analytic_engine/src/sst/factory.rs
```diff
@@ -123,13 +120,17 @@ impl Factory for FactoryImpl {
...
-        match options.sst_type {
-            SstType::Parquet | SstType::Auto => Some(Box::new(ParquetSstBuilder::new(
-                path,
-                store_picker,
-                options,
-            ))),
-        }
+        let hybrid_encoding = match options.storage_format_hint {
+            StorageFormatHint::Specific(format) => matches!(format, StorageFormat::Hybrid),
+            StorageFormatHint::Auto => false,
```
jiacai2050, line 125:
```
// Auto is mapped to columnar parquet format now, may change in future.
```

jiacai2050 (changes requested):

analytic_engine/src/sst/factory.rs
ShiKaiWi: `AsyncParquetReader` should not have knowledge about the `StorageForamt`, especially, in the future, some other format will be introduced.

> What if we add a third format besides columnar/hybrid, but they are all based on parquet
As for the third format based on parquet, I guess we should define a different reader rather than reusing the `AsyncParquetReader`.

proto/protos/sst.proto
ShiKaiWi: It sounds great!

analytic_engine/src/sst/factory.rs
jiacai2050: Make sense.

jiacai2050 (approved):
LGTM

PR #562, opened 2023-01-12
refactor: separate object store from async reader

ShiKaiWi:

Prepare for #495

# Rationale for this change
 In order to control how to decode our custom sst file, the object store must be separated from the Parquet SST async reader because the api of object store has no flexibility to introduce another footer after the original file content.

(Actually, this is a similar with #503, which is reverted by #536, and we need to move on to the custom storage format which needs this PR.)

Separate the ObjectStore from the parquet sst async reader.

None.

# How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/sst/file_reader.rs
+++ b/analytic_engine/src/sst/file_reader.rs
@@ -1,0 +1,74 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+use std::{
+    ops::Range,
+    sync::{Arc, RwLock},
+};
+
+use async_trait::async_trait;
+use bytes::Bytes;
+use common_util::error::{GenericError, GenericResult};
+use object_store::{ObjectStoreRef, Path};
+
+pub type AsyncFileReaderRef = Arc<dyn AsyncFileChunkReader>;
+
+#[async_trait]
+pub trait AsyncFileChunkReader: Send + Sync {
+    async fn file_size(&self) -> GenericResult<usize>;
+
+    async fn get_byte_range(&self, range: Range<usize>) -> GenericResult<Bytes>;
+
+    async fn get_byte_ranges(&self, ranges: &[Range<usize>]) -> GenericResult<Vec<Bytes>>;
+}
+
+pub struct FileReaderOnObjectStore {
+    path: Path,
+    store: ObjectStoreRef,
+    cached_file_size: RwLock<Option<usize>>,
+}
+
+impl FileReaderOnObjectStore {
+    pub fn new(path: Path, store: ObjectStoreRef) -> Self {
+        Self {
+            path,
+            store,
+            cached_file_size: RwLock::new(None),
+        }
+    }
+}
+
+#[async_trait]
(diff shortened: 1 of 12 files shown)


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -51,12 +56,67 @@ use crate::sst::{
 
 type SendableRecordBatchStream = Pin<Box<dyn Stream<Item = Result<ArrowRecordBatch>> + Send>>;
 
+pub type AsyncFileReaderRef = Arc<dyn AsyncFileReader>;
+
+#[async_trait]
+pub trait AsyncFileReader: Send + Sync {
```
jiacai2050, line 62: This trait seems unnecessary, I can't see why this is required.

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -51,12 +56,67 @@ use crate::sst::{
...
+
+    async fn get_byte_range(&self, range: Range<usize>) -> GenericResult<Bytes>;
+
+    async fn get_byte_ranges(&self, ranges: &[Range<usize>]) -> GenericResult<Vec<Bytes>>;
+}
+
+/// Fetch and parse [`ParquetMetadata`] from the file reader.
+///
+/// Referring to: https://github.com/apache/arrow-datafusion/blob/ac2e5d15e5452e83c835d793a95335e87bf35569/datafusion/core/src/datasource/file_format/parquet.rs#L390-L449
+async fn fetch_parquet_metadata_from_file_reader(
```
jiacai2050, line 73: How about move this function to parquet_ext component?

jiacai2050 (changes requested):

analytic_engine/src/sst/parquet/async_reader.rs
ShiKaiWi: It sounds great.

analytic_engine/src/sst/parquet/async_reader.rs
ShiKaiWi: Actually, it is. This is an alternative for `ObjectStore` in the async_reader module.

analytic_engine/src/sst/file_reader.rs
```diff
@@ -0,0 +1,75 @@
...
+    async fn file_size(&self) -> GenericResult<usize> {
+        // check cached filed_size first
+        {
+            let file_size = self.cached_file_size.read().unwrap();
+            if let Some(s) = file_size.as_ref() {
+                return Ok(*s);
+            }
+        }
+
+        // fetch the size from the underlying store
```
jiacai2050, line 52: double check here?

jiacai2050 (approved):
LGTM

analytic_engine/src/sst/file_reader.rs
ShiKaiWi: A double check requires head fetch from oss with the lock held, which may lead to poor performance.

PR #565, opened 2023-01-12
Refactor separate object store from sst builder

ShiKaiWi:

Part of #495 

# Rationale for this change
 In order to introduce our custom file format, `ObjectStore` should be avoided being used by `SstBuilder` and a `FileChunkWriter` will replace it so that we can append a footer containing the file format information to the encoded sst.

- Replace the `ObjectStore` with `FileChunkWriter` in the SST builder.

None.

# How does this change test
Existing tests.


ShiKaiWi:
A new design has been proposed, so let's drop this changeset.

PR #566, opened 2023-01-12
chore: change the prefix of sub partition table name to "__"

chunshao90:

Closes #

# Rationale for this change
 At present, the prefix of  sub partition table name is "____",  which is too long.

Change the prefix of sub partition table name to "__".

No.

# How does this change test
No need.


ShiKaiWi (approved):
LGTM

PR #567, opened 2023-01-13
chore: revert #562

ShiKaiWi:

Part of #495 

# Rationale for this change
 We decide to keep the sst file readable by third-party tools in the parquet ecosystem, that is to say, our custom format should not append any footer or header to the sst file (encoded in parquet format).

Finally, we decide to keep the same header/footer length (4Byte)  of any other new sst format as the one of parquet format, so the `FileChunkReader` is useless.

- Revert #562 

# How does this change test


jiacai2050 (approved):
LGTM

PR #570, opened 2023-01-13
feat: detect sst header

### LINKED ISSUE #495: Impose a custom format of SST

ShiKaiWi:

Closes #495

# Rationale for this change
 In order to introduce our custom storage format, we choose to put the 4B header into the sst file, which is uniform with the parquet file format.

- Add `SstReadHint` to avoid some other extra io;
- Add `HeaderParser` for supporting detect file format;
- According to the header, decide which reader to use.

None.

# How does this change test
Existing.

### DIFF
--- a/analytic_engine/src/sst/factory.rs
+++ b/analytic_engine/src/sst/factory.rs
@@ -4,11 +4,14 @@
 
 use std::{fmt::Debug, sync::Arc};
 
+use async_trait::async_trait;
 use common_types::projected_schema::ProjectedSchema;
 use common_util::runtime::Runtime;
+use log::error;
 use object_store::{ObjectStoreRef, Path};
 use table_engine::predicate::PredicateRef;
 
+use super::header::HeaderParser;
 use crate::{
     sst::{
         builder::SstBuilder,
@@ -42,8 +45,9 @@ impl ObjectStorePicker for ObjectStoreRef {
     }
 }
 
+#[async_trait]
 pub trait Factory: Send + Sync + Debug {
-    fn new_sst_reader<'a>(
+    async fn new_sst_reader<'a>(
         &self,
         options: &SstReaderOptions,
         path: &'a Path,
@@ -51,7 +55,7 @@ pub trait Factory: Send + Sync + Debug {
         store_picker: &'a ObjectStorePickerRef,
     ) -> Option<Box<dyn SstReader + Send + 'a>>;
 
-    fn new_sst_builder<'a>(
+    async fn new_sst_builder<'a>(
         &self,
         options: &SstBuilderOptions,
         path: &'a Path,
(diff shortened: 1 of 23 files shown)


analytic_engine/src/sst/factory.rs
```diff
@@ -42,21 +54,25 @@ impl ObjectStorePicker for ObjectStoreRef {
...
         path: &'a Path,
-        storage_format: StorageFormat,
+        options: &SstReadOptions,
+        hint: SstReadHint,
         store_picker: &'a ObjectStorePickerRef,
-    ) -> Option<Box<dyn SstReader + Send + 'a>>;
+    ) -> Result<Box<dyn SstReader + Send + 'a>>;
 
-    fn new_sst_builder<'a>(
+    async fn create_builder<'a>(
```
jiacai2050, line 70:
```
create_writer
```

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -54,9 +55,10 @@ type SendableRecordBatchStream = Pin<Box<dyn Stream<Item = Result<ArrowRecordBat
 pub struct Reader<'a> {
     /// The path where the data is persisted.
     path: &'a Path,
-    hybrid_encoding: bool,
     /// The storage where the data is persist.
     store: &'a ObjectStoreRef,
+    /// The hint for the sst file size. One io will be avoided if provided.
```
jiacai2050, line 60:
```
One io will be avoided if provided.
```
It seems this comment is not suitable here, a better place could be in `ReadHint`

jiacai2050 (approved):
LGTM

PR #576, opened 2023-01-16
refactor: avoid meta data for write sst

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, `SstMetaData` is used for `SstWriter`, but the writer should have no knowledge about some fields in the `SstMetaData`, e.g. `bloom_filter`. To fix this, `MetaData` is defined in the `sst::writer` to replace the `SstMetaData`.

- Replace `SstMetaData` with `MetaData` in `SstWriter`;
- Rename `sst::builder` to `sst::writer`;

None.

# How does this change test
Existing tests.


jiacai2050 (approved):
LGTM

PR #578, opened 2023-01-17
feat: implement prom remote storage api

### LINKED ISSUE #577: Implement prometheus remote storage API
### Describe This Problem

Prometheus is a popular choice for monitoring, and it supports long-term storage via remote storage API. 

CeresDB can gain more popularity by integrating with Prometheus.

### Proposal

Implement remote storage API in CeresDB 

### Additional Context

- https://prometheus.io/docs/prometheus/latest/storage/#remote-storage-integrations
- https://crates.io/crates/prom-remote-api

jiacai2050:

Closes #577

# Rationale for this change
 

Add remote storage API support in HTTP module.

Yes, add two HTTP API, prometheus users can use them by add following config to their prometheus.yml

```yml
remote_write:
  - url: "http://127.0.0.1:5000/prom/v1/write"

remote_read:
  - url: "http://127.0.0.1:5000/prom/v1/read"
```

# How does this change test

Manually

### DIFF
--- a/server/src/handlers/prom.rs
+++ b/server/src/handlers/prom.rs
@@ -1,0 +1,35 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+use async_trait::async_trait;
+use prom_remote_api::types::{ReadRequest, ReadResponse, RemoteStorage, Result, WriteRequest};
+use query_engine::executor::Executor as QueryExecutor;
+
+use crate::instance::InstanceRef;
+
+pub struct CeresDBStorage<Q: QueryExecutor + 'static> {
+    #[allow(dead_code)]
+    instance: InstanceRef<Q>,
+}
+
+impl<Q: QueryExecutor + 'static> CeresDBStorage<Q> {
+    pub fn new(instance: InstanceRef<Q>) -> Self {
+        Self { instance }
+    }
+}
+
+#[async_trait]
+impl<Q: QueryExecutor + 'static> RemoteStorage for CeresDBStorage<Q> {
+    /// Write samples to remote storage
+    async fn write(&self, req: WriteRequest) -> Result<()> {
+        println!("mock write, req:{req:?}");
+
+        Ok(())
+    }
+
+    /// Read samples from remote storage
+    async fn read(&self, req: ReadRequest) -> Result<ReadResponse> {
+        println!("mock query, req:{req:?}");
+
+        Ok(ReadResponse::default())
+    }
+}
--- a/server/src/handlers/prom.rs
+++ b/server/src/handlers/prom.rs
@@ -1,10 +1,27 @@
 // Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
 
 use async_trait::async_trait;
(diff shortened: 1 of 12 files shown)


server/src/handlers/prom.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+}
+
+impl<Q: QueryExecutor + 'static> CeresDBStorage<Q> {
+    /// This function will separate measurement from labels, and sort labels by
+    /// name.
+    fn normalize_labels(labels: Vec<Label>) -> Result<(String, Vec<Label>)> {
+        let mut new_labels = Vec::with_capacity(labels.len());
+        let mut measurement = None;
```
ShiKaiWi, line 132: In the prometheus context, `metric` seems better than `measurement` which is usually used by influxdb.

ShiKaiWi, line 132: In order to extract the metric from the `labels`, `swap_move` may be more efficient than creating a new vector.

server/src/handlers/prom.rs
```diff
@@ -0,0 +1,427 @@
...
+        let measurement = measurement.context(MissingName)?;
+        new_labels.sort_unstable_by(|a, b| a.name.cmp(&b.name));
+
+        Ok((measurement, new_labels))
+    }
+
+    fn convert_write_request(req: WriteRequest) -> Result<WriteRequestPb> {
+        let mut req_by_metric = HashMap::new();
+        for timeseries in req.timeseries {
+            let (measurement, labels) = Self::normalize_labels(timeseries.labels)?;
```
ShiKaiWi, line 150: Maybe we should normalize all `measurement` to `metric`.

server/src/handlers/prom.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        let measurement = measurement.context(MissingName).unwrap();
+        let sql = format!(
+            "select * from {} where {} order by {}, {}",
+            measurement,
+            filters.join(" and "),
+            TSID_COLUMN,
+            TIMESTAMP_COLUMN
+        );
```
ShiKaiWi, line 294: Actually, I guess a separate method can be defined for this conversion and afterwards we can add some unit tests for it if you will.

server/src/handlers/prom.rs
```diff
@@ -0,0 +1,427 @@
...
+            .await
+            .map_err(Box::new)
+            .context(SqlHandle)?;
+
+        Self::convert_query_result(measurement, result)
+    }
+}
+
+/// Converter convert Arrow's RecordBatch into Prometheus's QueryResult
+struct Converter {
```
ShiKaiWi, line 306: Shall we add some unit tests for this converter?

jiacai2050: Yeah, I will add some tests for basic case.

server/src/handlers/prom.rs
jiacai2050: OK

jiacai2050: 👍
Fixed, this function looks more simple and elegant now.

ShiKaiWi (approved):
LGTM

PR #580, opened 2023-01-17
feat: improve the release profile and add release-slim profile

ShiKaiWi:

Closes #

# Rationale for this change
Current release profile is not the best concerning performance, and some options can be configured for better performance, but with slow compiling time and high memory consumption.

- Add `lto=true` and `codegen-units=1` to `release` profile;
- Remove `better-release` profile, and add `release-slim` profile for a smaller and slower binary;
- Fix the bug that `cargo-sort` doesn't work in the precommit;

None.

# How does this change test
Current ci.


Rachelint (approved):
LGTM

PR #591, opened 2023-01-29
chore: add codecov.yml to ignore some dirs that don't need code coverage

chunshao90:

Closes #

# Rationale for this change
 Exclude some directories that don't need code coverage.

Ignore following dirs in codecov:
  - "benchmarks"
  - "integration_test"
  - "meta_client"
  - "remote_engine_client"
  - "wal/src/message_queue_impl"
  - "wal/src/table_kv_impl"

No.

# How does this change test
No need.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,8 @@
+ignore:
+  - "**/error.rs"
+  - "benchmarks"
+  - "integration_test"
+  - "meta_client"
+  - "remote_engine_client"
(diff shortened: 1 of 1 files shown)


.github/codecov.yml
```diff
@@ -0,0 +1,8 @@
+ignore:
+  - "**/error.rs"
```
ShiKaiWi, line 2: It's weird to ignore such files.

ShiKaiWi (approved):
LGTM

PR #598, opened 2023-01-31
feat restrict partition num

Rachelint:

Closes #

# Rationale for this change
Now partition num in table partition is not restricted,  you can set the partition num to an unaccpetable huge number. 

In this pr we make a restriction on it, it cannot be greater than MAX_PARTITION_NUM(1024 now).

User can not set a partition num greater than MAX_PARTITION_NUM while creating partitioned table.

# How does this change test
Test by ut.


chunshao90 (approved):
LGTM

PR #607, opened 2023-02-02
feat: store manifest to oss

### LINKED ISSUE #579: Store snapshot of manifest to object store
### Describe This Problem

Currently, the snapshot and normal updates of manifest are appended to WAL, which leads to such problems:
- Complexity of processing the concurrent writing to WAL of snapshotting and normal updating;
- One giant snapshot may exceed the limit of the length of one WAL entry;

### Proposal

The basic idea of the proposal is simple: store the snapshot to the object store.

### Additional Context

This is a breaking change #402.

ShiKaiWi:

Closes #579 

# Rationale for this change
 Details have been provided in the #579. The basic targes include:
- Avoid any large log entry written to the WAL, which may lead to unrecoverable write/query failures;
- Simplify the snapshot procedure with the help of ObjectStore.

Implement the proposal: https://github.com/CeresDB/ceresdb/issues/579#issuecomment-1413207521

None.

# How does this change test
Existing tests.


analytic_engine/src/manifest/details.rs
```diff
@@ -291,86 +323,215 @@ impl Manifest for ManifestImpl {
...
         let snapshotter = Snapshotter {
             location,
-            log_store: RegionWal::new(location, self.wal_manager.clone()),
-            options: self.opts.clone(),
+            log_store,
+            snapshot_store,
         };
         let snapshot = snapshotter.create_latest_snapshot().await?;
-        Ok(snapshot.data)
+        println!("snapshot is {:?}", snapshot);
```
Rachelint, line 366: Remove it.

jiacai2050:
So many changes, could you point me which parts are most important to review?

ShiKaiWi:
> So many changes, could you point me which parts are most important to review?

Actually, the massive changes result from the file path renaming: `meta/details.rs` -> `manifest/details.rs`. I will revert the renaming for a clean change set.

analytic_engine/src/meta/details.rs
```diff
@@ -107,20 +152,20 @@ define_result!(Error);
 
 #[async_trait]
 trait MetaUpdateLogEntryIterator {
```
Rachelint, line 154: It seems we wanted to rename it to `MetaUpdateReader` but forgot?

analytic_engine/src/meta/details.rs
```diff
@@ -291,86 +327,231 @@ impl Manifest for ManifestImpl {
...
+        );
+
+        if load_req.do_snapshot {
+            if let Some(snapshot) = self
+                .maybe_do_snapshot(load_req.space_id, load_req.table_id, location, true)
+                .await?
+            {
                 return Ok(snapshot.data);
             }
         }
```
Rachelint, line 369: Maybe, we remove it?
I think `load_data` should be a read-only operation?

analytic_engine/src/meta/details.rs
```diff
@@ -291,86 +327,231 @@ impl Manifest for ManifestImpl {
...
 #[async_trait]
 trait MetaUpdateLogStore: std::fmt::Debug {
     type Iter: MetaUpdateLogEntryIterator;
-    async fn scan(
-        &self,
-        ctx: &ScanContext,
-        start: ReadBoundary,
-        end: ReadBoundary,
-    ) -> Result<Self::Iter>;
+    async fn scan(&self, start: ReadBoundary, end: ReadBoundary) -> Result<Self::Iter>;
```
Rachelint, line 394: Is end necessary?
As I see,  scan usually only need to define the start point. Maybe we can remove end now?

analytic_engine/src/meta/details.rs
```diff
@@ -291,86 +327,231 @@ impl Manifest for ManifestImpl {
...
+            } else {
+                debug!("Succeed in purging old snapshot:{}", old_snapshot_path);
+            }
         }
+
+        Ok(())
     }
+
+    /// Load the `current_snapshot` file from the underlying store, and with the
+    /// mapping info in it load the latest snapshot file then.
```
Rachelint, line 519: Mapping info

analytic_engine/src/meta/details.rs
```diff
@@ -291,86 +327,231 @@ impl Manifest for ManifestImpl {
...
+        // Update the current snapshot.
+        let current_snapshot = meta_update::CurrentSnapshot {
+            sequence: snapshot.end_seq,
+        };
+        let payload = current_snapshot.encode_to_vec();
+        let path = self.current_snapshot_path();
+        self.store
+            .put(&path, payload.into())
+            .await
+            .context(UpdateCurrentSnapshot)?;
```
Rachelint, line 500: Maybe add comment to decribe the corruption risk?

analytic_engine/src/meta/details.rs
ShiKaiWi: Actually, this name is still suitable even if the return value is changed from `MetaUpdateLogEntry` to `MetaUpdate`, because the iterator returns not only the `MetaUpdate` but also `SequenceNumber` which indicates that this is a log.

analytic_engine/src/meta/details.rs
ShiKaiWi: done

analytic_engine/src/meta/details.rs
ShiKaiWi: Done.

analytic_engine/src/meta/details.rs
ShiKaiWi: I guess it should be `mapping info` rather than `Mapping info` here.

analytic_engine/src/meta/details.rs
ShiKaiWi: Done.

analytic_engine/src/meta/details.rs
```diff
@@ -217,61 +235,79 @@ pub struct ManifestImpl {
...
-            .await
-            .context(WriteWal)
+        log_store.append(meta_update).await
     }
 
     /// Do snapshot if no other snapshot is triggered.
     ///
     /// Returns the latest snapshot if snapshot is done.
-    async fn maybe_do_snapshot(&self, location: WalLocation) -> Result<Option<Snapshot>> {
+    async fn maybe_do_snapshot(
```
jiacai2050, line 270: I find return value of this function is never used, remove it in this PR?

analytic_engine/src/meta/mod.rs
```diff
@@ -9,9 +9,20 @@ pub mod meta_update;
...
+
+use crate::{
+    meta::{meta_data::TableManifestData, meta_update::MetaUpdateRequest},
+    space::SpaceId,
+};
+
+pub struct LoadRequest {
+    pub space_id: SpaceId,
+    pub table_id: TableId,
+    pub cluster_version: u64,
```
jiacai2050, line 23: this name is not consistent with WalLocation

```
pub struct WalLocation {
    pub versioned_region_id: VersionedRegionId,
    pub table_id: TableId,
}

```

analytic_engine/src/meta/details.rs
ShiKaiWi: It is used in the L316.

analytic_engine/src/meta/mod.rs
ShiKaiWi: Maybe we should keep the original name of the field, because `Manifest` trait doesn't expose anything relevant with WAL.

analytic_engine/src/meta/details.rs
```diff
@@ -1317,73 +1250,59 @@ mod tests {
 
     #[derive(Debug)]
     struct MemLogStore {
```
chunshao90, line 1252: Extract `MemLogStore` into a separate file.

ShiKaiWi: This belongs to the test module, so no need to set it in a separate file.

jiacai2050 (approved):
LGTM

PR #609, opened 2023-02-03
chore: remove lto

jiacai2050:

Closes #

# Rationale for this change
 
When lto is true, release build would fail on machine with 16G(9G free), so remove it.

# How does this change test


ShiKaiWi (approved):
LGTM

PR #611, opened 2023-02-06
feat: prevent to push down filter of non primary key or ts key

### LINKED ISSUE #605: Query stale data in overwritten mode when filter match the old one
### Describe this problem

Two records with a same primary key may exist in two ssts even in overwritten mode(we will do the dedup work while querying according to primary key).

In the following case, the overwritten logic mentioned above will fail. Assume that we have record definition with two fields in it : key(primary key), and is_deleted(a normal field, not tag, not primary key). And two record with the same key are flushed into two different sst in overwritten mode. 

Obviously the new one should overwrite the old one in the right logic, and we should never get the old one while querying. But we can still get the old one in theory, and that happened actually.
```

  ┌────────────────────────┐
  │        Storage         │
  │                        │            ┌─────────────────────────┐
  │   ┌─────────────────┐  │            │                         │
  │   │     SST1        │  │            │                         │
  │   │                 │  │            │                         │
  │   │  ┌────┬───────┐ │  │            │    Filter:              │
  │   │  │ pk │ false │ │  │            │   ┌────────────────┐    │
  │   │  └────┴───────┘ │  ◄────────────┼───┤ deleted = false│    │
  │   │                 │  │            │   └────────────────┘    │
  │   └─────────────────┘  │            │                         │
  │                        │            │                         │
  │   ┌─────────────────┐  │            │    Result:              │
  │   │      SST2       │  │            │   ┌────┬───────┐        │
  │   │                 │  ├────────────┼───► pk │ false │        │
  │   │  ┌────┬──────┐  │  │            │   └────┴───────┘        │
  │   │  │ pk │ true │  │  │            │                         │
  │   │  └────┴──────┘  │  │            │                         │
  │   │                 │  │            │                         │
  │   └─────────────────┘  │            │                         │
  │                        │            └─────────────────────────┘
  └────────────────────────┘
```   

### Steps to reproduce

I am trying to reproduce it in a simple way.

### Expected behavior

_No response_

### Additional Information

_No response_

Rachelint:

Closes #605 

# Rationale for this change
 See #605 

Check if the column in push down filter is primary key or ts key, if not prevent to pass such filter to table scan.

None.

# How does this change test
Test by ut.


table_engine/src/provider.rs
```diff
@@ -122,12 +124,61 @@ impl TableProviderAdapter {
...
+                }
+            }
+        } else {
+            // When tsid does not exist, that means user defined primary key is used.
+            // So, only filters of primary key can be pushed down.
+            for primary_idx in self.read_schema.primary_key_indexes() {
+                let primary_column = self.read_schema.column(*primary_idx);
+                key_set.insert(&primary_column.name);
+            }
+        }
```
ShiKaiWi, line 148: I guess we can encapsulate the block code into a method called `collect_unique_keys` on the type of `read_schema`.

table_engine/src/provider.rs
```diff
@@ -122,12 +124,61 @@ impl TableProviderAdapter {
...
+            // Once found a column not primary key or timestamp key in `filter`,
+            // the `filter` will be marked as unable to push down.
+            if !key_set.contains(column.as_str()) {
+                should_push_down = false;
+                break;
+            }
+        }
+
+        should_push_down
+    }
```
ShiKaiWi, line 181:
```suggestion
    fn only_filter_unique_key_columns(filter: &Expr, key_set: &HashSet<&str>) -> bool {
        let filter_cols = visitor::find_columns_by_expr(filter);
        for filter_col in filter_cols {
            // If a column which is not part of the unique key occured in `filter`, the `filter` shouldn't be pushed down.
            if !key_set.contains(filter_col.as_str()) {
                return false;
            }
        }

        true
    }
```

table_engine/src/provider.rs
```diff
@@ -122,12 +124,61 @@ impl TableProviderAdapter {
         Ok(Arc::new(scan_table))
     }
 
-    fn predicate_from_filters(&self, filters: &[Expr]) -> PredicateRef {
+    fn check_and_build_predicate_from_filters(&self, filters: &[Expr]) -> PredicateRef {
+        // 1.Only filters of primary key or timestamp key can be pushed down.
```
ShiKaiWi, line 128:
```suggestion
        // Only filters on the columns belonging to the unique key can be pushed down.
```

table_engine/src/provider.rs
```diff
@@ -329,3 +380,133 @@ impl fmt::Debug for ScanTable {
             .finish()
     }
 }
+
+#[cfg(test)]
```
jiacai2050, line 384: Better to add a integration test to reproduce example in origin issue.

common_types/src/schema.rs
```diff
@@ -855,6 +855,33 @@ impl Schema {
         }
     }
 
+    pub fn unique_keys(&self) -> Vec<&str> {
+        // Only filters on the columns belonging to the unique key can be pushed down.
+        // Build key set.
+        let mut keys = Vec::new();
+        // Timestamp key should be always picked.
+        keys.push(self.timestamp_name());
+        if self.tsid_column().is_some() {
```
jiacai2050, line 864: tsid column should also included in this case.

jiacai2050: How about this logic
```
if let Some(tsid) = self.tsid_column() {
  return self.columns.iter().filter(|is_tag| ...).collect().extend(vec![tsid, timestamp]);
}

self.primary_key_indexes().iter().map(|idx| self.column[idx]).collect()
```

table_engine/src/provider.rs
```diff
@@ -122,12 +124,41 @@ impl TableProviderAdapter {
         Ok(Arc::new(scan_table))
     }
 
-    fn predicate_from_filters(&self, filters: &[Expr]) -> PredicateRef {
+    fn check_and_build_predicate_from_filters(&self, filters: &[Expr]) -> PredicateRef {
+        let unique_keys = self
+            .read_schema
+            .unique_keys()
+            .into_iter()
+            .collect::<HashSet<_>>();
```
jiacai2050, line 132: IMO Vec is enough, short length vector is very fast when loop.

jiacai2050 (approved):
LGTM

PR #613, opened 2023-02-06
feat: print error in datanode

Rachelint:

Closes #

# Rationale for this change
 Now error occured while handling meta's request isn't printed in datanode, that make trouble in debugging.

Print error datanode before returning to meta.

None.

# How does this change test
Test manually.

### DIFF
--- a/server/src/grpc/meta_event_service/mod.rs
+++ b/server/src/grpc/meta_event_service/mod.rs
@@ -24,7 +24,7 @@ use ceresdbproto::meta_event::{
 use cluster::ClusterRef;
 use common_types::schema::SchemaEncoder;
 use common_util::{runtime::Runtime, time::InstantExt};
-use log::info;
+use log::{error, info};
 use paste::paste;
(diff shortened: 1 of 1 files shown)


server/src/grpc/meta_event_service/mod.rs
```diff
@@ -74,7 +74,10 @@ macro_rules! handle_request {
 
                 let res = handle
                     .await
-                    .map_err(|e| Box::new(e) as _)
+                    .map_err(|e| {
+                        error!("Fail to process request from meta, err:{}", e);
```
jiacai2050, line 78: I think we should add this log in L92, it has more condition check `Ok(Err(e)) | Err(e)`

Rachelint: Yes, the outer is runtime's error, mainly what we wanted is the error in the inner result...

ShiKaiWi (approved):
LGTM

PR #614, opened 2023-02-06
chore: add logs

jiacai2050:

Closes #

# Rationale for this change
 
When troubleshooting in local my host, I found those logs are useful.

No

# How does this change test

This is what looks now
```
2023-02-06 16:07:24.514 INFO [server/src/http.rs:462] HTTP server starts listening on 127.0.0.1:5440
2023-02-06 16:07:24.515 INFO [server/src/server.rs:114] Server start, open local tables
2023-02-06 16:07:24.515 INFO [server/src/server.rs:128] Server start, create default schema if not exist
2023-02-06 16:07:24.516 INFO [server/src/server.rs:131] Server start, start services
2023-02-06 16:07:24.516 INFO [server/src/mysql/service.rs:54] MySQL server starts listening on 127.0.0.1:3307
2023-02-06 16:07:24.516 INFO [server/src/server.rs:138] Server start finished
2023-02-06 16:07:24.516 INFO [server/src/grpc/mod.rs:168] Grpc server starts listening on 127.0.0.1:8831
2023-02-06 16:07:24.517 INFO [server/src/grpc/mod.rs:177] Grpc server serves remote engine rpc service
```

### DIFF
--- a/components/message_queue/src/kafka/kafka_impl.rs
+++ b/components/message_queue/src/kafka/kafka_impl.rs
@@ -200,7 +200,7 @@ impl MessageQueue for KafkaImpl {
 
         // Create topic in Kafka.
         let topic_management_config = &self.config.topic_management_config;
-        info!("Try to create topic:{} in kafka", topic_name);
+        info!("Try to create topic, name:{}.", topic_name);
         let result = self
             .controller_client
             .create_topic(
@@ -211,6 +211,7 @@ impl MessageQueue for KafkaImpl {
             )
             .await;
 
+        info!("Create topic finish, name:{}", topic_name);
         match result {
             // Race condition between check and creation action, that's OK.
             Ok(_)
--- a/components/message_queue/src/kafka/kafka_impl.rs
+++ b/components/message_queue/src/kafka/kafka_impl.rs
@@ -211,7 +211,10 @@ impl MessageQueue for KafkaImpl {
             )
             .await;
 
-        info!("Create topic finish, name:{}", topic_name);
+        info!(
+            "Create topic finish, name:{}, result:{:?}",
+            topic_name, result
+        );
         match result {
             // Race condition between check and creation action, that's OK.
             Ok(_)
--- a/server/src/http.rs
+++ b/server/src/http.rs
@@ -460,7 +460,7 @@ impl<Q: QueryExecutor + 'static> Builder<Q> {
         };
 
         info!(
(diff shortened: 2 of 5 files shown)


server/src/mysql/service.rs
```diff
@@ -51,14 +51,14 @@ impl<Q: QueryExecutor + 'static> MysqlService<Q> {
 
         let rt = self.runtimes.clone();
         self.tx = Some(tx);
+        info!("MySQL server starts listening on {}", self.socket_addr);
```
Rachelint, line 54: Why move it to here?

server/src/http.rs
```diff
@@ -459,6 +459,11 @@ impl<Q: QueryExecutor + 'static> Builder<Q> {
             enable_tenant_as_schema: self.enable_tenant_as_schema,
         };
 
+        info!(
+            "HTTP server starts listening on {}",
+            &self.config.endpoint.to_string()
+        );
```
Rachelint, line 465: How about printing this after spawning?

components/message_queue/src/kafka/kafka_impl.rs
```diff
@@ -211,6 +211,7 @@ impl MessageQueue for KafkaImpl {
             )
             .await;
 
+        info!("Create topic finish, name:{}", topic_name);
```
Rachelint, line 214: I think the result should be printed in this log.

server/src/http.rs
ShiKaiWi: I guess you won't know which port will be used by the http service, so how about updating this log to:
```suggestion
        info!(
            "HTTP server tries to listen on {}",
            &self.config.endpoint.to_string()
        );
```

And add one more log after it succeeds in binding:
```
        info!(
            "HTTP server listens on {}",
            &self.config.endpoint.to_string()
        );
```

server/src/mysql/service.rs
ShiKaiWi: Maybe we should keep both the logs.

server/src/http.rs
jiacai2050: The log is verbose, I prefer to only add the first one.

jiacai2050: I add panic logic when listen failed for grpc and mysql, http server already does this.

ShiKaiWi (approved):
LGTM

PR #615, opened 2023-02-06
feat: make bloom filter optional on column level

ShiKaiWi:

Closes #

# Rationale for this change
Currently, the bloom filter is only optional on file level, and the bloom filters on column level are still required. However, in some cases, a column bloom filter may be useless, which can be avoided after this PR.

And for compatibility, the version of bloom filter should be introduced.

- Introduce the version of bloom filter. [Breaking Change]
- Make bloom filter optional on column level.

None.

# How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/sst/parquet/meta_data.rs
+++ b/analytic_engine/src/sst/parquet/meta_data.rs
@@ -60,10 +60,6 @@ impl RowGroupBloomFilter {
         }
     }
 
-    pub fn push_column_filter(&mut self, column_filter: Option<Bloom>) {
-        self.column_filters.push(column_filter);
-    }
-
     /// Accrue the data belonging to one column.
     ///
     /// Caller should ensure the `column_idx` is in the range.
@@ -84,9 +80,7 @@ impl RowGroupBloomFilter {
 
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct BloomFilter {
-    // Every filter is a row group filter consists of column filters.
-    //
-    // The row group filter can be None if the row group is not indexed.
+    /// Every filter is a row group filter consists of column filters.
     row_group_filters: Vec<RowGroupBloomFilter>,
 }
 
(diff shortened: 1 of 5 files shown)


analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -38,68 +45,125 @@ pub enum Error {
...
+            .row_group_filters
             .iter()
             .map(|row_group_filter| {
                 let column_filters = row_group_filter
+                    .column_filters
                     .iter()
-                    .map(|column_filter| column_filter.data().to_vec())
+                    .map(|column_filter| {
+                        column_filter
+                            .map(|v| v.data().to_vec())
```
jiacai2050, line 114: I think we need to add a test for this special handling, 
> first encode to pb, and decode it back

proto/protos/sst.proto
```diff
@@ -12,7 +12,8 @@ message SstBloomFilter {
     repeated bytes column_filters = 1;
   };
 
-  repeated RowGroupFilter row_group_filters = 1;
+  uint32 version = 1;
```
jiacai2050, line 15: 👍 Version seems goods, we may change bits used for one filter in future.

analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -38,68 +45,125 @@ pub enum Error {
...
+    }
+
+    pub fn push_column_filter(&mut self, column_filter: Option<Bloom>) {
+        self.column_filters.push(column_filter);
+    }
+
+    /// Accrue the data belonging to one column.
+    ///
+    /// Caller should ensure the `column_idx` is in the range.
+    pub fn accrue_column_data(&mut self, column_idx: usize, data: &[u8]) {
```
jiacai2050, line 70: How about impl this via IndexMut trait?

analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -38,68 +45,125 @@ pub enum Error {
...
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct BloomFilter {
-    // Two level vector means
-    // 1. row group
-    // 2. column
-    filters: Vec<Vec<Bloom>>,
+    // Every filter is a row group filter consists of column filters.
```
jiacai2050, line 87: `///`

analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -38,68 +45,125 @@ pub enum Error {
...
+    //
+    // The row group filter can be None if the row group is not indexed.
```
jiacai2050, line 89: The column of row group can be None?

analytic_engine/src/sst/parquet/meta_data.rs
ShiKaiWi: If so, I guess we have to expose the underlying `ethbloom::Bloom` outside, which seems a little ugly.

analytic_engine/src/sst/parquet/meta_data.rs
ShiKaiWi: Oops. Forgot to remove this comment after the codes are updated.

analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -38,68 +45,125 @@ pub enum Error {
 
 define_result!(Error);
 
+const DEFAULT_BLOOM_FILTER_VERSION: u32 = 0;
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct RowGroupBloomFilter {
+    // The column filter can be None if the column is not indexed.
+    column_filters: Vec<Option<Bloom>>,
```
jiacai2050, line 53: I think another issue for using col_idx to refer column, it will block us to implement  some DDL, such as add column after existing one.

analytic_engine/src/sst/parquet/meta_data.rs
jiacai2050: Didn't you already leak it in push_column_filter above ?

analytic_engine/src/sst/parquet/meta_data.rs
jiacai2050: I will open an issue to track this.

jiacai2050: https://github.com/CeresDB/ceresdb/issues/616

analytic_engine/src/sst/parquet/meta_data.rs
ShiKaiWi: Oops. I forget to remove that method.

jiacai2050 (approved):
LGTM

PR #621, opened 2023-02-08
primary keys defaults to tsid, timestamp

jiacai2050:

Closes #

# Rationale for this change
 
I find this bug when write docs https://github.com/CeresDB/docs/pull/25

Primary key already is `tsid, timestamp` by default in SQL, update grpc to follow this.

No 
# How does this change test

UT


Rachelint (approved):
LGTM

PR #601, opened 2023-01-31
feat: Upgrade to Datafusion 17

chunshao90:
# Which issue does this PR close?

Closes #

# Rationale for this change
 Upgrade to datafusion 17 and arrow 31.

# What changes are included in this PR?
Jobs left:
* #617 
* #622 

Changelog of those two packages:

https://github.com/apache/arrow-datafusion/blob/master/datafusion/CHANGELOG.md#1700-2023-01-27
https://github.com/apache/arrow-rs/blob/master/CHANGELOG-old.md#3100-2023-01-13

# Are there any user-facing changes?
No.

# How does this change test

PR #624, opened 2023-02-09
fix: avoid some unwrap with result

chunshao90:

Closes #

# Rationale for this change
 In #601 , some `Result`s are handled directly using `unwrap`, which is too rough.

Return error instead of unwrap.

No.

# How does this change test
Old tests are covered.


jiacai2050 (approved):
LGTM

PR #632, opened 2023-02-09
chore: run ci in parallel

jiacai2050:

Closes #

# Rationale for this change
 
To make CI run faster

- style-check, unit test, integration test have no relation, which means they can run in parallel.

image

No

# How does this change test
CI itself.

- https://github.com/CeresDB/ceresdb/actions/runs/4135058171

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,8 @@ env:
   LOCK_FILE: Cargo.lock
 
 jobs:
-  linux-ci:
-    name: linux-ci
+  style-check:
+    name: style-check
     runs-on: ubuntu-latest
     timeout-minutes: 60
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -59,17 +59,14 @@ jobs:
           restore-keys: |
             debug-${{ runner.os }}-${{ hashFiles('rust-toolchain') }}-
             debug-${{ runner.os }}-
-            debug-${{ runner.os }}
+            debug-
       - name: Ensure Disk Quota
```
Rachelint, line 63: Why only ensuring disk quota in style check?

jiacai2050: Cache is shared among jobs, so we only need to check disk in one job.

Rachelint (approved):
LGTM

PR #640, opened 2023-02-13
chore: refacor ceresdb config

chunshao90:

Closes #

# Rationale for this change
 It is strange that `DeployMode` is placed in the top-level configuration, it needs to be placed in a sub-configuration.

The `service` section in config is renamed to `server`.

The `service` section in config is renamed to `server`.

# How does this change test
Manual testing.


jiacai2050 (approved):
LGTM

PR #641, opened 2023-02-14
fix: sql identifier default to case-sensitive

### LINKED ISSUE #442: Identifiers should be case-sensitive for all protocols by default

jiacai2050:

Closes #442

# Rationale for this change
 

- https://github.com/apache/arrow-datafusion/blob/master/datafusion/CHANGELOG.md
  - Add option to determine whether to convert identifiers #4551
- Bump rust to nightly 2023-02-08, (required by one of arrow deps)

Main code changes are:
- Build `SqlToRel` with `enable_ident_normalization: false` option
- Build `TableReference` from `ResolvedTableReference`

Other changes are clippy fix.

No

# How does this change test

integration tests


Makefile
```diff
@@ -54,7 +54,9 @@ check-license:
 	cd $(DIR); sh scripts/check-license.sh
 
 clippy:
-	cd $(DIR); cargo clippy --all-targets --all-features --workspace -- -D warnings
```
jiacai2050, line 57: https://github.com/CeresDB/ceresdb/issues/647
Fix in next PR

ShiKaiWi (approved):
LGTM

PR #642, opened 2023-02-14
feat: allow mapping between sequence and kafka offset incomplete

Rachelint:

Closes #

# Rationale for this change
Now I assume that in recovery, no flush of specific table will be triggered on the sequence which is not the marked deleted one among the meta snapshot. So mapping between sequence and kafka offset is not stored in the meta snapshot.

However, this assumption is unreasonable. Fortunately, it is easy to process this situation by just not updating the marked deleted sequence in this flush.

Allow the mapping above incomplete.

None.

# How does this change test
Test by ut.


ShiKaiWi (approved):
LGTM

PR #643, opened 2023-02-14
fix: correct the order of sync meta snapshot and clean logs in wal on kafka

Rachelint:

Closes #

# Rationale for this change
Now, the order of sync meta snapshot and clean logs in wal onk kafka is unreasonable.

We must ensure that the meta snapshot has been persisted successfully before cleaning the related logs. 

See title.

None.

# How does this change test
Test by ut.


ShiKaiWi (approved):
LGTM

PR #645, opened 2023-02-14
feat: map shard id to region id in manifest

Rachelint:

Closes #

# Rationale for this change
Before, manifest is only stored in wal storage, so region id can olny be mapped to table id(otherwise, manifest of specific table will lose after being moved to new shard).

But now, if can mao shard id to region id again because we store the manifest snapshot in oss. We can force it to do snapshot before closing to solve the problem above.

See title.

None.

# How does this change test
Test by ut.


analytic_engine/src/manifest/details.rs
```diff
@@ -293,13 +298,13 @@ impl Manifest for ManifestImpl {
     async fn store_update(
         &self,
         request: MetaUpdateRequest,
-    ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync>> {
-        debug!("try to store update:{:?}", request);
+    ) -> std::result::Result<(), GenericError> {
+        info!("Manifest store update, request:{:?}", request);
```
ShiKaiWi, line 302: The info level may lead to massive logs if the number of tables is large.

analytic_engine/src/manifest/details.rs
```diff
@@ -342,8 +347,28 @@ impl Manifest for ManifestImpl {
             snapshot_store,
         };
         let snapshot = snapshotter.create_latest_snapshot().await?;
+
         Ok(snapshot.and_then(|v| v.data))
     }
+
+    async fn do_snapshot(&self, request: SnapshotRequest) -> std::result::Result<(), GenericError> {
```
ShiKaiWi, line 354:
```suggestion
    async fn do_snapshot(&self, request: SnapshotRequest) -> GenericResult<()> {
```

analytic_engine/src/manifest/details.rs
```diff
@@ -317,11 +322,11 @@ impl Manifest for ManifestImpl {
     async fn load_data(
         &self,
         load_req: &LoadRequest,
-    ) -> std::result::Result<Option<TableManifestData>, Box<dyn std::error::Error + Send + Sync>>
-    {
+    ) -> std::result::Result<Option<TableManifestData>, GenericError> {
```
ShiKaiWi, line 325:
```suggestion
    ) -> GenericResult<Option<TableManifestData>> {
```

analytic_engine/src/manifest/mod.rs
```diff
@@ -10,21 +10,22 @@ use std::{fmt, sync::Arc};
...
 use common_types::table::ShardId;
+use common_util::error::GenericError;
 use table_engine::table::TableId;
 
 use crate::{
     manifest::{meta_data::TableManifestData, meta_update::MetaUpdateRequest},
     space::SpaceId,
 };
-
+#[derive(Debug)]
```
ShiKaiWi, line 20: Insert a newline  above.

ShiKaiWi (approved):
LGTM

PR #649, opened 2023-02-15
refactor: adjust configuration

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, the config of the CeresDB is casual, which needs some adjustment for elegance.

- Move the top-level config from server module to the root module;
- Rename all the `path` config to `data_dir`;
- Replace the `Deployment` with the `Option<ClusterDeployment>`;
- Rename some configurations;
- Add sub sections `tracing` and `logger` under the top level;

Something irrelevant to configurations:
- Remove crate `serde_derive` which is actually re-imported by `serde` crate.

Config file needs updating after this PR.

# How does this change test
Existing tests.


src/config.rs
```diff
@@ -0,0 +1,99 @@
...
+            zone: "".to_string(),
+            idc: "".to_string(),
+            binary_version: "".to_string(),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct Config {
```
jiacai2050, line 34:
```
#[serde(deny_unknown_fields)]
```
I prefer to add this to prevent misconfigure.

src/config.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(default)]
+pub struct RuntimeConfig {
+    // Runtime for reading data
+    pub read_thread_num: usize,
+    // Runtime for writing data
+    pub write_thread_num: usize,
+    // Runtime for communicating with meta cluster
+    pub meta_thread_num: usize,
```
jiacai2050, line 85: `Option<usize>` since meta is not a must.

jiacai2050: Or maybe we can move this field into ClusterDeployment?

ShiKaiWi: I guess it makes code complex if the `meta_runtime` is optional, and such idle `meta_runtime` causes no performance problems.

components/logger/src/lib.rs
```diff
@@ -77,7 +77,7 @@ pub fn file_drainer(path: &Option<String>) -> Option<CeresFormat<PlainDecorator<
     }
 }
 
-// dispacher
+/// Dispatcher for log.
```
Rachelint, line 80: Dispatcher for logs.

analytic_engine/src/lib.rs
```diff
@@ -226,24 +225,24 @@ impl From<WalNamespaceConfig> for NamespaceConfig {
...
+    pub kafka: KafkaConfig,
 
-    /// Wal config
-    pub wal_config: MessageQueueWalConfig,
+    /// Common config for message queue config.
+    pub message_queue: MessageQueueWalConfig,
 }
 
-/// Config for wal based on rocksDB
+/// Config for wal based on RocksDB.
```
Rachelint, line 234: It isn't a sentence, how about remove "."?

Rachelint (approved):
LGTM

Issue #616, opened 2023-02-07
Use column id instead of index to refer column when persist to storage


jiacai2050:
### Describe This Problem

When persist schema info, we use column index to refer corresponding column, this works fine if schema is not updated, and will block us to implement some DDL, such as add column after another.

### Proposal

Use column id to refer column when need to persist to underlying storage.

### Additional Context

AFAIK, primary keys/bloom filters in encoded using index now.

ShiKaiWi:
Assuming that **schema info** mentioned here is the one in the SST, I guess this is not necessary, because the existing sst won't be updated even if the schema is updated.

jiacai2050:
I grep TableSchema, schema is not only used in SST, but also in meta_update.

```
7 candidates:
./common.proto:46:message TableSchema {
./common.proto:67:  common.TableSchema table_schema = 1;
./meta_update.proto:43:  common.TableSchema schema = 4;
./meta_update.proto:98:  common.TableSchema schema = 3;
./remote_engine.proto:64:  common.TableSchema table_schema = 1;
./sst.proto:34:  common.TableSchema schema = 5;
./table_requests.proto:14:  common.TableSchema schema = 2;
```

PR #652, opened 2023-02-16
feat: replace column index with id when persisting schema in manifest

Rachelint:

Part of #616 

# Rationale for this change
See #616 

See title.

None.

# How does this change test
Test by ut.


ShiKaiWi (approved):
LGTM

PR #655, opened 2023-02-16
refactor: use enum instead of version for encoding format in pb

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, `version` number has been introduced to record the encoding format in the protobuf, and this is very messy compared with `enum` (`one_of` in the protobuf).

Adapt to  https://github.com/CeresDB/ceresdbproto/pull/51

None.

# How does this change test
Existing tests.


Rachelint (approved):
LGTM

PR #656, opened 2023-02-16
refactor: reuse client in manifest wal and normal wal

Rachelint:

Closes #

# Rationale for this change
Now, manifest wal and normal wal use respective kafka client, it may be unreasonable.
 

+ Reuse client in them.
+ Modify config.

None.

# How does this change test
Maybe not need for this pr.


ShiKaiWi (approved):
LGTM

PR #659, opened 2023-02-22
chore: normalize http api

jiacai2050:

Closes #

# Rationale for this change
 
Make
- All internal apis begin with `debug`
- Admin related apis begin with `admin`

Yes

# How does this change test

No need

### DIFF
--- a/server/src/http.rs
+++ b/server/src/http.rs
@@ -157,6 +157,7 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
             .and(write_api.or(query_api))
     }
 
+    // GET /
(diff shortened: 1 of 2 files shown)


server/src/http.rs
```diff
@@ -195,10 +196,11 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
             })
     }
 
+    // POST /debug/flush_memtable
```
ShiKaiWi, line 199: Maybe this url should be prefixed with `admin`.

jiacai2050: I think this interface is mainly used for debugging, not for operation purpose.

ShiKaiWi (approved):
LGTM

PR #660, opened 2023-02-22
feat: remove avro encoder

chunshao90:

Closes #

# Rationale for this change
 Remove avro encoding, use arrow-ipc. 
* Reduce dependencies.
* The performance of arrow-ipc  will be better.

Remove avro encoding, use arrow-ipc. 

No.

# How does this change test
Manual testing.


Cargo.toml
```diff
@@ -116,8 +116,8 @@ message_queue = { path = "components/message_queue" }
 zstd = { version = "0.12", default-features = false }
 
 [workspace.dependencies.ceresdbproto]
-git = "https://github.com/CeresDB/ceresdbproto.git"
-rev = "4dbd2b36262affd83f8a2d4a99273e43aebcb067"
+git = "https://github.com/chunshao90/ceresdbproto.git"
```
ShiKaiWi, line 119: Use the correct url.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -161,10 +164,12 @@ impl<Q: QueryExecutor + 'static> RemoteEngineService for RemoteEngineServiceImpl
...
                                 ..Default::default()
                             },
-                            Ok(CompressOutput { payload, .. }) => ReadResponse {
+                            Ok(CompressOutput { payload, method }) => ReadResponse {
                                 header: Some(build_ok_header()),
-                                version: RemoteEngineVersion::ArrowIPCWithZstd.as_u32(),
-                                rows: payload,
+                                output: Some(Arrow(ArrowPayload {
+                                    record_batches: vec![payload],
+                                    compression: method as i32,
```
ShiKaiWi, line 171: The type of `method` is not the same as the type of `compression` field, and maybe we need an explicit conversion between them.

table_engine/src/remote/model.rs
```diff
@@ -148,22 +149,33 @@ impl TryFrom<ceresdbproto::remote_engine::WriteRequest> for WriteRequest {
...
-                version: row_group_pb.version,
+        let rows = row_group_pb.rows.context(EmptyRowGroup)?;
+        let row_group = match rows {
+            Arrow(v) => {
+                ensure!(
+                    !v.record_batches.is_empty(),
+                    InvalidRecordBatchNumber {
+                        batch_num: v.record_batches.len(),
+                    }
+                );
```
ShiKaiWi, line 160: - Why can't the `record_batches` be empty?
- The error message for the `InvalidRecordBatchNumber` says `Invalid record batches number in the response, expect only one`, which seems wrong with the exact meaning here.

table_engine/src/remote/model.rs
```diff
@@ -202,3 +240,48 @@ impl TryFrom<WriteRequest> for ceresdbproto::remote_engine::WriteRequest {
         })
     }
 }
+
+fn build_row_group_with_record_batch(
```
ShiKaiWi, line 244:
```suggestion
fn build_row_group_from_record_batch(
```

table_engine/src/remote/model.rs
```diff
@@ -202,3 +240,48 @@ impl TryFrom<WriteRequest> for ceresdbproto::remote_engine::WriteRequest {
...
+    record_batches: Vec<arrow::record_batch::RecordBatch>,
+) -> Result<RowGroup> {
+    ensure!(
+        record_batches.len() == 1,
+        InvalidRecordBatchNumber {
+            batch_num: record_batches.len(),
+        }
+    );
```
ShiKaiWi, line 252: Just take one record batch as input parameter if only one batch is allowed in the `record_batches`.

Cargo.toml
```diff
@@ -116,8 +116,8 @@ message_queue = { path = "components/message_queue" }
 zstd = { version = "0.12", default-features = false }
 
 [workspace.dependencies.ceresdbproto]
-git = "https://github.com/CeresDB/ceresdbproto.git"
-rev = "4dbd2b36262affd83f8a2d4a99273e43aebcb067"
```
jiacai2050, line 120: Does 0.1.0 contains changes you want?
- https://crates.io/crates/ceresdbproto

benchmarks/src/bin/ipc.rs
```diff
@@ -45,23 +45,4 @@ pub fn main() {
         let cost = begin.saturating_elapsed();
         println!("Arrow IPC encode/decode cost:{}ms", cost.as_millis());
     }
-
```
jiacai2050, line 48: Since avro is not used any more, I prefer to delete this benchmark file

table_engine/src/remote/model.rs
```diff
@@ -148,22 +149,33 @@ impl TryFrom<ceresdbproto::remote_engine::WriteRequest> for WriteRequest {
...
+            Arrow(v) => {
+                ensure!(
+                    !v.record_batches.is_empty(),
+                    InvalidRecordBatchNumber {
+                        batch_num: v.record_batches.len(),
+                    }
+                );
+
+                let compression = match v.compression() {
+                    arrow_payload::Compression::None => CompressionMethod::None,
```
jiacai2050, line 163: Add a from impl?

jiacai2050 (approved):
LGTM

PR #661, opened 2023-02-23
refactor: remove encoder method from WalManager

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, the default implementation of `encoder` method in `WalManager` trait is useless because callers can use the `LogBatchEncoder` directly to create such encoder.

Remove the `encode` method defined in `WalManager`.

None.

# How does this change test
Existing tests.


Rachelint (approved):
LGTM

Issue #589, opened 2023-01-28
More advanced techniques to read parquet files efficiently


jiacai2050:
### Describe This Problem

Usually the IO part of a query is the most time consuming, so reducing time spent on this would improve query latency quietly a lot.

In current implementation, we have already applies some tricks to optimize this, to name a few:
1. concurrent reads even for one file
2. min/max prune
3. custom bloom filter prune

There is an [awesome blog](https://arrow.apache.org/blog/2022/12/26/querying-parquet-with-millisecond-latency/) written by @tustvold and @alamb introducing some more advanced techniques to further improve read speed, which is definitely a must-read for developer in Arrow ecosystem.

### Proposal

Explore ideas introduced in [Querying Parquet with Millisecond Latency](https://arrow.apache.org/blog/2022/12/26/querying-parquet-with-millisecond-latency/), Some notable ideas are:
- Page prune
- Late materialization

PR #664, opened 2023-02-24
feat: add parquet page filter

jiacai2050:
## Rationale
Part of #589

## Detailed Changes
- Introduce `PagePruningPredicate` when build `ParquetRecordBatchStream`


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -182,6 +195,36 @@ impl<'a> Reader<'a> {
...
+        let partition = 0;
+        let metrics = ParquetFileMetrics::new(partition, self.path.as_ref(), &self.df_plan_metrics);
+        let exprs = datafusion::optimizer::utils::conjunction(self.predicate.exprs().to_vec());
+        let exprs = match exprs {
+            Some(exprs) => exprs,
+            None => return Ok(None),
+        };
+
+        let df_schema = arrow_schema
+            .clone()
```
ShiKaiWi, line 214: Can this clone be avoided?

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -182,6 +195,36 @@ impl<'a> Reader<'a> {
...
 
+    fn build_row_selection(
+        &self,
+        arrow_schema: SchemaRef,
+        row_groups: &[usize],
+        file_metadata: &parquet_ext::ParquetMetaData,
+    ) -> Result<Option<RowSelection>> {
+        // TODO: remove fixed partition
+        let partition = 0;
+        let metrics = ParquetFileMetrics::new(partition, self.path.as_ref(), &self.df_plan_metrics);
```
ShiKaiWi, line 206: Move this to L222.

analytic_engine/src/sst/parquet/async_reader.rs
jiacai2050: I'm afraid we cannot, this method is in datafusion. But It's an Arc object, so I think it's acceptable.

ShiKaiWi (approved):
LGTM

PR #631, opened 2023-02-09
feat: replace bloom with xor8 filter

jiacai2050:
# Which issue does this PR close?

Closes #510

# Rationale for this change
 
See issue above

# What changes are included in this PR?

- Replace bloom with xor8 filter
- Add `Filter` trait to represent different filters

# Are there any user-facing changes?

No

# How does this change test

Old UT

PR #666, opened 2023-02-27
fix: xor filter build

jiacai2050:

Closes #

# Rationale for this change
 
In #631 xor filter is introduced, however there is one bug on that PR, filter is not built correctly.

- Fix xor filter build
- Add string type testcase in min-max prune

No

# How does this change test

Newly added UT.


ShiKaiWi (approved):
LGTM

PR #668, opened 2023-02-27
test: add partition_table test in integration_test

chunshao90:

Closes #

# Rationale for this change
 Add partition_table test in `integration_test`.

Add partition_table test in `integration_test`.

No.

# How does this change test

### DIFF
--- a/integration_tests/cases/env/cluster/05_ddl/partition_table.result
+++ b/integration_tests/cases/env/cluster/05_ddl/partition_table.result
@@ -16,6 +16,16 @@ CREATE TABLE `partition_table_t`(
 
 affected_rows: 0
 
+SHOW CREATE TABLE partition_table_t;
+
+Table,Create Table,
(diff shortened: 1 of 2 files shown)


integration_tests/cases/env/cluster/05_ddl/partition_table.result
```diff
@@ -0,0 +1,78 @@
+DROP TABLE IF EXISTS `partition_table_t`;
+
+affected_rows: 0
+
+DROP TABLE IF EXISTS `__partition_table_t_0`;
```
ZuLiangWang, line 5: Add read/write case for sub table like this.

integration_tests/cases/env/cluster/05_ddl/partition_table.sql
```diff
@@ -0,0 +1,36 @@
...
+SELECT * from partition_table_t where name = "ceresdb1";
+
+SELECT * from partition_table_t where name = "ceresdb2";
+
+SELECT * from partition_table_t where name = "ceresdb3";
+
+SELECT * from partition_table_t where name in ("ceresdb0", "ceresdb1", "ceresdb2", "ceresdb3", "ceresdb4") order by name;;
+
+SELECT * from partition_table_t where name in ("ceresdb5", "ceresdb6", "ceresdb7","ceresdb8", "ceresdb9", "ceresdb10") order by name;;
+
```
ZuLiangWang, line 36: Maybe we should add show tables case and show create table case for partition table.

ZuLiangWang (approved):
LGTM

Issue #663, opened 2023-02-23
Support shard based table recovering


Rachelint:
### Describe This Problem

We have supported shard based reading and writing in wal, the further target is to support shard based table recovering.

### Proposal

It may be a big work, and it should be split into targets:
+ Implement shard based manifest recovering.
+ Implement shard based tables recovering (problem is how we support single table's recovering after refactoring?)
+ Maybe refactor the wal module?

Rachelint:
Deprecated.

PR #669, opened 2023-02-27
refactor: remove cluster version

ShiKaiWi:

Part of #663

# Rationale for this change
 Remove the useless cluster version in the WAL location, which is introduced by #442 to solve the problems mentioned in #441.

- Introduce the `close_region` in the `WalManager` trait;
- Solve the #441 by doing `close_region` instead of using different cluster version; 
- Refactor the setup procedure to obtain the different `Wal`s in the top level structs;

None.

# How does this change test
Existing tests.

### DIFF
--- a/server/src/grpc/meta_event_service/shard_operation.rs
+++ b/server/src/grpc/meta_event_service/shard_operation.rs
@@ -1,0 +1,40 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+// Copyright 2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+//! Implementation of operations on shards.
+
+// TODO: Currently, only a specific operation (close wal region) is implemented,
+// and it is expected to encapsulate more operations on **Shard** in the future.
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use common_types::table::ShardId;
+use common_util::error::{BoxError, GenericResult};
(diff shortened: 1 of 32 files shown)


server/src/grpc/meta_event_service/shard_operation.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use common_types::table::ShardId;
+use common_util::error::{BoxError, GenericResult};
+use wal::manager::WalManagerRef;
+
+#[async_trait]
+pub trait WalRegionCloser: std::fmt::Debug + Send + Sync {
```
Rachelint, line 18: Why not define a `ShardOperator`?
And we place `open_tables_on_shard` and `close_tables_on_shard` into it?

Rachelint: I think if all actions about opening and closing should be placed together.
Otherwise, it's easy to forget to modify some of them in later?

ShiKaiWi: Actually, `ShardOperator` needs more refactoring, which includes not only operation about wal region but also some other operations about the engine. And I will do this in the following PRs.

Rachelint: Ok.

Rachelint (approved):
LGTM

PR #678, opened 2023-03-01
fix: can't recognize aliyun oss NotFoundError when load manifest

ShiKaiWi:

Closes #

# Rationale for this change
 Currently, the not found error won't be thrown when access an object file in the aliyun oss by oss sdk, leading to possible panic when restarting.

Take a workaround way (check whether the error message contains '404' or 'not found') to detect the not found error.

None.

# How does this change test
Manual test.

### DIFF
--- a/analytic_engine/src/manifest/details.rs
+++ b/analytic_engine/src/manifest/details.rs
@@ -435,7 +435,7 @@ impl MetaUpdateSnapshotStore for ObjectStoreBasedSnapshotStore {
         // is not thrown as [object_store::ObjectStoreError::NotFound].
         if let Err(err) = &get_res {
             let err_msg = err.to_string().to_lowercase();
-            if err_msg.contains("404") | err_msg.contains("not found") {
+            if err_msg.contains("404") || err_msg.contains("not found") {
                 warn!("Current snapshot file doesn't exist, err:{}", err);
                 return Ok(None);
             }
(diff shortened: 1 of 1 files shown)


analytic_engine/src/manifest/details.rs
```diff
@@ -431,6 +431,16 @@ impl MetaUpdateSnapshotStore for ObjectStoreBasedSnapshotStore {
             return Ok(None);
         };
 
+        // TODO: currently, this is just a workaround to handle the case where the error
+        // is not thrown as [object_store::ObjectStoreError::NotFound].
+        if let Err(err) = &get_res {
+            let err_msg = err.to_string().to_lowercase();
+            if err_msg.contains("404") | err_msg.contains("not found") {
```
MachaelLee, line 438:
```suggestion
            if err_msg.contains("404") || err_msg.contains("not found") {
```

analytic_engine/src/manifest/details.rs
```diff
@@ -431,6 +431,16 @@ impl MetaUpdateSnapshotStore for ObjectStoreBasedSnapshotStore {
             return Ok(None);
         };
 
+        // TODO: currently, this is just a workaround to handle the case where the error
```
Rachelint, line 434: How about wrap  decision about not found into a function?

ShiKaiWi: This way is just a temporary workaround, and it will be removed in recent future.

jiacai2050 (approved):
LGTM

PR #679, opened 2023-03-01
fix: remove auto create table option

jiacai2050:

Closes #

# Rationale for this change
 
Auto create table is an important feature, so we decide to support this always. 

- Remove `auto_create_tables` options

# How does this change test

CI


ShiKaiWi (approved):
LGTM

Issue #682, opened 2023-03-01
Tracking issue: InfluxQL support


ShiKaiWi:
### Describe This Problem

InfluxQL is widely used in the timeseries DBMS field, and CeresDB should support such a protocol for integration into the ecosystem.

### Proposal

The work includes three parts:
- Write: the parser of line protocol;
- Query: 
  - The parser for InfluxQL;
  - The planner for InfluxQL;

Rachelint:
Works of query part may can be divided into three part:
+ Support statements compatible with sql statmens (may be need some rewrites)
   + Wildcard and regrex in select fields, from clause, group by clause.
   + From multiple tables.
+ Support statements can be represent by `udf`
   + Selector aggregate function.
   + Some special udf.
+ Support statemetns only can be represent by `extension logical node`
   + Some special clause, such as slimit clause, soffset clause, fill clause.

ShiKaiWi:
In favor of #698

PR #689, opened 2023-03-02
fix: move grpc create table to sql crates

jiacai2050:

Part of #682, prepare for influxdb line protocol write.

# Rationale for this change
 
Create table plan from grpc is in grpc crate, which is wrong.

All plan creation logic should be placed in sql crates.

Note: sql may not a good name, I plan to rename this crate to `planner` in future.

Move create plan in grpc to sql crates.

No

# How does this change test

CI


ShiKaiWi (approved):
LGTM

PR #692, opened 2023-03-02
fix:  Parse string without specify time zone to local time stamp

MachaelLee:

Closes #

# Rationale for this change

I found the following query response nothing :
```
curl -H POST 'http://127.0.0.1:5440/sql' -H 'Content-Type: application/json' -d '{
    "query": "INSERT INTO demo(`t`, name,value, dat,`time`) VALUES(1667374200022, '\''ceresdb05'\'', 1.0, '\''9999-10-13'\'','\''-23:59:59.999'\'')"
}'

curl -H POST 'http://127.0.0.1:5000/sql' -H 'Content-Type: application/json' -d '{
    "query": "select * from demo where t > '\''2023-03-02 15:40:00'\''"
}'

```

Actually it should response:

```
{"rows":[{"tsid":7907694565448534541,"t":1677743400000,"name":"ceresdb01","value":1.0}]}
```

And I found the reason is , according to  https://github.com/apache/arrow-rs/pull/2814, function `string_to_timestamp_nanos` returns a UTC timestamp instead of  local timestamp;

1. When no time zone is specified,  parse it to local time stamp
2. Otherwise, use `string_to_timestamp_nanos` to parse it to utc time stamp

# How does this change test

By unit test

### DIFF
--- a/query_engine/src/logical_optimizer/type_conversion.rs
+++ b/query_engine/src/logical_optimizer/type_conversion.rs
@@ -2,7 +2,8 @@
 
 use std::{mem, sync::Arc};
 
-use arrow::{compute, compute::kernels::cast_utils::string_to_timestamp_nanos};
+use arrow::{compute, compute::kernels::cast_utils::string_to_timestamp_nanos, error::ArrowError};
+use chrono::{Local, LocalResult, NaiveDateTime, TimeZone, Utc};
 use datafusion::{
     arrow::datatypes::DataType,
     common::DFSchemaRef,
@@ -186,7 +187,10 @@ impl<'a> TypeRewriter<'a> {
     fn cast_scalar_value(value: &ScalarValue, data_type: &DataType) -> Result<ScalarValue> {
         if let DataType::Timestamp(_, _) = data_type {
             if let ScalarValue::Utf8(Some(v)) = value {
-                return string_to_timestamp_ms(v);
+                return match string_to_timestamp_ms_workaround(v) {
+                    Ok(v) => Ok(v),
+                    _ => string_to_timestamp_ms(v),
+                };
             }
         }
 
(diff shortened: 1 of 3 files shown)


query_engine/src/logical_optimizer/type_conversion.rs
```diff
@@ -281,16 +282,59 @@ impl<'a> ExprRewriter for TypeRewriter<'a> {
...
+
+    // without a timezone specifier as a local time, using 'T' as a separator
+    // Example: 2020-09-08T13:42:29.190855
+    if let Ok(ts) = NaiveDateTime::parse_from_str(string, "%Y-%m-%dT%H:%M:%S%.f") {
+        let mills = naive_datetime_to_timestamp(string, ts).map_err(DataFusionError::from)?;
+        return Ok(ScalarValue::TimestampMillisecond(Some(mills), None));
+    }
+
+    // without a timezone specifier as a local time, using ' ' as a separator
+    // Example: 2020-09-08T13:42:29.190855
```
chunshao90, line 298:
```suggestion
    // Example: 2020-09-08 13:42:29.190855
```

query_engine/src/logical_optimizer/type_conversion.rs
```diff
@@ -281,16 +282,59 @@ impl<'a> ExprRewriter for TypeRewriter<'a> {
 }
 
 fn string_to_timestamp_ms(string: &str) -> Result<ScalarValue> {
+    // TODO(lee): remove following codes after PR(https://github.com/apache/arrow-rs/pull/3787) merged
+    // Because function `string_to_timestamp_nanos` returns a NaiveDateTime's
+    // nanoseconds from a string without a specify time zone, We need to convert
+    // it to local timestamp.
+
+    // without a timezone specifier as a local time, using 'T' as a separator
```
chunshao90, line 290: When no time zone is specified, the local time zone is used.

query_engine/src/logical_optimizer/type_conversion.rs
ShiKaiWi: It seems there is no difference between this example and the one on L291. And I wonder what is the meaning of `T`?

MachaelLee: i have fixed this comment

query_engine/src/logical_optimizer/type_conversion.rs
```diff
@@ -281,16 +282,59 @@ impl<'a> ExprRewriter for TypeRewriter<'a> {
...
+    }
+
+    // without a timezone specifier as a local time, using ' ' as a separator
+    // Example: 2020-09-08 13:42:29.190855
+    if let Ok(ts) = NaiveDateTime::parse_from_str(string, "%Y-%m-%d %H:%M:%S%.f") {
+        let mills = naive_datetime_to_timestamp(string, ts).map_err(DataFusionError::from)?;
+        return Ok(ScalarValue::TimestampMillisecond(Some(mills), None));
+    }
+
+    let result = string_to_timestamp_nanos(string);
```
ShiKaiWi, line 304:
```suggestion
    let ts = string_to_timestamp_nanos(string).map(|t| t / 1_000_000).map_err(DataFusionError::from)?;
```

query_engine/src/logical_optimizer/type_conversion.rs
```diff
@@ -281,16 +282,59 @@ impl<'a> ExprRewriter for TypeRewriter<'a> {
...
+            "Error parsing '{s}' as timestamp: local time representation is invalid"
+        ))),
+        LocalResult::Single(local_datetime) => {
+            Ok(local_datetime.with_timezone(&Utc).timestamp_nanos() / 1_000_000)
+        }
+        // Ambiguous times can happen if the timestamp is exactly when
+        // a daylight savings time transition occurs, for example, and
+        // so the datetime could validly be said to be in two
+        // potential offsets. However, since we are about to convert
+        // to UTC anyways, we can pick one arbitrarily
```
ShiKaiWi, line 331:
```suggestion
        // The ambiguous result will happen if the timestamp is exactly when
        // a daylight savings time transition occurs, and in this case
        // the datetime could validly be said to be in two
        // potential offsets. However, since we are about to convert it
        // to UTC anyways, just pick the first one here.
```

query_engine/src/logical_optimizer/type_conversion.rs
```diff
@@ -2,7 +2,8 @@
 
 use std::{mem, sync::Arc};
 
-use arrow::{compute, compute::kernels::cast_utils::string_to_timestamp_nanos};
+use arrow::{compute, compute::kernels::cast_utils::string_to_timestamp_nanos, error::ArrowError};
+use chrono::{prelude::*, LocalResult};
```
ShiKaiWi, line 6: Replace `prelude::*` with specific modules used in this crate.

query_engine/src/logical_optimizer/type_conversion.rs
```diff
@@ -281,16 +282,59 @@ impl<'a> ExprRewriter for TypeRewriter<'a> {
 }
 
 fn string_to_timestamp_ms(string: &str) -> Result<ScalarValue> {
+    // TODO(lee): remove following codes after PR(https://github.com/apache/arrow-rs/pull/3787) merged
```
jiacai2050, line 285: I suggest add a new function for this workaround, something like
```rust
fn string_to_timestamp_ms_workaround() -> Result<Option<_>>

// Then you can just 
let ts = string_to_timestamp_ms_workaround(str).unwrap_or_else(string_to_timestamp_nanos(string))
```

query_engine/src/logical_optimizer/type_conversion.rs
MachaelLee: done

query_engine/src/logical_optimizer/type_conversion.rs
MachaelLee: This function is copied from arrow, I update comment to indicate this.

query_engine/src/logical_optimizer/type_conversion.rs
MachaelLee: done

query_engine/src/logical_optimizer/type_conversion.rs
MachaelLee: ok，done

jiacai2050 (approved):
LGTM

PR #694, opened 2023-03-03
refactor: replace tokio lock with std lock in some sync scenarios

ShiKaiWi:

Closes #

# Rationale for this change
 The tokio lock has performance issues when hot contention occurs, which may leading to yielding by tokio runtime schedule. That is to say, we should use std lock if the critical section is short or doesn't run across await point.

And the [api docs](https://docs.rs/tokio/1.26.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) of tokio has made a explanation for the choice between tokio lock and std lock.

- Replace the tokio lock with std lock as much as possible.

None.

# How does this change test
Existing tests.

### DIFF
--- a/common_util/src/partitioned_lock.rs
+++ b/common_util/src/partitioned_lock.rs
@@ -6,10 +6,8 @@ use std::{
     collections::hash_map::DefaultHasher,
     hash::{Hash, Hasher},
     num::NonZeroUsize,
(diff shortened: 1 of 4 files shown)


Rachelint (approved):
LGTM

PR #700, opened 2023-03-06
ci: upgrade the Rust version used in CI

ShiKaiWi:

Closes #

# Rationale for this change
 The rust version used in CI workflows is stale.

Update the rust version defined in CI workflows.

None.

# How does this change test
https://github.com/ShiKaiWi/ceresdb/actions/runs/4351144689

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,22 +31,20 @@ env:
   CERESMETA_IMAGE_NAME: ceresdb/ceresmeta-server:latest
   CERESDB_IMAGE_NAME: ceresdb/ceresdb-server:latest
   LOCK_FILE: Cargo.lock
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        rust: [ nightly-2022-08-08 ]
+        rust: [ nightly-2023-02-02 ]
```
jiacai2050, line 42: I suggest we remove this matrix, and add a env instead in L13, 

Then we can just use it like
```
rustup toolchain install ${{ env.RUST_VERSION }} --profile minimal

``

jiacai2050 (approved):
LGTM

PR #706, opened 2023-03-08
refactor: implement the distributed lock of shard

ZuLiangWang:

Closes #

# Rationale for this change
In order to ensure that user data in distributed mode will not be lost, we need a mechanism to ensure that CeresDB will not have multiple leader shard under any circumstances.

* Introduce `etcd` client in CeresDB.
* Distributed locks are implemented through `etcd`, and the corresponding locks need to be acquired every time when operating on the shard.

None

# How does this change test
Pass all unit tests and integration tests.


cluster/src/shard_lock_manager.rs
```diff
@@ -0,0 +1,149 @@
...
+            let (mut keeper, mut stream) = client.lease_keep_alive(lease_id).await.unwrap();
+            loop {
+                keep_alive_interval.tick().await;
+
+                info!("lease {:?} keep alive start", lease_id);
+                let resp = keeper.keep_alive().await;
+                match resp {
+                    Ok(_) => {}
+                    Err(e) => {
+                        // Close Shard
```
ShiKaiWi, line 117: More things besides remove shard from the cache need to be done when the lock lease is expired. Add a TODO here.

ShiKaiWi (commented):
Maybe we should create a branch for merging changes concerning the new cluster implementation.

cluster/src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub mod config;
 pub mod shard_tables_cache;
 // FIXME: Remove this lint ignore derive when topology about schema tables is
 // finished.
+pub mod shard_lock_manager;
 #[allow(dead_code)]
 pub mod topology;
```
chunshao90, line 33: Comment dose not match the code.

cluster/src/lib.rs
```diff
@@ -64,6 +65,12 @@ pub enum Error {
         source: GenericError,
     },
 
+    #[snafu(display("Fail to close shard, source:{}.", source))]
+    CloseShardWithCause {
+        shard_id: ShardId,
+        source: GenericError,
+    },
```
chunshao90, line 72: Print `shard_id`.

cluster/src/shard_lock_manager.rs
```diff
@@ -0,0 +1,326 @@
...
+                        error!("keep alive failed, shard will be removed, err msg {}", e);
+                        lease_failed_flag = true
+                    }
+                }
+
+                if lease_failed_flag {
+                    // TODO: Consider how to property close shard when lease is expired.
+                    let callback_result = lock_deleted_callback(shard_id);
+                    info!("invoke callback, result {:?}", callback_result);
```
ShiKaiWi, line 142: Log with shard id.

ShiKaiWi (commented):
Capitalize the first letter of the log message.

cluster/src/shard_lock_manager.rs
```diff
@@ -0,0 +1,505 @@
...
+            id,
+            ttl,
+            expired_at: Arc::new(AtomicU64::new(0)),
+            keep_alive_handle: None,
+        }
+    }
+
+    /// Check whether lease is expired.
+    fn is_expired(&self) -> bool {
+        let now = common_util::time::current_time_millis();
```
jiacai2050, line 120: This is not a monotonically clock, you should use Instant instead.
- https://docs.rs/chrono/latest/chrono/#date-and-time

cluster/src/shard_lock_manager.rs
```diff
@@ -0,0 +1,505 @@
...
+            ttl_sec,
+
+            lease_id: None,
+            lease_check_handle: None,
+            lease_keepalive_stopper: None,
+        }
+    }
+
+    fn lock_key(key_prefix: &str, shard_id: ShardId) -> Bytes {
+        let key = format!("{key_prefix}/{shard_id:0>20}");
```
jiacai2050, line 241: Only after see your tests I know `0>20` meaning, better to leave a comment here.

cluster/src/shard_lock_manager.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// be triggered.
+    pub async fn revoke_lock(&self, shard_id: u32) -> Result<bool> {
+        info!("Try to revoke lock for shard:{shard_id}");
+
+        let mut shard_locks = self.shard_locks.write().await;
+        let shard_lock = shard_locks.remove(&shard_id);
+        match shard_lock {
+            Some(mut v) => {
+                let mut etcd_client = self.etcd_client.clone();
+                v.revoke(&mut etcd_client).await?;
```
jiacai2050, line 473: What if this revoke failed?

server/src/grpc/meta_event_service/mod.rs
```diff
@@ -192,21 +194,79 @@ struct HandlerContext {
...
+    async fn acquire_shard_lock(&self, shard_id: ShardId) -> Result<()> {
+        let lock_mgr = self.cluster.shard_lock_manager();
+        let new_ctx = self.clone();
+        let on_lock_expired = |shard_id| async move {
+            warn!("Shard lock is released, try to close the tables and shard, shard_id:{shard_id}");
+            let close_shard_req = CloseShardRequest { shard_id };
+            let res = handle_close_shard(new_ctx, close_shard_req).await;
+            match res {
+                Ok(_) => info!("Close shard success, shard_id:{shard_id}"),
+                Err(e) => error!("Close shard failed, shard_id:{shard_id}, err:{e}"),
```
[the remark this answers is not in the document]
ShiKaiWi: Yes. In this case where the lease is expired, the network partition between ceresdb-server and etcd cluster has occurred, and what we can't do more than just hoping it to close shard successfully.

server/src/grpc/meta_event_service/mod.rs
```diff
@@ -192,21 +194,79 @@ struct HandlerContext {
...
+            .grant_lock(shard_id, on_lock_expired)
             .await
             .box_err()
             .context(ErrWithCause {
                 code: StatusCode::Internal,
-                msg: "fail to open shards in cluster",
+                msg: "fail to acquire shard lock",
             })?;
 
+        if !granted_by_this_call {
```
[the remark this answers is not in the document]
ShiKaiWi: When retrying close/open shard.

cluster/src/shard_lock_manager.rs
ShiKaiWi: I guess the interval of expire check is not very accurate, so the time offsetting is allowed.

cluster/src/shard_lock_manager.rs
ShiKaiWi: The error will be thrown up to the CeresMeta finally. And in the ceresmeta, the close shard call will fail and the following operations will be cancelled.

cluster/src/shard_lock_manager.rs
```diff
@@ -0,0 +1,505 @@
...
+
+/// Lock for a shard.
+///
+/// The lock is a temporary key in etcd, which is created with a lease.
+pub struct ShardLock {
+    shard_id: ShardId,
+    /// The temporary key in etcd
+    key: Bytes,
+    /// The value of the key in etcd
+    value: Bytes,
```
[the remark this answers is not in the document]
ShiKaiWi: The ceresmeta will read it to draw the whole topology: shard -> node.

PR #713, opened 2023-03-08
feat: support auto create table config

MachaelLee:

Closes #

## Rationale for this change
 
Currently CeresDB create table automatically when writing data, if table does not exist, the time column is named as 'timestamp' which may not be not the same as user defined

Add config item of `auto_create_tables`

User can configure `auto_create_tables' to allow the automatic creation of tables when writing data.


.gitignore
```diff
@@ -2,3 +2,4 @@ target
 .DS_Store
 .idea/
 .vscode
+proto
```
chunshao90, line 5: Why ignore proto?

server/src/config.rs
```diff
@@ -106,6 +106,9 @@ pub struct ServerConfig {
 
     /// Config for forwarding
     pub forward: forward::Config,
+
+    /// auto create tables, default true
+    pub auto_create_tables: bool,
```
chunshao90, line 111: Rename to `auto_create_table`?

server/src/config.rs
```diff
@@ -106,6 +106,9 @@ pub struct ServerConfig {
 
     /// Config for forwarding
     pub forward: forward::Config,
+
+    /// auto create tables, default true
```
jiacai2050, line 110:
```suggestion
    /// Whether to create table when data is first written, only used in gRPC
```

server/src/grpc/storage_service/write.rs
```diff
@@ -168,6 +169,7 @@ pub async fn execute_plan<Q: QueryExecutor + 'static>(
         })
 }
 
+#[allow(clippy::too_many_arguments)]
 pub async fn write_request_to_insert_plan<Q: QueryExecutor + 'static>(
```
jiacai2050, line 173: I think we can add a `WriteContext` to wrap args here:
```rs
pub struct WriteContext {
    pub request_id: RequestId,
    pub deadline: Option<Instant>,
    pub catalog: String,
    pub schema: String,
    pub auto_create_tables: bool
}

```

.gitignore
MachaelLee: I will remove it.

jiacai2050 (approved):
LGTM

PR #720, opened 2023-03-09
refactor: remove custom oss impl

### LINKED ISSUE #396: Support multipart upload sst file for aliyun oss
### Describe This Problem

According to [Multipart Upload](https://help.aliyun.com/document_detail/31991.htm), it will fail if we try to upload a large file whose size exceeds 100MB directly.

### Proposal

We should support multipart upload file to aliyun oss.

### Additional Context

_No response_

### LINKED ISSUE #608: Use OpenDAL to access object store
### Describe This Problem

[OpenDAL](https://github.com/datafuselabs/opendal) is widely adopted in Rust community to access files in many backends, such as S3, OSS, etc...

By integrate with it, we can gain more features from community. 

### Proposal

Replace AliyunOSS implementation with OpenDAL
- https://github.com/CeresDB/ceresdb/blob/main/components/object_store/src/aliyun.rs#L93

### Additional Context

_No response_

jiacai2050:

Closes #608, close #396

## Rationale for this change
 
Since OSS is S3-compatible, so we can access OSS via S3 client directly.
- https://www.alibabacloud.com/help/en/object-storage-service/latest/use-aws-s3-sdk-to-access-oss

- Use S3 client in object_store crate to access OSS

No

## How does this change test

Manually

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1755,7 +1755,7 @@ dependencies = [
  "lazy_static",
  "log",
  "num_cpus",
- "object_store 0.5.3",
+ "object_store 0.5.5",
  "parking_lot 0.12.1",
  "parquet",
  "paste 1.0.8",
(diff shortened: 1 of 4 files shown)


ShiKaiWi (approved):
LGTM

PR #723, opened 2023-03-10
feat: add influxdb write

jiacai2050:

Part of #682 

- Refactor influxdb http handler to make it more modular  
- Support influxdb line protocol write(convert to gRPC write internal)

No

## How does this change test

CI

### DIFF
--- a/integration_tests/src/database.rs
+++ b/integration_tests/src/database.rs
@@ -178,7 +178,7 @@ impl CeresDB {
     }
 
     async fn execute_influxql(query: String, http_client: HttpClient) -> Box<dyn Display> {
-        let url = format!("http://{}/influxql", http_client.endpoint);
+        let url = format!("http://{}/influxdb/v1/query", http_client.endpoint);
         let query_request = InfluxQLRequest { query };
         let resp = http_client
             .client
--- a/integration_tests/src/database.rs
+++ b/integration_tests/src/database.rs
@@ -18,7 +18,6 @@ use ceresdb_client::{
     RpcContext,
 };
 use reqwest::ClientBuilder;
-use serde::Serialize;
 use sql::{
     ast::{Statement, TableName},
     parser::Parser,
@@ -44,11 +43,6 @@ pub enum DeployMode {
 struct HttpClient {
     client: reqwest::Client,
     endpoint: String,
-}
-
-#[derive(Clone, Serialize)]
-struct InfluxQLRequest {
-    query: String,
 }
 
 impl HttpClient {
@@ -179,11 +173,10 @@ impl CeresDB {
 
     async fn execute_influxql(query: String, http_client: HttpClient) -> Box<dyn Display> {
         let url = format!("http://{}/influxdb/v1/query", http_client.endpoint);
-        let query_request = InfluxQLRequest { query };
         let resp = http_client
             .client
             .post(url)
-            .json(&query_request)
+            .body(query)
             .send()
             .await
             .unwrap();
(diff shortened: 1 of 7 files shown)


integration_tests/src/database.rs
```diff
@@ -178,12 +172,11 @@ impl CeresDB {
     }
 
     async fn execute_influxql(query: String, http_client: HttpClient) -> Box<dyn Display> {
-        let url = format!("http://{}/influxql", http_client.endpoint);
-        let query_request = InfluxQLRequest { query };
+        let url = format!("http://{}/influxdb/v1/query", http_client.endpoint);
         let resp = http_client
             .client
             .post(url)
-            .json(&query_request)
```
jiacai2050, line 186: Only support raw text, json support is removed.

integration_tests/src/database.rs
```diff
@@ -178,12 +172,11 @@ impl CeresDB {
...
+            .body(query)
```
ShiKaiWi, line 179: The integration tests need updating if so.

server/src/handlers/error.rs
```diff
@@ -70,6 +72,11 @@ pub enum Error {
         source: tokio::time::error::Elapsed,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("Influxdb handler failed, msg:{}, source:{}", msg, source))]
+    InfluxdbHandler { msg: String, source: GenericError },
```
ShiKaiWi, line 77: Shall name it InfluxDB or InfluxDb?

server/src/handlers/influxdb.rs
```diff
@@ -0,0 +1,366 @@
...
+        debug!(
+            "Remote write finished, catalog:{}, schema:{}, success:{}",
+            catalog, schema, success
+        );
+
+        Ok(())
+    }
+}
+
+fn convert_write_request(req: WriteRequest) -> Result<Vec<WriteTableRequest>> {
```
ShiKaiWi, line 143: This conversion seems very inefficient. Actually, the underlying write procedure handles the request row by row, so maybe we can convert the write request into `Rows`(defined in common_types/src/row/mod.rs) rather than  `WriteTableRequest` here.

jiacai2050: > This conversion seems very inefficient. 

Could you explain in more details? All SDK use this `WriteTableRequest` type to do write.

> so maybe we can convert the write request into Rows

Here I reuse write logic for gRPC, and this level reuse is also applied for prom write.  

IMO `Rows` isn't suitable at protocol level. If you want to reuse lower level logic, then more work is needed.

server/src/handlers/error.rs
jiacai2050: `InfluxDb` seems more suitable here.

integration_tests/src/database.rs
jiacai2050: Only this place need to update.

server/src/handlers/influxdb.rs
ShiKaiWi: > Could you explain in more details? All SDK use this WriteTableRequest type to do write.
Too many clones here. Furthermore, another conversion from the `WriteTableRequest` to `RowGroup` follows, which makes this conversion here redundant.

ShiKaiWi: Maybe we can leave the work in the future.

jiacai2050: I will create a new issue to track this.

ShiKaiWi (approved):
LGTM

PR #727, opened 2023-03-13
Feat proxy prom query

chunshao90:

Closes #

## Rationale for this change
 Impl prom query with proxy.

* Impl prom query with proxy.
* Refactor some code in write and sql query.


ShiKaiWi (approved):
LGTM

PR #728, opened 2023-03-14
fix: Panicked when OceanBase table client is initialing

MachaelLee:

Closes #

## Rationale for this change
 
related with https://github.com/oceanbase/obkv-table-client-rs/pull/21

Table client will init timeout when one OB server is invalid, eg, OUT_OF_MEM, SERVER_IS_INIT.

Update OceanBase rust client

## How does this change test

Integration test


ShiKaiWi (approved):
LGTM

PR #730, opened 2023-03-14
feat: reuse logical planner in influxdb_iox

### LINKED ISSUE #724: Reuse the logical planner in influxdb_iox
### Describe This Problem

Influxdb_iox has done many works in planner of influxql, maybe it's not reasonable to do the duplicated works rather than resuing and building it together.

### Proposal

- [ ] Extract the logical planner from influxdb_iox
  - [x] Make `Schema` a trait and make it able to build.
  - [ ] Make effort to pass the exist uts.
- [ ] Make use of it in CeresDB by impl the related traits.

### Additional Context

_No response_

Rachelint:

Closes #724 

## Rationale for this change
Reust logical planner in influxdb_iox.
[See detail](https://github.com/influxdata/influxdb_iox/issues/6112)

See title.

None.

## How does this change test
Test by ut.


sql/src/influxql/provider.rs
```diff
@@ -0,0 +1,466 @@
...
+            | DatumKind::Float
+            | DatumKind::Varbinary
+            | DatumKind::UInt32
+            | DatumKind::UInt16
+            | DatumKind::UInt8
+            | DatumKind::Int32
+            | DatumKind::Int16
+            | DatumKind::Int8
+            | DatumKind::Date
+            | DatumKind::Time => BuildSchema {
```
ShiKaiWi, line 169: Why is conversion for these types disallowed?

sql/src/influxql/provider.rs
```diff
@@ -0,0 +1,466 @@
...
+            }
+            .fail()
+        }
+    } else {
+        // Field column
+        map_field_type_to_influx_field_type(column).map(InfluxColumnType::Field)
+    }
+}
+
+fn map_field_type_to_influx_field_type(column: &ColumnSchema) -> Result<InfluxFieldType> {
```
ShiKaiWi, line 150: I guess the input should be limited to only two parameters: `nullable` and `data_type`, which will be friendly for building unit test.

sql/src/influxql/provider.rs
```diff
@@ -0,0 +1,466 @@
...
+};
+
+use crate::{
+    influxql::error::*,
+    provider::{ContextProviderAdapter, MetaProvider},
+};
+
+/// Influx schema used for build logical plan
+pub struct InfluxSchemaImpl {
+    columns: Vec<(InfluxColumnType, ArrowField)>,
```
ShiKaiWi, line 23: How about define a type called `InfluxColumnSchema` to replace this tuple?

sql/src/influxql/provider.rs
```diff
@@ -0,0 +1,466 @@
...
+        // Tag column
+        if matches!(column.data_type, DatumKind::String) && column.is_nullable {
+            Ok(InfluxColumnType::Tag)
+        } else {
+            BuildSchema {
+                msg: format!("invalid tag column, column:{column:?}"),
+            }
+            .fail()
+        }
+    } else {
```
ShiKaiWi, line 144: I guess we should define two functions called `map_column_type_to_influx_time_column` and `map_column_type_to_influx_tag_column`, which accepts the parameters and return `Option<InfluxColumnType>`. With the two function, it may make the structure of this function more simple.

sql/src/influxql/provider.rs
```diff
@@ -0,0 +1,466 @@
...
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    fn time(&self) -> &ArrowField {
+        // Time column must exist, has checked it when building.
+        let time_column = self
```
jiacai2050, line 96: How about add a field `timestamp_idx` in InfluxSchemaImpl to avoid this `find`?

sql/src/influxql/provider.rs
```diff
@@ -0,0 +1,466 @@
...
+    }
+}
+
+fn map_column_type_to_influx_column_type(
+    column: &ColumnSchema,
+    is_timestamp_key: bool,
+) -> Result<InfluxColumnType> {
+    // Time column
+    if is_timestamp_key {
+        if column.name == "time" && !column.is_nullable {
```
jiacai2050, line 126: I prefer to return early to avoid too many `else if` clause.

jiacai2050: Why this column must be `time`? in our codebase default ts column name is `timestamp`, not `time`.

I think the only requirement is there must exists a timestamp column in schema, no name constraint.

jiacai2050 (changes requested):

sql/src/influxql/provider.rs
Rachelint: I add a comment to analyze this, because they are not supported in influxdb.
Maybe we can add a todo for supporting them?

ShiKaiWi: Ok. Let's add a todo here.

ShiKaiWi:
@Rachelint  I'm OK with this PR. Let's wait for suggestions from @jiacai2050.

jiacai2050 (approved):
LGTM

Issue #681, opened 2023-03-01
Tracking issue: Introduce proxy module


ShiKaiWi:
### Describe This Problem

Currently, it is just a simple module called `forwarder` that forwards the sql query to the right server, but more work can't be handled by the `forwarder`, including:
- Forward according to the sql rather than the table hints in the sql query;
- Forward all the requests besides the sql query;
- Able to finish query/write about the partitioned table;

### Proposal

Design a proxy module which can serve as a separate service or a module in the ceresdb-server.

chunshao90:
Support automatic table creation in proxy.

jiacai2050:
> Forward according to the sql rather than the table hints in the sql query;

SQL may not enough, other query language should also be supported.

chunshao90:
I am going to introduce the proxy module, and all protocol layers request the proxy module. The proxy module supports `Forword`, `Auth`, `PartitionTable` query.

<img width="732" alt="image" src="https://user-images.githubusercontent.com/15178480/223653720-d92e0992-a61b-457f-b58f-25254fdc1964.png">

chunshao90:
I split the work into the following tasks:

- [ ] The proxy module first supports storage service.
- [ ] Integrate forward in proxy.
- [ ] The proxy module supports mysql service.
- [ ] The proxy module supports http service.

PR #732, opened 2023-03-14
feat: introduce proxy module

chunshao90:

Closes #

## Rationale for this change
 Introduce proxy module. Refer to #681 .

* Introduce proxy module.
* Impl storage_service with proxy.
* Support forwarding stream query.

No.

## How does this change test
Integration test.


docs/minimal.toml
```diff
@@ -1,3 +1,6 @@
+[node]
+addr = "ceresdb0"
```
ShiKaiWi, line 2: It is weird to use a name for addr.

server/src/proxy/mod.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright 2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+pub(crate) mod error;
```
jiacai2050, line 3: Add crates description explain what it does

docs/minimal.toml
jiacai2050: is this a required config, even for standalone mode?

server/src/proxy/forward.rs
```diff
@@ -0,0 +1,506 @@
...
+    /// default keep http2 connections alive while idle
+    pub keep_alive_while_idle: bool,
+    pub connect_timeout: Duration,
+    pub forward_timeout: Duration,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            enable: false,
```
ShiKaiWi, line 103: Shall we deprecate this option? It should be always enabled.

server/src/proxy/error.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use crate::error_util;
+
+define_result!(Error);
+
+#[derive(Snafu, Debug)]
+#[snafu(visibility(pub))]
+pub enum Error {
+    #[snafu(display("Internal error, message:{}, cause:{}", msg, source))]
+    Internal { msg: String, source: GenericError },
```
ShiKaiWi, line 16: Any differences between `Internal` and `ErrNoCause`?

server/src/proxy/grpc/sql_query.rs
```diff
@@ -0,0 +1,490 @@
...
+        let req = match self.maybe_forward_sql_query(&req).await {
+            Some(resp) => return resp,
+            None => req,
+        };
+
+        let output = self.fetch_sql_query_output(ctx, &req).await?;
+        convert_output(&output, self.resp_compress_min_length)
+    }
+
+    async fn handle_stream_query_internal(
```
ShiKaiWi, line 82:
```suggestion
    async fn handle_stream_sql_query_internal(
```

ShiKaiWi (approved):
LGTM

PR #737, opened 2023-03-15
feat: impl stream write with proxy

chunshao90:

Closes #

## Rationale for this change
 Impl stream write with proxy.

 Impl stream write with proxy.


ShiKaiWi (approved):
LGTM

PR #739, opened 2023-03-15
fix: remove compaction retry when memory limit

MachaelLee:

Closes #

## Rationale for this change
 
Compaction retry immediately  when memory limit, result in lots of logs will prints.

* remove compaction retry
* modify log level to `warn`

None

## How does this change test

None


ShiKaiWi (approved):
LGTM

PR #740, opened 2023-03-15
feat: remove replace table level metrics with aggregate metrics

### LINKED ISSUE #718: Simplify the metrics
### Describe This Problem

Currently, the metrics has been recorded for every single table, which leading to massive metrics (The size of the metrics may reach up to 100MB+ when the table number is 10K+).

### Proposal

Simplify the metrics and use the aggregated metrics rather than the table level metrics, which should be replaced with the trace.

### Additional Context

_No response_

ShiKaiWi:

Closes #718 

## Rationale for this change
Refer to #718 

- Replace the table-level metrics with aggregate metrics

The metrics changes.

## How does this change test
Manually.


integration_tests/build_meta.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -exo
+
+if [ ! -d ceresmeta ]; then
+    git clone https://github.com/ceresdb/ceresmeta.git
```
jiacai2050, line 6: `git clone -–depth 1`

chunshao90 (approved):
LGTM

PR #742, opened 2023-03-16
feat: impl stream query with proxy

chunshao90:

Closes #

## Rationale for this change
 Impl stream query with proxy.

* Impl stream query with proxy.
* Support forwarding stream query.

No.

## How does this change test
Integration test.


server/src/grpc/storage/mod.rs
```diff
@@ -224,7 +252,16 @@ impl<Q: QueryExecutor + 'static> StorageServiceImpl<Q> {
...
             };
 
-            let write_resp = self.proxy.handle_write(Context::default(), write_req).await;
+            let write_resp = self
+                .proxy
+                .handle_write(
+                    Context {
+                        runtime: self.runtimes.write_runtime.clone(),
+                        timeout: self.timeout,
+                    },
```
ShiKaiWi, line 261: Initialize this context outside this function call.

ShiKaiWi (approved):
LGTM

PR #722, opened 2023-03-10
fix: failure to open a single table does not interrupt the shard's opening process

MachaelLee:
## Which issue does this PR close?

Closes #721 

## Rationale for this change
 
 The shard will stop opening if opening a table fails, and no other normal tables will be opened in that shard.
 Access to these tables will fail.

## What changes are included in this PR?

* Print warn log if opening a table fails
* Continue to open shard

PR #743, opened 2023-03-16
feat: don't allow create table which failed when open

jiacai2050:

Closes #

## Rationale for this change
 
Followup PR of #722

When table open failed when recovery, we should prevent user from creating table with same name, otherwise schema maybe wong, and seq will be reset to 0.

- Add open failed table in Space, and check it when create new table.

No

## How does this change test

CI, but I don't test this corner case since I don't know how to mock recovery failure of one table.

### DIFF
--- a/analytic_engine/src/instance/engine.rs
+++ b/analytic_engine/src/instance/engine.rs
@@ -214,7 +214,7 @@ pub enum Error {
     },
 
     #[snafu(display(
-        "{} open failed and can not be created again.\nBacktrace:\n{}",
+        "Table open failed and can not be created again, table:{}.\nBacktrace:\n{}",
         table,
         backtrace,
     ))]
--- a/analytic_engine/src/instance/engine.rs
+++ b/analytic_engine/src/instance/engine.rs
@@ -212,6 +212,13 @@ pub enum Error {
         table: String,
         source: GenericError,
     },
+
+    #[snafu(display(
+        "{} open failed and can not be created again.\nBacktrace:\n{}",
+        table,
+        backtrace,
+    ))]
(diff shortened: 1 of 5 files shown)


analytic_engine/src/instance/engine.rs
```diff
@@ -212,6 +212,13 @@ pub enum Error {
         table: String,
         source: GenericError,
     },
+
+    #[snafu(display(
+        "{} open failed and can not be created again.\nBacktrace:\n{}",
```
chunshao90, line 217:
```suggestion
        "Table open failed and can not be created again, table:{}.\nBacktrace:\n{}",
```

analytic_engine/src/instance/open.rs
```diff
@@ -189,13 +189,19 @@ impl Instance {
...
+        if let Err(e) = self
+            .recover_table_from_wal(
+                worker_local,
+                table_data.clone(),
+                replay_batch_size,
+                &read_ctx,
+            )
+            .await
+        {
+            error!("Recovery table failed, table_data:{table_data:?}");
```
chunshao90, line 201: Print error.

analytic_engine/src/instance/open.rs
```diff
@@ -189,13 +189,19 @@ impl Instance {
...
-        )
-        .await?;
+        if let Err(e) = self
+            .recover_table_from_wal(
+                worker_local,
+                table_data.clone(),
+                replay_batch_size,
+                &read_ctx,
+            )
+            .await
```
chunshao90, line 199: .map_err(...)

chunshao90 (approved):
LGTM

PR #745, opened 2023-03-16
feat: return error while encountering unsupport from in influxql

Rachelint:

Closes #

## Rationale for this change
Now, we are unable to support from multiple measurements in influxql.
So, we should check and reject them. 

See title.

Will get error information while pass the influxql which selects from multiple measurement.

## How does this change test
Test by ut.

### DIFF
--- a/sql/src/influxql/mod.rs
+++ b/sql/src/influxql/mod.rs
@@ -27,13 +27,13 @@ pub mod error {
         BuildPlanWithCause { msg: String, source: GenericError },
 
         #[snafu(display(
-            "Failed to build influxdb plan with no cause, msg:{}.\nBacktrace:{}",
+            "Failed to build influxql plan with no cause, msg:{}.\nBacktrace:{}",
             msg,
             backtrace
         ))]
         BuildPlanNoCause { msg: String, backtrace: Backtrace },
 
-        #[snafu(display("Unimplemented influxql statement, msg{}", msg))]
+        #[snafu(display("Unimplemented influxql statement, msg:{}", msg))]
         Unimplemented { msg: String },
     }
     define_result!(Error);
--- a/sql/src/influxql/mod.rs
+++ b/sql/src/influxql/mod.rs
@@ -19,11 +19,22 @@ pub mod error {
         ))]
         BuildSchema { msg: String, backtrace: Backtrace },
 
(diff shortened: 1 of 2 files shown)


sql/src/influxql/mod.rs
```diff
@@ -19,11 +19,22 @@ pub mod error {
...
+        #[snafu(display(
+            "Failed to build influxdb plan with no cause, msg:{}.\nBacktrace:{}",
+            msg,
+            backtrace
+        ))]
+        BuildPlanNoCause { msg: String, backtrace: Backtrace },
 
-        #[snafu(display("Unimplemented influxql statement, statement:{}", stmt))]
-        Unimplemented { stmt: String },
+        #[snafu(display("Unimplemented influxql statement, msg{}", msg))]
```
jiacai2050, line 36:
```suggestion
        #[snafu(display("Unimplemented influxql statement, msg:{}.", msg))]
```

sql/src/influxql/mod.rs
```diff
@@ -19,11 +19,22 @@ pub mod error {
         ))]
         BuildSchema { msg: String, backtrace: Backtrace },
 
-        #[snafu(display("Failed to build influxql plan, msg:{}, err:{}", msg, source))]
-        BuildPlan { msg: String, source: GenericError },
+        #[snafu(display(
+            "Failed to build influxql plan with cause, msg:{}, err:{}",
```
jiacai2050, line 23:
```suggestion
            "Failed to build influxql plan with cause, msg:{}, source:{}",
```

sql/src/influxql/mod.rs
```diff
@@ -19,11 +19,22 @@ pub mod error {
...
+            msg,
+            source
+        ))]
+        BuildPlanWithCause { msg: String, source: GenericError },
+
+        #[snafu(display(
+            "Failed to build influxql plan with no cause, msg:{}.\nBacktrace:{}",
```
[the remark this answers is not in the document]
jiacai2050: It's best we can add a macro to fix this. I always forge this.

```
display_backtrace!("Failed to build influxql plan with no cause, msg:{}.")
```
It will rewrite to line above.

jiacai2050 (approved):
LGTM

PR #747, opened 2023-03-16
feat: build sst in stream way

### LINKED ISSUE #486: Build sst file in more resource friendly way
### Describe This Problem

In current implementation, sst write involves two loop:
1. First loop to calculate bloom filters of each row group
2. Second loop to write to parquet

Two loops mean more cpu usage, what's worse it that this may eat too many memory.

### Proposal

It's best we can reduce build procedure to one loop, this depends on https://github.com/apache/arrow-rs/issues/3356

If this is not possible, then we may need to spit RecordBatch to disk in order to reduce memory consumption.

### Additional Context

_No response_

ShiKaiWi:

Closes #486 

## Rationale for this change
 Refer to #486 

- Introduce `AsyncArrowWriter` in the `parquet_ext` crate;
- Build sst in stream way using `AsyncArrowWriter`;

None.

## How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -22,7 +22,7 @@ pub mod table_options;
 #[cfg(any(test, feature = "test"))]
 pub mod tests;
 
-use common_util::config::ReadableDuration;
+use common_util::config::{ReadableDuration, ReadableSize};
 use manifest::details::Options as ManifestOptions;
 use message_queue::kafka::config::Config as KafkaConfig;
 use serde::{Deserialize, Serialize};
@@ -76,7 +76,7 @@ pub struct Config {
     /// Sst background reading parallelism
     pub sst_background_read_parallelism: usize,
     /// Max buffer size for writing sst
-    pub write_sst_max_buffer_size: usize,
+    pub write_sst_max_buffer_size: ReadableSize,
 
     /// Wal storage config
     ///
@@ -110,7 +110,7 @@ impl Default for Config {
             db_write_buffer_size: 0,
             scan_batch_size: 500,
             sst_background_read_parallelism: 8,
-            write_sst_max_buffer_size: 1024 * 1024 * 10,
+            write_sst_max_buffer_size: ReadableSize::mb(10),
             wal: WalStorageConfig::RocksDB(Box::default()),
             remote_engine_client: remote_engine_client::config::Config::default(),
         }
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -75,6 +75,8 @@ pub struct Config {
     pub scan_batch_size: usize,
     /// Sst background reading parallelism
     pub sst_background_read_parallelism: usize,
+    /// Max buffer size for writing sst
+    pub write_sst_max_buffer_size: usize,
 
     /// Wal storage config
     ///
@@ -108,6 +110,7 @@ impl Default for Config {
             db_write_buffer_size: 0,
             scan_batch_size: 500,
             sst_background_read_parallelism: 8,
+            write_sst_max_buffer_size: 1024 * 1024 * 10,
             wal: WalStorageConfig::RocksDB(Box::default()),
             remote_engine_client: remote_engine_client::config::Config::default(),
         }
--- a/analytic_engine/src/sst/parquet/writer.rs
+++ b/analytic_engine/src/sst/parquet/writer.rs
@@ -163,7 +163,7 @@ impl RecordBatchGroupWriter {
 
     async fn write_all<W: AsyncWrite + Send + Unpin + 'static>(mut self, sink: W) -> Result<usize> {
         let mut prev_record_batch: Option<RecordBatchWithKey> = None;
-        let mut parquet_filter = ParquetFilter::with_row_group_num(0);
+        let mut parquet_filter = ParquetFilter::default();
         let mut arrow_row_group = Vec::new();
         let mut total_num_rows = 0;
 
--- a/analytic_engine/src/sst/parquet/writer.rs
+++ b/analytic_engine/src/sst/parquet/writer.rs
@@ -73,28 +73,9 @@ struct RecordBytesReader {
     num_rows_per_row_group: usize,
     compression: Compression,
     total_row_num: Arc<AtomicUsize>,
-    // Record batch partitioned by exactly given `num_rows_per_row_group`
-    // There may be more than one `RecordBatchWithKey` inside each partition
-    partitioned_record_batch: Vec<Vec<RecordBatchWithKey>>,
 }
 
 impl RecordBytesReader {
-    // Partition record batch stream into batch vector with exactly given
-    // `num_rows_per_row_group`
-    async fn partition_record_batch(&mut self) -> Result<()> {
-        let mut prev_record_batch: Option<RecordBatchWithKey> = None;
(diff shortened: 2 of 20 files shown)


components/parquet_ext/src/async_arrow_writer.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.async_writer
+            .shutdown()
+            .await
+            .map_err(|e| ParquetError::External(Box::new(e)))?;
+
+        Ok(metadata)
+    }
+
+    async fn flush(
+        shared_buffer: &SharedBuffer,
```
jiacai2050, line 97: Why this flush is not an associated method? `fn flush(&mut self ...)`

components/parquet_ext/src/async_arrow_writer.rs
```diff
@@ -0,0 +1,121 @@
...
+        Write::write(&mut *buffer, buf)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let mut buffer = self.buffer.lock().unwrap();
+        Write::flush(&mut *buffer)
+    }
+}
+
+/// Async arrow writer for parquet file with a buffer whose size limit is
```
jiacai2050, line 42: I guess you should explain how this async writer is implemented based on a sync one.

And `whose size limit is configurable` seems useless.

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -149,75 +128,131 @@ impl RecordBytesReader {
...
             .context(EncodeRecordBatch)?;
-        Ok(bytes)
+
+        Ok(total_num_rows)
+    }
+}
+
+struct ObjectStoreMultiUploadAborter<'a> {
+    location: &'a Path,
+    sink_id: String,
```
jiacai2050, line 227: `session_id` seems more suitable here

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -448,54 +491,53 @@ mod tests {
...
-        reader.partition_record_batch().await.unwrap();
-
-        for (i, expected_row_num) in expected_row_nums.into_iter().enumerate() {
-            let actual: usize = reader.partitioned_record_batch[i]
-                .iter()
-                .map(|b| b.num_rows())
-                .sum();
-            assert_eq!(expected_row_num, actual);
-        }
+        // init_log_for_test();
```
jiacai2050, line 498: Reopen this test?

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -149,75 +128,131 @@ impl RecordBytesReader {
...
+
+impl<'a> ObjectStoreMultiUploadAborter<'a> {
+    async fn initialize_upload(
+        object_store: &'a ObjectStoreRef,
+        location: &'a Path,
+    ) -> Result<(
+        ObjectStoreMultiUploadAborter<'a>,
+        Box<dyn AsyncWrite + Unpin + Send>,
+    )> {
+        let (sink_id, sink) = object_store
```
jiacai2050, line 239: this module never has `sink` variable, I prefer to name it `writer`.

jiacai2050:
I guess features should be tested against OSS.

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -149,75 +128,131 @@ impl RecordBytesReader {
         Ok(curr_row_group)
     }
 
-    fn build_parquet_filter(&self) -> Result<ParquetFilter> {
+    /// Build the parquet filter for the given `row_group`, and then update it
+    /// into `exist_filter`.
+    fn build_parquet_filter(
+        &self,
+        row_group_batch: &[RecordBatchWithKey],
+        exist_filter: &mut ParquetFilter,
```
jiacai2050, line 136: How about add parquetFilter as field of `RecordBatchGroupWriter`?

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -227,40 +262,47 @@ impl<'a> SstWriter for ParquetSstWriter<'a> {
...
-            .await
-            .context(Storage)?;
 
-        let file_head = self.store.head(self.path).await.context(Storage)?;
+        let (aborter, sink) =
+            ObjectStoreMultiUploadAborter::initialize_upload(self.store, self.path).await?;
+        let total_num_rows = match group_writer.write_all(sink).await {
+            Ok(v) => v,
+            Err(e) => {
+                if let Err(e) = aborter.abort().await {
```
jiacai2050, line 287: Leave a comment that this may leave temp file in underlying storage, and we may need to provide a repair command to clean those.

components/parquet_ext/src/async_arrow_writer.rs
ShiKaiWi: When implementing `close` method, the `sync_writer` has to be consumed before `flush` which requires the `flush` method not to be the associated method.

analytic_engine/src/sst/parquet/writer.rs
ShiKaiWi: How about `upload_writer`?

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -227,40 +262,47 @@ impl<'a> SstWriter for ParquetSstWriter<'a> {
...
+        let file_head = self.store.head(self.path).await.context(Storage)?;
         let storage_format = if self.hybrid_encoding {
             StorageFormat::Hybrid
         } else {
             StorageFormat::Columnar
         };
         Ok(SstInfo {
             file_size: file_head.size,
-            row_num: total_row_num.load(Ordering::Relaxed),
+            row_num: total_num_rows,
```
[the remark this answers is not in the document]
ShiKaiWi: `row_num` is not a good name because `num_rows` is used across current code base including `arrow` dependency.

analytic_engine/src/lib.rs
```diff
@@ -75,6 +75,8 @@ pub struct Config {
     pub scan_batch_size: usize,
     /// Sst background reading parallelism
     pub sst_background_read_parallelism: usize,
+    /// Max buffer size for writing sst
+    pub write_sst_max_buffer_size: usize,
```
jiacai2050, line 79: ReadableSize

components/parquet_ext/src/async_arrow_writer.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl<W: AsyncWrite + Unpin + Send> AsyncArrowWriter<W> {
+    pub fn try_new(
+        writer: W,
+        arrow_schema: SchemaRef,
+        max_buffer_size: usize,
+        props: Option<WriterProperties>,
+    ) -> Result<Self> {
+        let shared_buffer = SharedBuffer::default();
```
jiacai2050, line 59: `SharedBuffer::with_capacity(max_buffer_size)`

ShiKaiWi: Maybe we should let it grow by itself, whose cost will be amortized by the whole encoding procedure, otherwise it may consume much memory if a very large `max_buffer_size` is provided for an unlimited purpose.

jiacai2050 (approved):
LGTM

PR #748, opened 2023-03-16
feat: implement route cache

### LINKED ISSUE #738: Avoid route request penetration into CeresMeta
### Describe This Problem

Current route request implementation is just forwarding the request to CeresMeta, which is very dangerous to CeresMeta when massive route requests are received.

### Proposal

Add a cache for route requests, which should be refreshed periodically. With the cache, the route requests from the client won't be forwarded to CeresMeta.

### Additional Context

_No response_

MachaelLee:

Closes #738 

## Rationale for this change
 
see #738 

add route cache 

None

## How does this change test

unit test

### DIFF
--- a/router/src/cluster_based.rs
+++ b/router/src/cluster_based.rs
@@ -2,23 +2,54 @@
 
 //! A router based on the [`cluster::Cluster`].
 
+use std::time::Duration;
+
 use async_trait::async_trait;
 use ceresdbproto::storage::{Route, RouteRequest};
 use cluster::ClusterRef;
 use common_util::error::BoxError;
 use meta_client::types::RouteTablesRequest;
+use moka::future::Cache;
 use snafu::ResultExt;
 
-use crate::{endpoint::Endpoint, OtherWithCause, ParseEndpoint, Result, Router};
+use crate::{endpoint::Endpoint, OtherWithCause, ParseEndpoint, Result, RouteCacheConfig, Router};
 
 pub struct ClusterBasedRouter {
     cluster: ClusterRef,
+    cache: Cache<String, Route>,
 }
 
 impl ClusterBasedRouter {
-    pub fn new(cluster: ClusterRef) -> Self {
-        Self { cluster }
+    pub fn new(cluster: ClusterRef, cache_config: RouteCacheConfig) -> Self {
+        Self {
+            cluster,
+            cache: Cache::builder()
+                .time_to_live(Duration::from_secs(cache_config.ttl))
+                .time_to_idle(Duration::from_secs(cache_config.tti))
+                .max_capacity(cache_config.capacity)
+                .build(),
+        }
     }
+}
+
+/// route table from local cache, return cache routes and  tables which are not
+/// in cache
+fn route_from_cache(
+    cache: &Cache<String, Route>,
+    tables: Vec<String>,
+) -> (Vec<Route>, Vec<String>) {
+    let mut routes = vec![];
+    let mut miss = vec![];
+
+    for table in tables {
+        if let Some(route) = cache.get(&table) {
+            routes.push(route);
+        } else {
+            miss.push(table.clone());
+        }
+    }
+
+    (routes, miss)
 }
 
 /// Make a route according to the table name and the raw endpoint.
@@ -35,10 +66,15 @@ fn make_route(table_name: &str, endpoint: &str) -> Result<Route> {
 impl Router for ClusterBasedRouter {
     async fn route(&self, req: RouteRequest) -> Result<Vec<Route>> {
         let req_ctx = req.context.unwrap();
+
+        // Firstly route table from local cache
+        let (mut routes, miss) = route_from_cache(&self.cache, req.tables);
+
         let route_tables_req = RouteTablesRequest {
             schema_name: req_ctx.database,
-            table_names: req.tables,
+            table_names: miss,
         };
+
         let route_resp = self
             .cluster
             .route_tables(&route_tables_req)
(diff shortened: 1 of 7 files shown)


router/src/lib.rs
```diff
@@ -63,3 +63,29 @@ pub type RouterRef = Arc<dyn Router + Sync + Send>;
...
+    /// Time to idle (TTI) in second
+    tti: u64,
+    /// how many route records can store in cache
+    capacity: u64,
+}
+
+impl RouteCacheConfig {
+    pub fn new() -> Self {
+        Self {
+            ttl: 5 * 60,
```
chunshao90, line 80: 5 minutes is too long, which will cause the routing cache in CeresDB to be inconsistent with the meta, which can be changed to 5 seconds.

router/src/lib.rs
```diff
@@ -63,3 +63,29 @@ pub type RouterRef = Arc<dyn Router + Sync + Send>;
...
+            tti: 2 * 60,
+            capacity: 10_000,
+        }
+    }
+}
```
chunshao90, line 85:
```suggestion
        Self::default()
    }
}
```

server/src/config.rs
```diff
@@ -125,6 +128,7 @@ impl Default for ServerConfig {
             resp_compress_min_length: ReadableSize::mb(4),
             forward: forward::Config::default(),
             auto_create_table: true,
+            route_cache: router::RouteCacheConfig::new(),
```
chunshao90, line 131:
```suggestion
            route_cache: router::RouteCacheConfig::default(),
```

router/src/cluster_based.rs
```diff
@@ -35,31 +66,85 @@ fn make_route(table_name: &str, endpoint: &str) -> Result<Route> {
...
-        let mut routes = Vec::with_capacity(route_resp.entries.len());
-
-        // Now we pick up the nodes who own the leader shard for the route response.
-        for (table_name, route_entry) in route_resp.entries {
-            for node_shard in route_entry.node_shards {
-                if node_shard.shard_info.is_leader() {
-                    let route = make_route(&table_name, &node_shard.endpoint)?;
-                    routes.push(route);
+
+        // Firstly route table from local cache
```
chunshao90, line 70: Add a period after each sentence.

router/src/cluster_based.rs
```diff
@@ -2,23 +2,54 @@
...
+        }
+    }
+}
+
+/// route table from local cache, return cache routes and  tables which are not
+/// in cache
+fn route_from_cache(
+    cache: &Cache<String, Route>,
+    tables: Vec<String>,
+) -> (Vec<Route>, Vec<String>) {
```
chunshao90, line 40:
```suggestion
fn route_from_cache(
    &self,
    tables: Vec<String>,
) -> (Vec<Route>, Vec<String>) {
```

router/src/cluster_based.rs
```diff
@@ -35,31 +66,85 @@ fn make_route(table_name: &str, endpoint: &str) -> Result<Route> {
...
+                .with_context(|| OtherWithCause {
+                    msg: format!("Failed to route tables by cluster, req:{route_tables_req:?}"),
+                })?;
+
+            // Now we pick up the nodes who own the leader shard for the route response.
+            for (table_name, route_entry) in route_resp.entries {
+                for node_shard in route_entry.node_shards {
+                    if node_shard.shard_info.is_leader() {
+                        let route = make_route(&table_name, &node_shard.endpoint)?;
+                        self.cache.insert(table_name.clone(), route.clone()).await;
```
chunshao90, line 93: There may be data race here, add a comment description.

router/src/lib.rs
MachaelLee: > 

This will make the unit test difficult to write.

router/src/cluster_based.rs
MachaelLee: > 

This will make the unit test difficult to write.

router/src/cluster_based.rs
```diff
@@ -35,31 +66,86 @@ fn make_route(table_name: &str, endpoint: &str) -> Result<Route> {
...
+        let cache: Cache<String, Route> = Cache::builder()
+            .time_to_live(Duration::from_secs(4))
+            .time_to_idle(Duration::from_secs(2))
+            .max_capacity(2)
+            .build();
+
+        let tables = vec!["table1".to_string(), "table2".to_string()];
+
+        // first case get two tables, miss 0
+        let route = make_route("table0", "127.0.0.0:8831").unwrap();
```
chunshao90, line 122: Can you define some variables?

router/src/cluster_based.rs
chunshao90: The testing for this function may be limited and it feels like it is testing the functionality of Moka.

router/src/cluster_based.rs
```diff
@@ -35,31 +66,86 @@ fn make_route(table_name: &str, endpoint: &str) -> Result<Route> {
...
-        for (table_name, route_entry) in route_resp.entries {
-            for node_shard in route_entry.node_shards {
-                if node_shard.shard_info.is_leader() {
-                    let route = make_route(&table_name, &node_shard.endpoint)?;
-                    routes.push(route);
+
+        // Firstly route table from local cache.
+        let (mut routes, miss) = route_from_cache(&self.cache, req.tables);
+
+        if !miss.is_empty() {
```
chunshao90, line 73: Remove this level.
```
if miss.is_empty(){
return Ok(routes)
}

...
```

router/src/lib.rs
```diff
@@ -63,3 +63,15 @@ pub type RouterRef = Arc<dyn Router + Sync + Send>;
 pub trait Router {
     async fn route(&self, req: RouteRequest) -> Result<Vec<Route>>;
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize, Default)]
+pub struct RouteCacheConfig {
+    // enable route cache, default false
+    enable: bool,
```
ShiKaiWi, line 70: I guess the cache should be always enabled.

router/src/lib.rs
```diff
@@ -63,3 +63,15 @@ pub type RouterRef = Arc<dyn Router + Sync + Send>;
...
+    /// Time to live (TTL) in second.
+    ttl: u64,
```
ShiKaiWi, line 72: The type of `ttl`  and `tti` should be `ReadableDuration`.

router/src/lib.rs
MachaelLee: I guess route cache is not necessary, and route cache may return wrong route endpoint. user can enable it if they need.

router/src/cluster_based.rs
```diff
@@ -48,18 +92,170 @@ impl Router for ClusterBasedRouter {
...
+    };
+
+    use super::*;
+
+    struct MockClusterImpl {}
+
+    #[async_trait]
+    impl Cluster for MockClusterImpl {
+        async fn start(&self) -> cluster::Result<()> {
+            todo!()
```
ShiKaiWi, line 138: Maybe we should use `unimplemeted!` here.

server/src/config.rs
```diff
@@ -125,6 +128,7 @@ impl Default for ServerConfig {
             resp_compress_min_length: ReadableSize::mb(4),
             forward: forward::Config::default(),
             auto_create_table: true,
+            route_cache: router::RouteCacheConfig::default(),
```
ShiKaiWi, line 131: Maybe we should provide default implementation for the RouteCacheConfig?

router/src/cluster_based.rs
```diff
@@ -7,17 +7,52 @@ use ceresdbproto::storage::{Route, RouteRequest};
...
+                    .build(),
+            )
+        } else {
+            None
+        };
+
+        Self { cluster, cache }
+    }
+
+    /// route table from local cache, return cache routes and  tables which are
```
ShiKaiWi, line 37:
```suggestion
    /// route table from local cache, return cache routes and tables which are
```

ShiKaiWi (approved):
Good Job!

PR #749, opened 2023-03-20
feat: auto add column

### LINKED ISSUE #634: Supports automatic field creation
### Describe This Problem

When ceresdb writes data for the first time, it supports automatic table creation, but when the data written next contains new fields, it does not support automatic creation of new fields. It is hoped that new fields can be automatically created in this scenario

### Proposal

Supports automatic field creation

### Additional Context

_No response_

chunshao90:

Closes #634 

## Rationale for this change
 Refer to #634 

* Support auto create new columns.
* Add auto create new columns in test of golang sdk.

## How does this change test
Add auto create new columns in test of golang sdk.


server/src/proxy/grpc/write.rs
```diff
@@ -167,26 +169,44 @@ pub async fn write_request_to_insert_plan<Q: QueryExecutor + 'static>(
...
+                        &write_table_req,
+                        &schema_config,
+                        deadline,
+                    )
+                    .await?;
+                    // try to get table again
+                    table = try_get_table(&catalog, &schema, instance.clone(), table_name)?;
+                }
+            }
+            Some(t) => {
```
jiacai2050, line 194: We should only alter schema when write failed with specific error code.

Current changes will be invoked at every write(hot path), which is not ideal way to do this.

server/src/proxy/grpc/write.rs
```diff
@@ -207,7 +227,7 @@ pub async fn write_request_to_insert_plan<Q: QueryExecutor + 'static>(
     Ok(plan_vec)
 }
 
-pub async fn execute_plan<Q: QueryExecutor + 'static>(
+pub async fn execute_insert_plan<Q: QueryExecutor + 'static>(
```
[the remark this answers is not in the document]
chunshao90: I extracted the execute_plan method to reuse some duplicate code.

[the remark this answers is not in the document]
chunshao90: Currently, only the write module has been modified, and the other parts of the proxy will be refactored using execute_plan later.

server/src/proxy/grpc/write.rs
```diff
@@ -576,6 +527,98 @@ fn convert_proto_value_to_datum(
...
+        .map_err(|e| {
+            error!("Add colum failed, err:{e}");
+            e
+        })?;
+
+    info!("Add columns success, request_id:{request_id}, table:{table_name}");
+    Ok(())
+}
+
+async fn execute_plan<Q: QueryExecutor + 'static>(
```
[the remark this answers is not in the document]
chunshao90: The functions are different. execute_insert_plan will call execute_plan.

server/src/proxy/grpc/write.rs
chunshao90: * If judged based on the error message returned, it may cause data that has already been successfully written to be written again and affect the accuracy of the data.
* Currently, the decision to add columns is made at the request level, not at the row level, so the cost is relatively small.

server/src/proxy/grpc/write.rs
```diff
@@ -576,6 +537,98 @@ fn convert_proto_value_to_datum(
...
+        .context(ErrWithCause {
+            code: StatusCode::INTERNAL_SERVER_ERROR,
+            msg: "Build schema from write table request failed",
+        })?;
+
+    let columns = new_schema.columns();
+    let old_columns = schema.columns();
+    let mut new_columns = Vec::new();
+
+    columns.iter().for_each(|column| {
```
jiacai2050, line 556:
```
columns.iter().filter(|col|).collect::<Vec<_>>()
```

server/src/proxy/grpc/write.rs
```diff
@@ -576,6 +537,98 @@ fn convert_proto_value_to_datum(
...
+    info!("Add columns start, request_id:{request_id}, table:{table_name}, columns:{columns:?}");
+
+    let plan = Plan::AlterTable(AlterTablePlan {
+        table,
+        operations: AlterTableOperation::AddColumn(columns),
+    });
+    let _ = execute_plan(request_id, catalog, schema, instance, plan, deadline)
+        .await
+        .map_err(|e| {
+            error!("Add colum failed, err:{e}");
```
jiacai2050, line 583: This error log seems not necessary, you can use with_context to add info you want.

jiacai2050 (approved):
LGTM

PR #751, opened 2023-03-20
feat: block all query requests

MachaelLee:

Closes #

## Rationale for this change

Sometimes, I may want CeresDB to block all query and write requests. For instance, if CeresDB is experiencing a memory leak, I might want to investigate whether queries cause this. In such situations, blocking all requests can be useful.

* add a block rule to block all query request
* add a block rule to block all write request

## How does this change test

By unit test


ShiKaiWi (approved):
LGTM

PR #754, opened 2023-03-22
feat: support split write request to batches for small wal logs

ShiKaiWi:

Closes #

## Rationale for this change
 Currently, one large write request may lead to one large WAL log, which is not friendly for some wal implementations.

- Support split write table request to multiple batches.

None.

## How does this change test
New unit tests and existing integraion tests.

### DIFF
--- a/analytic_engine/src/instance/write.rs
+++ b/analytic_engine/src/instance/write.rs
@@ -183,7 +183,7 @@ impl WriteRowGroupSplitter {
     /// NOTE: The length of the `encoded_rows` should be the same as the number
     /// of rows in the `row_group`.
     pub fn split<'a>(
-        &'a self,
+        &'_ self,
         encoded_rows: Vec<ByteVec>,
         row_group: &'a RowGroup,
     ) -> SplitResult<'a> {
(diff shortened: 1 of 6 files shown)


analytic_engine/src/instance/write.rs
```diff
@@ -150,6 +150,107 @@ impl EncodeContext {
...
+    /// `max_bytes_per_batch`.
+    max_bytes_per_batch: usize,
+}
+
+enum SplitResult<'a> {
+    Splitted {
+        encoded_batches: Vec<Vec<ByteVec>>,
+        row_group_batches: Vec<RowGroupSlicer<'a>>,
+    },
+    Integrate {
```
jiacai2050, line 168: It seems this item could be implemented using Splitted with len is 1?
```
struct SplitResult {
  encoded_batches: Vec<Vec<ByteVec>>,
  row_group_batches: Vec<RowGroupSlicer<'a>>,
}
```

ShiKaiWi: This will lead to extra cost on allocating a vector if no split happens.

jiacai2050: That's one small point, I'm not strong with this, just want to remove some duplicated code below.

jiacai2050 (approved):
LGTM

PR #755, opened 2023-03-22
fix: move table engine proxy to table_engine crate

Rachelint:

Closes #

## Rationale for this change
Now, rust-analyzer will report cyclic dependecy like:
```
[ERROR project_model::workspace] cyclic deps: server(CrateId(478)) -> interpreters(CrateId(256)), alternative path: interpreters(CrateId(256)) -> catalog_impls(CrateId(97)) -> server(CrateId(478))
``` 
We should fix it.

Move table engine proxy to table_engine crate.

Rust-analyzer will not compare cyclic dependency now.

## How does this change test
Test by ut.

### DIFF
--- a/server/src/handlers/influxdb.rs
+++ b/server/src/handlers/influxdb.rs
@@ -226,8 +226,6 @@ fn convert_influx_value(field_value: FieldValue) -> Value {
     Value { value: Some(v) }
 }
 
-// fn convert_query_result(output: Output)
-
 // TODO: Request and response type don't match influxdb's API now.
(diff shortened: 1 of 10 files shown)


server/src/handlers/influxdb.rs
```diff
@@ -226,6 +226,8 @@ fn convert_influx_value(field_value: FieldValue) -> Value {
     Value { value: Some(v) }
 }
 
+// fn convert_query_result(output: Output)
```
ShiKaiWi, line 229: What is this?

ShiKaiWi (approved):

PR #756, opened 2023-03-23
chore: fix insert-license pre-commit

jiacai2050:

Closes #

Update license insert, add `use-current-year`

- When edit old file, pre-commit will update license header to `2022-{current-year}`

## How does this change test

Manually.

This is what other project does:
- https://github.com/facebook/rocksdb/blob/main/table/format.cc
- https://github.com/tikv/tikv/blob/master/src/coprocessor_v2/raw_storage_impl.rs

It seems year format is not very strict, and we can define format used in CeresDB.


ShiKaiWi (approved):
LGTM

PR #757, opened 2023-03-23
chore: use influxdb line protocol in crates

jiacai2050:

Closes #

## Rationale for this change
 
Influxdb line protocol is published at crates.io, no need to use forked version.

No

## How does this change test

CI.


ShiKaiWi (approved):
LGTM

PR #759, opened 2023-03-23
feat: configurable record batches in flight

### LINKED ISSUE #735: Tracking issue: streaming building/reading SST
### Describe This Problem

Current building/reading SST is not in a streaming way, that is to say, all the content will be buffered in the memory before dumping into the underlying storage for building procedure and all the content will be fetched into memory before feeding to the upside merge iterators. Such way to bulid/read sst may lead to high memory consumption.

### Proposal

- [x] Write SST in a stream way #486
- [ ] Read SST in a stream way.

### Additional Context

ShiKaiWi:

Closes #735

## Rationale for this change
Currently, streaming reading sst has already been supported. However, the max record batches in flight is not configurable, which is hard-coded as 1024. This pr makes it configurable, especially for normal query and compaction reading.

- Make `scan_max_record_batches_in_flight` configurable;
- Refactor the implementation for Stream based on a future method;

A new config options is provided called `scan_max_record_batches_in_flight.

## How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/instance/read.rs
+++ b/analytic_engine/src/instance/read.rs
@@ -9,12 +9,14 @@ use std::{
 };
 
 use common_types::{
-    projected_schema::ProjectedSchema, record_batch::RecordBatch, schema::RecordSchema,
+    projected_schema::ProjectedSchema,
+    record_batch::{RecordBatch, RecordBatchWithKey},
+    schema::RecordSchema,
     time::TimeRange,
 };
-use common_util::{define_result, error::BoxError, runtime::Runtime};
+use common_util::{define_result, error::BoxError};
 use futures::stream::Stream;
-use log::{debug, error, trace};
+use log::debug;
 use snafu::{ResultExt, Snafu};
 use table_engine::{
     stream::{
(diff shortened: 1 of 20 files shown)


analytic_engine/src/instance/read.rs
```diff
@@ -331,72 +326,109 @@ impl Instance {
...
+            if let Some(v) = iter.next_batch().await.transpose() {
+                return Some(v);
             }
+
+            self.advance();
         }
+    }
+}
+
+fn iters_to_stream(
```
jiacai2050, line 374: In your new implementation, no channel is used as buffer, does this will slow down query when iters are among multiple sst segments?

ShiKaiWi: The channel here was not used as buffer, but used as a hack way to implement stream. Current implementation will perform better because the overhead caused by the channel is avoided.

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -110,6 +111,10 @@ impl RecordBatchGroupWriter {
                 continue;
             }
 
+            if self.input_exhausted {
```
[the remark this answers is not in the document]
ShiKaiWi: Following the last batch is fetched, another fetch call will still poll the exhausted `input` stream, and it will cause panic.

jiacai2050 (approved):
LGTM

PR #763, opened 2023-03-24
chore: replace unfold with async-stream

chunshao90:

Closes #

## Rationale for this change
Replace unfold with async-stream.

Replace unfold with async-stream.

No.

## How does this change test
Integration test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,6 +81,7 @@ dependencies = [
  "arc-swap 1.5.1",
  "arena",
  "arrow 32.0.0",
+ "async-stream",
  "async-trait",
  "base64 0.13.0",
  "bytes 1.4.0",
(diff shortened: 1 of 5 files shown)


jiacai2050:
```
{
    let (mut __yield_tx, __yield_rx) = unsafe { $crate::__private::yielder::pair() };
    $crate::__private::AsyncStream::new(__yield_rx, async move {
        while let Some(value) = state.fetch_next_batch().await {
            let record_batch = value
                .box_err()
                .context(ErrWithSource {
                    msg: "Read record batch",
                })
                .and_then(|batch_with_key| {
                    batch_with_key
                        .try_project(&state.projected_schema)
                        .box_err()
                        .context(ErrWithSource {
                            msg: "Project record batch",
                        })
                });
            __yield_tx
                .send(::core::result::Result::Ok(match record_batch {
                    ::core::result::Result::Ok(v) => v,
                    ::core::result::Result::Err(e) => {
                        __yield_tx.send(::core::result::Result::Err(e.into())).await;
                        return;
                    }
                }))
                .await;
        }
    })
}

```
This is what `try_stream` does, we need to  check if it has some side-effect.

ShiKaiWi (approved):
LGTM

PR #767, opened 2023-03-25
fix: rename as_bytes to as_byte in ReadableSize (#428)

### LINKED ISSUE #428: Bad name in ReadableSize
### Describe this problem

https://github.com/CeresDB/ceresdb/blob/659b55985a14f835cff5bbb2df4eea1b0a540366/common_util/src/config.rs#L149

In `ReadableSize`, there is a method called `as_bytes()`, which is very confusing, since the original meaning of `as_bytes` is to return `&[u8]`, not `u64`

### Steps to reproduce

No required.

### Expected behavior

_No response_

### Additional Information

I suggest we name it `as_byte` to follow other methods in ReadableSize

zouxiang1993:

Closes #428

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -355,26 +355,29 @@ impl<Q: QueryExecutor + 'static> Builder<Q> {
             .context(BuildMysqlService)?;
 
         let router = self.router.context(MissingRouter)?;
-        let rpc_services = grpc::Builder::new()
-            .endpoint(
-                Endpoint::new(self.server_config.bind_addr, self.server_config.grpc_port)
-                    .to_string(),
-            )
-            .local_endpoint(Endpoint::new(self.node_addr, self.server_config.grpc_port).to_string())
-            .resp_compress_min_length(
-                self.server_config.resp_compress_min_length.as_byte() as usize
-            )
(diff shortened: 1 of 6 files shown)


ShiKaiWi:
@zouxiang1993 Thanks for your contribution. It looks pretty and will be merged when the status check finishes.

server/src/server.rs
```diff
@@ -355,26 +355,29 @@ impl<Q: QueryExecutor + 'static> Builder<Q> {
...
-            .cluster(self.cluster.clone())
-            .opened_wals(opened_wals)
-            .schema_config_provider(provider)
-            .forward_config(self.server_config.forward)
-            .timeout(self.server_config.timeout.map(|v| v.0))
-            .auto_create_table(self.server_config.auto_create_table)
-            .build()
-            .context(BuildGrpcService)?;
+        let rpc_services =
+            grpc::Builder::new()
```
chunshao90, line 359: No need to change here?

zouxiang1993: "cargo fmt" do it.

ShiKaiWi (approved):
LGTM

PR #782, opened 2023-03-28
feat: add default schema config

jiacai2050:

Closes #

## Rationale for this change
 
Simplify schema config setting. If we don't provide a define one, we need to set schema config one by one, something like

```
[[cluster_deployment.topology.schema_shards]]
schema = 'public_0'
default_timestamp_column_name = 'time'
[[cluster_deployment.topology.schema_shards.shard_views]]
shard_id = 0
[cluster_deployment.topology.schema_shards.shard_views.endpoint]
addr = '127.0.0.1'
port = 8831
[[cluster_deployment.topology.schema_shards.shard_views]]
shard_id = 1
[cluster_deployment.topology.schema_shards.shard_views.endpoint]
addr = '127.0.0.1'
port = 8831
```

This is very inconvient.

- Add default schema config in server

```toml
[server.default_schema_config]
default_timestamp_column_name = "time"
```
- Update bind_addr to 127.0.0.1, which is more safe than before.

No

## How does this change test

Manully

### DIFF
--- a/cluster/src/config.rs
+++ b/cluster/src/config.rs
@@ -5,7 +5,7 @@ use meta_client::meta_impl::MetaClientConfig;
 use serde::{Deserialize, Serialize};
 use table_engine::ANALYTIC_ENGINE_TYPE;
 
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


Rachelint (approved):
LGTM

PR #790, opened 2023-03-29
fix: avoid error when disk cache miss

### LINKED ISSUE #769: Buffer underflow when executing query
### Describe this problem

Error message is:
```text
{"code":500,"message":"Failed to handle request, err:Failed to execute interpreter, query:select count(*) from perflog where namespace=\"grpc.server.cost\" and subtag=\"video-conductor-proxy-service.task-service/BatchAckTask\" limit 10, err:Failed to execute select, err:Failed to execute logical plan, err:Failed to collect record batch stream, err:Stream error, msg:convert from arrow record batch, err:External error: Stream error, msg:poll read response failed, err:Failed to query from table in server, table_ident:TableIdentifier { catalog: \"ceresdb\", schema: \"public\", table: \"__perflog_9\" }, code:500, msg:record batch failed. Caused by: Stream error, msg:read record batch, err:Failed to read data from the sub iterator, err:PullRecordBatch { source: DecodeRecordBatch { source: ParquetError { source: General(\"Failed to fetch ranges from object store, err:Generic DiskCacheStore error: Failed to decode cache pb value, file:/data/ceresdb/sst_cache/0-14-357.sst-0-4194304, source:failed to decode Protobuf message: Bytes.value: buffer underflow.\\nbacktrace:\\n 0 <snafu::backtrace_shim::"}
```

### Server version

CeresDB Server 
CeresDB version: 1.0.0
Git branch: main
Git commit: 95ea870
Build time: 2023-03-06T03:30:09.845055021Z
Rustc version: 1.69.0-nightly

### Steps to reproduce

None

### Expected behavior

_No response_

### Additional Information

_No response_

ShiKaiWi:

Closes #769 

## Rationale for this change
 Error will be thrown breaking execution, when data in the disk cache is corrupted, which is unavoidable.

- Report cache miss when cache content is corrupted instead.

None.

## How does this change test
The new unit test for it is added.

### DIFF
--- a/components/object_store/src/disk_cache.rs
+++ b/components/object_store/src/disk_cache.rs
@@ -129,8 +129,11 @@ impl DiskCache {
         }
     }
 
+    /// Update the cache.
+    ///
+    /// The returned value denotes whether succeed.
     // TODO: We now hold lock when doing IO, possible to release it?
-    async fn update_cache(&self, key: String, value: Option<Bytes>) -> Result<()> {
+    async fn update_cache(&self, key: String, value: Option<Bytes>) -> bool {
         let mut cache = self.cache.lock().await;
         debug!(
             "Disk cache update, key:{}, len:{}, cap:{}.",
@@ -139,6 +142,7 @@ impl DiskCache {
             self.cap
         );
 
+        // TODO: remove a batch of files to avoid IO during the following update cache.
         if cache.len() >= self.cap {
             let (filename, _) = cache.pop_lru().unwrap();
             let file_path = std::path::Path::new(&self.root_dir)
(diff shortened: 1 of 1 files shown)


components/object_store/src/disk_cache.rs
```diff
@@ -169,14 +169,23 @@ impl DiskCache {
         self.update_cache(filename, None).await
     }
 
-    async fn get(&self, key: &str) -> Result<Option<Bytes>> {
+    async fn get(&self, key: &str) -> Option<Bytes> {
```
jiacai2050, line 172: Do we need to ignore error for other methods, like `insert`, `recover`?

ShiKaiWi: Yes. It should be.

jiacai2050 (approved):
lgtm

PR #791, opened 2023-03-29
feat: add rust sdk test

Rachelint:

Closes #

## Rationale for this change
 Add integration test for rust sdk.

see title.

None.

## How does this change test
None.

### DIFF
--- a/integration_tests/sdk/rust/src/main.rs
+++ b/integration_tests/sdk/rust/src/main.rs
@@ -81,10 +81,28 @@ async fn sql_query(
     test_data: &TestDatas,
     new_column: bool,
 ) {
+    let all_columns = test_data.col_names.clone();
+    let selections = if !new_column {
+        format!(
+            "`{}`,`{}`,`{}`,`{}`",
+            all_columns[0], all_columns[1], all_columns[2], all_columns[3]
+        )
+    } else {
+        format!(
+            "`{}`,`{}`,`{}`,`{}`,`{}`,`{}`",
+            all_columns[0],
+            all_columns[1],
+            all_columns[2],
+            all_columns[3],
+            all_columns[4],
+            all_columns[5]
+        )
+    };
+
     let test_table = format!("test_table_{timestamp}");
     let query_req = SqlQueryRequest {
         tables: vec![test_table.clone()],
-        sql: format!("SELECT * from {test_table}"),
+        sql: format!("SELECT {selections} from {test_table}"),
     };
     let resp = client.sql_query(rpc_ctx, &query_req).await.unwrap();
     assert_eq!(resp.affected_rows, 0);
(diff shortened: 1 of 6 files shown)


integration_tests/sdk/rust/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    let rows = if !new_column {
+        test_data.rows.iter().take(2).cloned().collect::<Vec<_>>()
+    } else {
+        vec![test_data.rows[2].clone(), test_data.rows[3].clone()]
+    };
+
+    for row in rows {
+        let point = {
+            let timestamp_val = match &row[0] {
```
jiacai2050, line 139: Isn't this same with timestamp in L123?

ShiKaiWi (approved):
LGTM

integration_tests/sdk/rust/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+    let col_names = vec![
+        "timestamp".to_string(),
+        "old-tag".to_string(),
+        "old-field0".to_string(),
+        "old-field1".to_string(),
+        "new-tag".to_string(),
+        "new-field".to_string(),
+    ];
+
+    let rows = vec![
```
jiacai2050, line 178: It's wired to put all those together. I suggest you can add two fields in TestData
```
{
  OriginRows: Vec<Row>,
  RowsWithAddedColumn: Vec<Row>
}
```

integration_tests/sdk/rust/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+        .unwrap_or(0)
+}
+
+fn extract_rows_from_sql_query(resp: &SqlQueryResponse) -> Vec<Vec<(String, Value)>> {
+    let mut rows = Vec::with_capacity(resp.rows.len());
+    for row in &resp.rows {
+        let mut column_iter = row.columns().iter();
+        // In the automatically created table schema, `tsid` column will be added by
+        // CeresDB, we just ignore it.
+        column_iter.next();
```
jiacai2050, line 229: If you don't want tsid, you can select what you want, not  just `select *`.

integration_tests/sdk/rust/src/main.rs
Rachelint: I want to generate test data together.
And we just process them for different usage?
Because for `test_add_column` case, we just need to write some new records like:
```
ts, old-tag3, old-field3, new-tag1, new-field1
ts, old-tag4, old-field4, new-tag2, new-field2
```
But when we query, old records with added colmns and new records will be returned and we should check, like:
```
ts, old-tag1, old-field1, default-new-tag, default-new-field
ts, old-tag2, old-field2, default-new-tag, default-new-field
ts, old-tag3, old-field3, new-tag1, new-field1
ts, old-tag4, old-field4, new-tag2, new-field2
```
Even though we split it into two fields, we need to process too.

jiacai2050 (approved):

PR #793, opened 2023-03-30
feat: impl getTableInfo in remoteEngine service

chunshao90:

Closes #

## Rationale for this change
The schema of partition table will be obtained from the first sub-table, so a remote engine is needed to implement the table information access interface.

Impl getTableInfo in remoteEngine service.

No.

## How does this change test
Testing will be conducted with the partition table, and it will be used internally by the engine without affecting external users for the time being.

### DIFF
--- a/common_types/src/schema.rs
+++ b/common_types/src/schema.rs
@@ -1,4 +1,4 @@
-// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+// Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
 
 //! Schema of table
 
@@ -987,6 +987,12 @@ impl From<&Schema> for schema_pb::TableSchema {
             columns,
             primary_key_ids,
         }
+    }
+}
+
+impl From<Schema> for schema_pb::TableSchema {
+    fn from(value: Schema) -> Self {
+        (&value).into()
     }
 }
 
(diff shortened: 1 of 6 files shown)


common_types/src/schema.rs
```diff
@@ -990,6 +990,12 @@ impl From<&Schema> for schema_pb::TableSchema {
     }
 }
 
+impl From<Schema> for schema_pb::TableSchema{
+    fn from(value: Schema) -> Self {
+        (&value).into()
```
ShiKaiWi, line 995: It will be better to avoid clone here.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -123,6 +124,39 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
         Ok(Response::new(resp))
     }
 
+    async fn get_table_info_internal(
+        &self,
+        request: Request<GetTableInfoRequest>,
+    ) -> std::result::Result<Response<GetTableInfoResponse>, Status> {
+        let begin_instant = Instant::now();
+        let ctx = self.handler_ctx();
+        let handle = self.runtimes.meta_runtime.spawn(async move {
```
ShiKaiWi, line 133: Why use `meta_runtime` rather than `read_runtime`?

ShiKaiWi (approved):
LGTM

PR #794, opened 2023-03-30
chore: route use read_runtime

chunshao90:

Closes #

## Rationale for this change
 Route in grpc service use read_runtime.

Route in grpc service use read_runtime.

No.

## How does this change test
No need.


ShiKaiWi (approved):
LGTM

PR #795, opened 2023-03-30
feat: influxql support show measurements

jiacai2050:

Closes #

## Rationale for this change
 
`show measurements` is used by grafana to check if influxdb is working, so add this support.

image

- Add show measurements support

No

## How does this change test

CI


server/src/handlers/influxdb.rs
```diff
@@ -311,6 +311,8 @@ impl InfluxqlResultBuilder {
             }
         );
 
+        // Query like `show measurements`, there will be not timestamp.
+        let has_timestamp = column_schemas.iter().any(|c| c.data_type.is_timestamp());
```
Rachelint, line 315: Name it `is_select`? We just use the existence of `timestamp` to get the query type?

Rachelint (approved):
LGTM

server/src/handlers/influxdb.rs
jiacai2050: Not make much sense.

PR #796, opened 2023-03-30
fix: avoid send empty record batch to client

ShiKaiWi:

Closes #

## Rationale for this change
 Currently, empty record batch will be generated by the datafusion, and will be sent to the client too. Avoid the empty record batch because it is a waste for both server and client.

- Avoid encoding the empty record batch into the `SqlQueryResponse`;

1. The client won't receive any empty record batch;
2. The schema information won't be sent to the client if the whole query response is empty;

As for the second cheanges, all the sdks has been tested for the compatibility.

## How does this change test
Existing tests.


chunshao90 (approved):
LGTM

PR #807, opened 2023-04-05
feat: http query with proxy

chunshao90:

Closes #

## Rationale for this change
 Support forward http sql query.

* Impl http sql query with proxy.
* Forward http sql query.

Support forward http sql query.

## How does this change test
Manual testing and CI.


server/src/proxy/http/query.rs
```diff
@@ -0,0 +1,323 @@
...
+    fn from(query: String) -> Self {
+        Self { query }
+    }
+}
+
+impl From<Bytes> for Request {
+    fn from(bytes: Bytes) -> Self {
+        Request::from(String::from_utf8_lossy(&bytes).to_string())
+    }
+}
```
ShiKaiWi, line 263: The two conversions seems weird, how about removing them?

sql/src/frontend.rs
```diff
@@ -277,8 +285,7 @@ mod tests {
                             format!("exists table `{table}`"),
         ];
         for sql in test_cases {
-            let statements = Parser::parse_sql(&sql).unwrap();
-            assert_eq!(parse_table_name(&statements), Some(table.to_string()));
+            assert_eq!(parse_table_name_with_sql(&sql), Some(table.to_string()));
```
ShiKaiWi, line 288:
```suggestion
            assert_eq!(frontend::parse_table_name_with_sql(&sql), Some(table.to_string()));
```

server/src/proxy/util.rs
```diff
@@ -1,11 +1,104 @@
...
+    },
 };
+use common_types::record_batch::RecordBatch;
+use common_util::error::BoxError;
+use interpreters::interpreter::Output;
+use snafu::{OptionExt, ResultExt};
+
+use crate::proxy::error::{Internal, InternalNoCause, Result};
+
+pub fn convert_sql_response_to_output(sql_query_response: SqlQueryResponse) -> Result<Output> {
```
ShiKaiWi, line 25: Shall we define this method just at the file using it?

sql/src/frontend.rs
```diff
@@ -247,9 +247,17 @@ pub fn parse_table_name(statements: &StatementVec) -> Option<String> {
     }
 }
 
+pub fn parse_table_name_with_sql(sql: &str) -> Option<String> {
+    if let Ok(v) = Parser::parse_sql(sql) {
```
ShiKaiWi, line 251: Why not just throw the error directly?

ShiKaiWi (approved):
LGTM

PR #812, opened 2023-04-07
feat: expose rocksdb background jobs option

jiacai2050:

Closes #

## Rationale for this change
 
Max background jobs is hard-code, expose an option to configure this.

- Add new options for background jobs.
- Add new http endpoint for debugging `GET /debug/stats`, this require following config:
```
[analytic.wal.data_namespace]
enable_statistics = true
max_background_jobs = 8
```

image

No

## How does this change test

No need.

### DIFF
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -30,7 +30,7 @@ use storage_options::StorageOptions;
 use table_kv::config::ObkvConfig;
 use wal::{
     message_queue_impl::config::Config as MessageQueueWalConfig,
-    table_kv_impl::model::NamespaceConfig,
+    rocks_impl::config::Config as RocksDBWalConfig, table_kv_impl::model::NamespaceConfig,
 };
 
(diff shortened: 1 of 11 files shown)


analytic_engine/src/lib.rs
```diff
@@ -254,12 +254,18 @@ pub struct KafkaWalConfig {
...
 }
 
 impl Default for RocksDBConfig {
     fn default() -> Self {
         Self {
             data_dir: "/tmp/ceresdb".to_string(),
+            // Same with rocksdb
+            // https://github.com/facebook/rocksdb/blob/v6.4.6/include/rocksdb/options.h#L537
+            sst_max_background_jobs: 2,
+            manifest_max_background_jobs: 2,
```
ShiKaiWi, line 268: I guess we should provide two separate config fields for the `data wal` and `manifest wal`, otherwise we have to prefix every options for these two wals in the future.

ShiKaiWi (approved):
LGTM

Issue #800, opened 2023-03-31
Tracking Issue: accelerate recover speed


ShiKaiWi:
### Describe This Problem

Current mechanism of recovering when starting up costs too much time when the number of tables belonging to the ceresdb server instance is large. And there are two known reasons accounting for the slow recovering:
- Recovering work is processed table by table, but data is actually organized by shard;
- Recovering work is handled in the same runtime with write work, that is to say, recovering work has intensive contention with the writing work.

### Proposal

Here is a simple version of the proposal to resolve the two problems:
- [ ] Recover data shard by shard. #799 
- [x] Recover data in a separate runtime from write work. Solved by #814

PR #814, opened 2023-04-09
feat: serialize table operations by lock rather than queue

ShiKaiWi:

Prepare for #800 

## Rationale for this change
Currently, the operations on tables are serialized by the worker queue, which has a serious disadvantage -- no work steal, which may leading to hungry of some tables' operations because of one big table.

This PR will replace the worker queue with a simple lock for every table.

- Replace the worker queue with lock for every table;

Remove the configs for the write workers.

## How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/instance/open.rs
+++ b/analytic_engine/src/instance/open.rs
@@ -25,7 +25,7 @@ use crate::{
         engine::{ApplyMemTable, FlushTable, ReadMetaUpdate, ReadWal, RecoverTableData, Result},
         flush_compaction::TableFlushOptions,
         mem_collector::MemUsageCollector,
-        serializer::TableOpSerializer,
+        serial_executor::TableOpSerialExecutor,
         write::MemTableWriter,
         Instance, SpaceStore, Spaces,
     },
@@ -290,7 +290,7 @@ impl Instance {
             .await
             .context(ReadWal)?;
 
-        let mut serializer = table_data.serializer.lock().await;
+        let mut serial_exec = table_data.serial_exec.lock().await;
         let mut log_entry_buf = VecDeque::with_capacity(replay_batch_size);
         loop {
             // fetch entries to log_entry_buf
@@ -301,7 +301,7 @@ impl Instance {
                 .context(ReadWal)?;
 
             // Replay all log entries of current table
-            self.replay_table_log_entries(&mut serializer, &table_data, &log_entry_buf)
+            self.replay_table_log_entries(&mut serial_exec, &table_data, &log_entry_buf)
                 .await?;
 
             // No more entries.
@@ -316,7 +316,7 @@ impl Instance {
     /// Replay all log entries into memtable and flush if necessary.
     async fn replay_table_log_entries(
         self: &Arc<Self>,
-        serializer: &mut TableOpSerializer,
+        serial_exec: &mut TableOpSerialExecutor,
         table_data: &TableDataRef,
         log_entries: &VecDeque<LogEntry<ReadPayload>>,
     ) -> Result<()> {
@@ -374,7 +374,7 @@ impl Instance {
 
                     let index_in_writer =
                         IndexInWriterSchema::for_same_schema(row_group.schema().num_columns());
-                    let memtable_writer = MemTableWriter::new(table_data.clone(), serializer);
+                    let memtable_writer = MemTableWriter::new(table_data.clone(), serial_exec);
                     memtable_writer
                         .write(sequence, &row_group.into(), index_in_writer)
                         .context(ApplyMemTable {
@@ -390,7 +390,7 @@ impl Instance {
                             compact_after_flush: None,
                         };
                         let flusher = self.make_flusher();
-                        let flush_scheduler = serializer.flush_scheduler();
+                        let flush_scheduler = serial_exec.flush_scheduler();
                         flusher
                             .schedule_flush(flush_scheduler, table_data, opts)
                             .await
--- a/analytic_engine/src/instance/open.rs
+++ b/analytic_engine/src/instance/open.rs
@@ -8,10 +8,9 @@ use std::{
 };
 
 use common_types::schema::IndexInWriterSchema;
-use log::{debug, error, info, trace, warn};
+use log::{debug, error, info, trace};
 use snafu::ResultExt;
 use table_engine::engine::OpenTableRequest;
-use tokio::sync::oneshot;
 use wal::{
     log_batch::LogEntry,
     manager::{ReadBoundary, ReadContext, ReadRequest, WalManagerRef},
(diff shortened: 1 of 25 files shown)


analytic_engine/src/instance/serializer.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub async fn flush_sequentially<F, T>(
+        &mut self,
+        table: String,
+        metrics: &Metrics,
+        flush_job: F,
+        on_flush_success: T,
+        block_on_write_thread: bool,
+        runtime: &Runtime,
+        res_sender: Option<oneshot::Sender<TableResult<()>>>,
+    ) -> flush_compaction::Result<()>
```
Rachelint, line 83: How about solving this clippy complain?

analytic_engine/src/instance/serializer.rs
```diff
@@ -0,0 +1,175 @@
...
+        // If flush operation is running, then we need to wait for it to complete first.
+        // Actually, the loop waiting ensures the multiple flush procedures to be
+        // sequential, that is to say, at most one flush is being executed at
+        // the same time.
+        {
+            let mut stall_begin: Option<Instant> = None;
+            let mut flush_state = self.schedule_lock.0.lock().unwrap();
+            loop {
+                match &*flush_state {
+                    FlushState::Ok => {
```
Rachelint, line 97: How about FlushState::Ready?

analytic_engine/src/instance/serializer.rs
```diff
@@ -0,0 +1,175 @@
...
+        F: Future<Output = flush_compaction::Result<()>> + Send + 'static,
+        T: Future<Output = ()> + Send + 'static,
+    {
+        // If flush operation is running, then we need to wait for it to complete first.
+        // Actually, the loop waiting ensures the multiple flush procedures to be
+        // sequential, that is to say, at most one flush is being executed at
+        // the same time.
+        {
+            let mut stall_begin: Option<Instant> = None;
+            let mut flush_state = self.schedule_lock.0.lock().unwrap();
```
Rachelint, line 94: Maybe tokio::Mutex is more suitable here?
Because, we need to wait io of last flushing one?

analytic_engine/src/instance/serializer.rs
```diff
@@ -0,0 +1,175 @@
...
+enum FlushState {
+    #[default]
+    Ok,
+    Flushing,
+    Failed {
+        err_msg: String,
+    },
+}
+
+type ScheduleLock = Arc<(Mutex<FlushState>, Condvar)>;
```
Rachelint, line 29: How about 
```
struct {
    flush_state: Mutex<FlushState>,
    flush_notifier: Condvar,
}
```

analytic_engine/src/instance/open.rs
```diff
@@ -167,58 +156,21 @@ impl Instance {
...
-            replay_batch_size,
-            &read_ctx,
-        )
-        .await
-        .map_err(|e| {
-            error!("Recovery table from wal failed, table_data:{table_data:?}, err:{e}");
-            space.insert_open_failed_table(table_data.name.to_string());
-            e
-        })?;
+        self.recover_table_from_wal(table_data.clone(), self.replay_batch_size, &read_ctx)
```
MachaelLee, line 164: This result in  using meta runtime to open table, which may block heartbeat to meta.

analytic_engine/src/instance/write.rs
```diff
@@ -251,54 +241,128 @@ impl WriteRowGroupSplitter {
...
+                .context(UpdateMemTableSequence)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl<'a> Writer<'a> {
     /// Do the actual write, must called by write worker in write thread
     /// sequentially.
```
MachaelLee, line 356: update this comment?

analytic_engine/src/instance/open.rs
ShiKaiWi: I've changed the runtime of `meta_event_service` to `default_runtime`.

analytic_engine/src/table/mod.rs
```diff
@@ -114,9 +113,15 @@ impl Table for TableImpl {
     }
 
     async fn write(&self, request: WriteRequest) -> Result<usize> {
-        let num_rows = self
-            .instance
-            .write_to_table(&self.space_table, request)
+        let mut serializer = self.table_data.serializer.lock().await;
```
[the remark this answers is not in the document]
ShiKaiWi: Yes. I can't think of a better name than it.

analytic_engine/src/instance/serializer.rs
ShiKaiWi: The lock won't come across the await point and is only be held for a while, so let's keep using the std::sync::Mutex. However, the CondVar may block the current thread which is unacceptable for tokio runtime.

analytic_engine/src/instance/write.rs
```diff
@@ -251,56 +238,132 @@ impl WriteRowGroupSplitter {
...
+pub struct Writer<'a> {
+    instance: InstanceRef,
+    space: SpaceRef,
+    table_data: TableDataRef,
+    serializer: &'a mut TableOpSerializer,
+}
 
-        self.validate_before_write(space_table, &request)?;
+impl<'a> Writer<'a> {
+    pub fn make(
```
jiacai2050, line 249: Isn't this called `new` usually?

analytic_engine/src/instance/serializer.rs
ShiKaiWi: Good name.

analytic_engine/src/instance/serializer.rs
jiacai2050: Isn't a channel more simple than use a condition variable here?

Condvar is a lower-level API, I think we could avoid it when possible.

analytic_engine/src/instance/serializer.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+impl TableFlushScheduler {
+    /// Control the flush procedure and ensure multiple flush procedures to be
+    /// sequential.
+    ///
+    /// REQUIRE: should only be called by the write thread.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn flush_sequentially<F, T>(
+        &mut self,
```
jiacai2050, line 75: If this method is `&mut self`, then this method is sequential execution, no concurrency will happen, right?

analytic_engine/src/table/mod.rs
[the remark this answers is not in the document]
jiacai2050: If you hold a lock here, why underlying operation, such as flush still hold a lock?

analytic_engine/src/instance/serializer.rs
ShiKaiWi: I have modified it to `tokio::sync::watch::channel`.

analytic_engine/src/table/mod.rs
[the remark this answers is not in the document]
ShiKaiWi: Schedule flush with the lock held, but do flush without it.

analytic_engine/src/instance/serializer.rs
ShiKaiWi: Yes.

jiacai2050 (approved):
LGTM

PR #822, opened 2023-04-11
feat: avoid blocking the write procedure because of compaction schedule

ShiKaiWi:

Closes #

## Rationale for this change
Currently, compaction will be scheduled after every flush, and the compaction schedule will block if the schedule channel is full, leading to the flushing is blocked and write stall.

- Make compaction schedule won't block, and it will fail fast if the schedule channel is full.

None.

## How does this change test
Existing tests.

### DIFF
--- a/analytic_engine/src/instance/write_worker.rs
+++ b/analytic_engine/src/instance/write_worker.rs
@@ -1,4 +1,4 @@
-// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+// Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
 
 //! Write workers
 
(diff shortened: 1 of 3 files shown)


analytic_engine/src/instance/write_worker.rs
```diff
@@ -956,7 +956,12 @@ impl WriteWorker {
             waiter,
         };
 
-        self.instance.schedule_table_compaction(request).await;
+        let table = request.table_data.name.clone();
+        let succeed = self.instance.schedule_table_compaction(request).await;
+        if !succeed {
+            error!("Failed to schedule compaction, table:{table}");
```
jiacai2050, line 962: I think debug log is enough. otherwise it will log too much in write-heavy case.

ShiKaiWi: Here is triggered by manual compaction, I guess the message will seldom occur.

PR #824, opened 2023-04-11
feat: influxql support fill syntax

jiacai2050:

Closes #

## Rationale for this change
 
Fill syntax is part of influxql.

- Bump iox, which contains changes to support ms time unit
- Register GapFill datafusion extension

Yes, fill is supported, like this:
```sql
SELECT count(water_level) FROM "h2o_feet"
where time < 1439828400000ms
group by location, time(5m) fill(1)
```

And we can query influxql in grafana directly.

image

## How does this change test

New integration testcase

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -509,7 +509,7 @@ dependencies = [
 [[package]]
 name = "arrow_util"
 version = "0.1.0"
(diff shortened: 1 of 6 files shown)


Rachelint (approved):
LGTM

Issue #799, opened 2023-03-31
Shard based recovery


Rachelint:
### Describe This Problem

Now table recovery in on table level but wal's storing in on shard level.
The recovery performance may be unsatisfied especially in kafka based wal.  

### Proposal

- Split actual table recovery from schema, and refactor table engine
- Impl shard based table meta recovery
- Impl shard based table data recovery

#### 1. Split actual table recovery from schema, and refactor table engine
We should begin at modifying the high level interface(Schema and TableEngine) for adapting to the new recovery process.

Now the path about `Schema` and `TableEngine` when opening tables on shard is like:
<img src="https://raw.githubusercontent.com/Rachelint/drawio-store/main/now.drawio.svg?sanitize=true">

For modify interfaces above to open whole tables on shard together rather than respectively, the most troublesome place is: 
+ Tables on the same shard may belong to different schema, so we are unable to add the api like `open_tables_on_shard` to `Schema`.

My solution about this is: 
- [x]  Split actual table recovery from schema, we just call the `TableEngine` directly, and just register the opened tables to `Schema` #808 

In this stage, we still keep the origin interface of `TableEngine`, the path may be like:

<img src="https://raw.githubusercontent.com/Rachelint/drawio-store/main/stage1.drawio.svg?sanitize=true">

- [x]  Refactor the `TableEngine` interface to support shard level opening
tbc...
- [ ] Furthermore, split `Schema` and `TableEngine` completely? 

+ Just keep `register_table` and `unregister_table` in `Schema`
+ Remove all table operation(create,drop,open,close), and call them directly in `TableEngine`

#### 2. Impl shard based table meta recovery
- [x] Refactor manifest module
    - Rename the original `TableData` to `TableContext`, and extract members which will be recovered from manifest as the new `TableData`.
    - Register the `TableData` into manifest when `create table` and `open table`, and unregister it when `drop table` and `close table`.
    - When do snapshot in `Manifest`, we just make use of the hold `TableData` rather than scanning the persist wals.
- [x] Place `TableData`s into `Manifest`, and we update the memory and storage in just one place.
- [ ] Shard based manifest recovery

#### 3. Impl shard based table data recovery
- [x] Region based wal replay  #976 
- [ ] Make wal replay more concurrently

#### 4. other
- [x] Add integration test about recovery #996 
- [ ] [WIP] retry when open shard failed
- [ ] Support limited retry in kafka client #1005 
- [ ] Fix logs deleting in wal on kafka

[...]

PR #825, opened 2023-04-11
feat: refactor manifest to get snapshot in memory

Rachelint:

Closes #
Part of #799 

## Rationale for this change
 Make snapshot able to get snapshot from table data.

Make snapshot able to get snapshot from table data.

None.

## How does this change test
Test by ut.

### DIFF
--- a/analytic_engine/src/manifest/details.rs
+++ b/analytic_engine/src/manifest/details.rs
@@ -166,8 +166,11 @@ impl MetaUpdateLogEntryIterator for MetaUpdateReaderImpl {
 ///
 /// Mainly for getting the [TableManifestData] from memory.
 pub(crate) trait TableSnapshotProvider: fmt::Debug + Send + Sync {
-    fn get_table_snapshot(&self, space_id: SpaceId, table_id: TableId)
-        -> Result<Option<TableManifestData>>;
+    fn get_table_snapshot(
+        &self,
+        space_id: SpaceId,
+        table_id: TableId,
+    ) -> Result<Option<TableManifestData>>;
 }
 
 #[derive(Clone)]
@@ -197,7 +200,7 @@ impl TableSnapshotProvider for TableSnapshotProviderImpl {
             .context(BuildSnapshotNoCause {
                 msg: format!("table data not exist, space_id:{space_id}, table_id:{table_id}",),
             })?;
-        
+
         // When table has been dropped, we should return None.
         let table_manifest_data_opt = if !table_data.is_dropped() {
             let table_meta = AddTableMeta {
@@ -223,9 +226,7 @@ impl TableSnapshotProvider for TableSnapshotProviderImpl {
             Some(TableManifestData {
                 table_meta,
                 version_meta: Some(version_meta),
-                }
-            )
-
+            })
         } else {
             None
         };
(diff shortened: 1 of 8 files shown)


analytic_engine/src/manifest/details.rs
```diff
@@ -317,12 +524,14 @@ impl Manifest for ManifestImpl {
...
-            location,
+        let storage_snapshot_builder = StorageSnapshotBuilder {
             log_store,
             snapshot_store,
         };
-        let snapshot = snapshotter.create_latest_snapshot().await?;
+        let snapshotter = Snapshotter {
+            snapshot_builder: storage_snapshot_builder,
+        };
+        let snapshot = snapshotter.snapshot().await?;
```
ShiKaiWi, line 534: In the previous implementation, only the latest snapshot is fetched but won't be stored and the old logs would also be kept. If so, a snapshot will be generated and stored every time read the table manifest data, and I guess it is unacceptable.

ShiKaiWi (commented):
This solution sounds good. However, there may be one potential problem in it. Here is an example for it:
1. The manifest wal is updated;
2. Snapshot is triggered;
3. Table metadata in the memory is used to form the latest snapshot;

In the step 3, the table metadata (snapshot) may be different from the logs from manifest wal.

analytic_engine/src/manifest/details.rs
```diff
@@ -588,99 +791,15 @@ impl From<Snapshot> for manifest_pb::Snapshot {
...
-            Ok(Some(Snapshot {
-                end_seq: latest_seq,
-                original_logs_num: num_logs,
-                data: manifest_data_builder.build(),
-            }))
-        } else {
-            Ok(None)
-        }
+    pub async fn snapshot(&self) -> Result<Option<Snapshot>> {
+        self.snapshot_builder.build().await
```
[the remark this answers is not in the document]
Rachelint: I have splitter `Snapshotter` to `Snapshotter` and `SnapshotterRecoverer`, and removed the `SnapshotBuilder` now.

Rachelint:
> This solution sounds good. However, there may be one potential problem in it. Here is an example for it:
> 
>     1. The manifest wal is updated;
> 
>     2. Snapshot is triggered;
> 
>     3. Table metadata in the memory is used to form the latest snapshot;
> 
> 
> In the step 3, the table metadata (snapshot) may be different from the logs from manifest wal.

In fact, this situation will not occurred.
Snapshot works always done before writing the new comming logs to log store.

analytic_engine/src/manifest/details.rs
```diff
@@ -155,6 +162,192 @@ impl MetaUpdateLogEntryIterator for MetaUpdateReaderImpl {
...
+        &self,
+        space_id: SpaceId,
+        table_id: TableId,
+    ) -> Result<Option<TableManifestData>>;
+}
+
+#[derive(Clone)]
+pub(crate) struct TableSnapshotProviderImpl {
+    pub(crate) spaces: SpacesRef,
+}
```
ShiKaiWi, line 179: Maybe we should put this implementation to the space or table module?

ShiKaiWi:
> > This solution sounds good. However, there may be one potential problem in it. Here is an example for it:
> > ```
> > 1. The manifest wal is updated;
> > 
> > 2. Snapshot is triggered;
> > 
> > 3. Table metadata in the memory is used to form the latest snapshot;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In the step 3, the table metadata (snapshot) may be different from the logs from manifest wal.
> 
> In fact, this situation will not be occurred. Snapshot works always done before writing the new comming logs to log store.

After some thoughts, I'm still worried about that the table metadata in the memory is not controlled by the manifest, that is to say, the order of updating manifest before the table metadata in memory is implicit. Maybe we can extend the TableSnapshotProvider to TabeSnapshotManager which is not only able to provide the latest snapshot but also receive metadata updates.

Rachelint:
> > > This solution sounds good. However, there may be one potential problem in it. Here is an example for it:
> > > ```
> > > 1. The manifest wal is updated;
> > > 
> > > 2. Snapshot is triggered;
> > > 
> > > 3. Table metadata in the memory is used to form the latest snapshot;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > In the step 3, the table metadata (snapshot) may be different from the logs from manifest wal.
> > 
> > 
> > In fact, this situation will not be occurred. Snapshot works always done before writing the new comming logs to log store.
> 
> After some thoughts, I'm still worried about that the table metadata in the memory is not controlled by the manifest, that is to say, the order of updating manifest before the table metadata in memory is implicit. Maybe we can extend the TableSnapshotProvider to TabeSnapshotManager which is not only able to provide the latest snapshot but also receive metadata updates.

I will sovle it in next pr.

ShiKaiWi (approved):
LGTM

PR #826, opened 2023-04-11
feat: replace bg runtime with default and compact runtime

ShiKaiWi:

Closes #

## Rationale for this change
 Currently, the `bg_runtime` is used for many scenarios, e.g. grpc service, hotspot recording, and so on. And `bg_runtime` is also used heavily by compaction, that is to say, the other scenarios will be hungry to run, leading to blocking in the key paths, e.g. write procedure.

- Replace the bg_runtime with default_runtime and compact_runtime.

The runtime config will be updated.

## How does this change test
By manually tests, it is found written rows per second changes from 4K(1 bg thread) -> 100K(1 default thread + 1 compact thread).

### DIFF
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -568,7 +568,7 @@ impl Instance {
             let sst_write_options = sst_write_options.clone();
 
             // spawn build sst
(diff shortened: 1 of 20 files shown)


chunshao90 (approved):
LGTM

PR #845, opened 2023-04-17
fix: wrong default write batch size for run_tsbs

ShiKaiWi:

Closes #

## Rationale for this change
 A bug in the run_tsbs.sh

- Fix the wrong syntax of the default write batch size in `run_tsbs.sh`.

None.

## How does this change test
Manually.


chunshao90 (approved):
LGTM

PR #848, opened 2023-04-19
feat: compaction support different level

jiacai2050:

Closes #

## Rationale for this change
 
Currently all sst are in level0, and we find that when flush memtable-> sst0,
build parquet filter cost lots of CPU, which affect write throughout indirectly.

Introduce SST at level 1, this is how data is moved between different components:
- memtable -> level0
- level0 -> level1
- level1 -> level1

Flush memtable will only output sst at level 0, and it will skip parquet filter building. 

Fow now, compaction strategy is same for 0->1 and 1->1, later we need to investigate different strategy for them.

No

## How does this change test

CI

### DIFF
--- a/analytic_engine/src/row_iter/merge.rs
+++ b/analytic_engine/src/row_iter/merge.rs
@@ -34,7 +34,7 @@ use crate::{
     space::SpaceId,
     sst::{
         factory::{FactoryRef as SstFactoryRef, ObjectStorePickerRef, SstReadOptions},
-        file::FileHandle,
+        file::{FileHandle, Level},
         manager::MAX_LEVEL,
     },
     table::version::{MemTableVec, SamplingMemTable},
@@ -151,8 +151,8 @@ impl<'a> MergeBuilder<'a> {
     }
 
     /// Returns file handles in `level`, panic if level >= MAX_LEVEL
-    pub fn mut_ssts_of_level(&mut self, level: u16) -> &mut Vec<FileHandle> {
-        &mut self.ssts[usize::from(level)]
+    pub fn mut_ssts_of_level(&mut self, level: Level) -> &mut Vec<FileHandle> {
+        &mut self.ssts[level.as_usize()]
     }
 
     pub async fn build(self) -> Result<MergeIterator> {
--- a/analytic_engine/src/row_iter/merge.rs
+++ b/analytic_engine/src/row_iter/merge.rs
@@ -34,7 +34,7 @@ use crate::{
     space::SpaceId,
     sst::{
         factory::{FactoryRef as SstFactoryRef, ObjectStorePickerRef, SstReadOptions},
-        file::{FileHandle, Level},
+        file::{FileHandle, Level, SST_LEVEL_NUM},
     },
     table::version::{MemTableVec, SamplingMemTable},
 };
@@ -126,7 +126,7 @@ impl<'a> MergeBuilder<'a> {
             config,
             sampling_mem: None,
             memtables: Vec::new(),
-            ssts: vec![Vec::new(); Level::TOTAL_NUM],
+            ssts: vec![Vec::new(); SST_LEVEL_NUM],
         }
     }
 
--- a/analytic_engine/src/row_iter/merge.rs
+++ b/analytic_engine/src/row_iter/merge.rs
@@ -35,7 +35,6 @@ use crate::{
     sst::{
         factory::{FactoryRef as SstFactoryRef, ObjectStorePickerRef, SstReadOptions},
         file::{FileHandle, Level},
-        manager::MAX_LEVEL,
     },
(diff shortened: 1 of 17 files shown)


analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -183,14 +187,23 @@ impl RecordBatchGroupWriter {
         )
         .box_err()
         .context(EncodeRecordBatch)?;
+        let mut parquet_filter = if self.level.is_min() {
```
ShiKaiWi, line 190: Shall we provide a method called `should_build_parquet_filter` which decided by `!self.level.is_min() && !self.hybrid_encoding`? After that, we can remove the option wrapper of the return value of `build_row_group_filter` method, which will look pretty.

jiacai2050: Good suggestion

analytic_engine/src/sst/manager.rs
```diff
@@ -71,20 +66,20 @@ impl LevelsController {
...
     pub fn remove_ssts_from_level(&mut self, level: Level, file_ids: &[FileId]) {
-        let level_handler = &mut self.levels[usize::from(level)];
+        let level_handler = &mut self.levels[level.as_usize()];
         level_handler.remove_ssts(file_ids);
     }
 
     /// Total number of levels.
-    pub fn num_levels(&self) -> Level {
-        self.levels.len() as Level
+    pub fn levels(&self) -> Vec<Level> {
```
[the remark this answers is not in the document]
jiacai2050: OK

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -139,12 +142,14 @@ impl RecordBatchGroupWriter {
...
 
-    /// Build the parquet filter for the given `row_group`, and then update it
-    /// into `self.parquet_filter`.
-    fn update_parquet_filter(&mut self, row_group_batch: &[RecordBatchWithKey]) -> Result<()> {
+    /// Build the parquet filter for the given `row_group`
+    fn build_row_group_filter(
+        &self,
+        row_group_batch: &[RecordBatchWithKey],
+    ) -> Result<Option<RowGroupFilter>> {
         // TODO: support filter in hybrid storage format [#435](https://github.com/CeresDB/ceresdb/issues/435)
```
Rachelint: This issue seems to be outdated?

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -139,12 +142,14 @@ impl RecordBatchGroupWriter {
         Ok(curr_row_group)
     }
 
-    /// Build the parquet filter for the given `row_group`, and then update it
-    /// into `self.parquet_filter`.
-    fn update_parquet_filter(&mut self, row_group_batch: &[RecordBatchWithKey]) -> Result<()> {
+    /// Build the parquet filter for the given `row_group`
```
Rachelint, line 145: /// Build the parquet filter for the given `row_group`.

analytic_engine/src/row_iter/merge.rs
```diff
@@ -127,7 +126,7 @@ impl<'a> MergeBuilder<'a> {
             config,
             sampling_mem: None,
             memtables: Vec::new(),
-            ssts: vec![Vec::new(); MAX_LEVEL],
+            ssts: vec![Vec::new(); Level::TOTAL_NUM],
```
Rachelint, line 129: It seems a bit strange, return the `TOTAL_NUM` from the single `Level`?
Maybe it should belong to something like levels?

analytic_engine/src/sst/parquet/writer.rs
jiacai2050: Which PR fix it?

analytic_engine/src/sst/parquet/writer.rs
jiacai2050: Isn't same with before?

analytic_engine/src/row_iter/merge.rs
jiacai2050: If you are uncomfortable, I can move it out of Level, and define a `SST_TOTAL_LEVEL` in file.rs directly.

Rachelint: It sounds better.

analytic_engine/src/sst/parquet/writer.rs
Rachelint: `.` is added.

analytic_engine/src/sst/parquet/writer.rs
Rachelint: I see it closed.
Reopened now.

ShiKaiWi (approved):
LGTM

PR #849, opened 2023-04-19
chore: rename module `sql` to `query_frontend`

Rachelint:

Closes #

## Rationale for this change
Sql module not only include parser and planner about sql, but more like promql and influxql, so I think naming it `query_frontend` may be more suitable. 

Rename the `sql` module.

None.

## How does this change test
Not need.


ShiKaiWi (approved):
LGTM

PR #852, opened 2023-04-21
feat: support domain name as the ceresdb node addr

ShiKaiWi:

Closes #

## Rationale for this change
 Currently, the node address reported to CeresMeta is not allowed to be a domain name, but it is necessary for k8s deployment.

- Allow the node address reported to CeresMeta to be anything.

None.

## How does this change test
Existing tests.


Rachelint (approved):
LGTM

Issue #830, opened 2023-04-12
Tracking issue for influxql support


jiacai2050:
### Describe This Problem

After #824, influxql support is almost finished, users can query ceresdb via influxdb datasource in grafana. 

This issue track some improvements:

- [x] In InfluxQLSchemaProvider we hold all tables in HashMap, this is expensive and in theory we could load tables in lazy way.
- [x] Currently depend on iox directly, maybe we could extract them to https://github.com/CeresDB/influxql, or better to contribute to iox project.

### Proposal

See above.

Rachelint:
All task finished, close it until new tasks came.

PR #854, opened 2023-04-22
feat: only collecting all tables on demand in influxql planner

Rachelint:

Closes #
Part of #830 

## Rationale for this change
Now, we collect and keep all tables in influxql planner even though it is unused.
It is wasteful to do this. So in this pr, I make it just do such works on demand.
 

+ Add `all_tables` method in `MetaProvider`.
+ Just call `MetaProvider::all_tables` to get tables while they are actually needed(such as while influxql's `SchemaProvider::table_names` is called).

None.

## How does this change test
Test by ut.

### DIFF
--- a/query_frontend/src/influxql/planner.rs
+++ b/query_frontend/src/influxql/planner.rs
@@ -37,9 +37,7 @@ pub const CERESDB_MEASUREMENT_COLUMN_NAME: &str = "iox::measurement";
 // Port from https://github.com/ceresdb/influxql/blob/36fc4d873e/iox_query_influxql/src/frontend/planner.rs#L28
 struct InfluxQLSchemaProvider<'a, P: MetaProvider> {
     context_provider: ContextProviderAdapter<'a, P>,
-    // TODO: avoid load all tables.
-    // if we can ensure `table_names` is only called once, then load tables lazily is better.
-    table_cache: OnceCell<Vec<TableRef>>,
+    tables_cache: OnceCell<Vec<TableRef>>,
 }
 
 impl<'a, P: MetaProvider> SchemaProvider for InfluxQLSchemaProvider<'a, P> {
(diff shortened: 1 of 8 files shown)


query_frontend/src/influxql/planner.rs
```diff
@@ -58,21 +66,53 @@ impl<'a, P: MetaProvider> SchemaProvider for InfluxQLSchemaProvider<'a, P> {
     }
 
     fn table_names(&self) -> Vec<&'_ str> {
-        self.tables.keys().map(|k| k.as_str()).collect::<Vec<_>>()
+        let tables = match self
```
jiacai2050, line 69: You can also to implement `table_exists` method, then it won't call `table_names`.

There is a `table` method in `MetaProvider` for this purpose.
```
fn table(&self, name: TableReference) -> Result<Option<TableRef>>;
```

query_frontend/src/provider.rs
```diff
@@ -72,6 +84,9 @@ pub trait MetaProvider {
 
     /// Get udaf by name.
     fn aggregate_udf(&self, name: &str) -> Result<Option<AggregateUdf>>;
+
+    /// Return all tables.
+    fn all_tables(&self) -> Result<Vec<TableRef>>;
```
jiacai2050, line 89: Add comments when this is used now, and warn users this maybe incur expensive cost.

query_frontend/src/provider.rs
```diff
@@ -138,6 +153,34 @@ impl<'a> MetaProvider for CatalogMetaProvider<'a> {
...
+    // refactor the tables collecting procedure.
+    fn all_tables(&self) -> Result<Vec<TableRef>> {
+        let catalog = match self
+            .manager
+            .catalog_by_name(self.default_catalog)
+            .with_context(|| FindCatalog {
+                name: self.default_catalog,
+            })? {
+            Some(catalog) => catalog,
+            None => return Ok(Vec::default()),
```
jiacai2050, line 167: Why not throw error when no catalog isn't found?

query_frontend/src/influxql/planner.rs
Rachelint: Thanks for reminding, I forgot this default impl method...

query_frontend/src/provider.rs
Rachelint: I think it again, better to return error really.

PR #861, opened 2023-04-26
fix: build filter when needed

jiacai2050:

Closes #

## Rationale for this change
 
An obvious bug when init filter

No


ShiKaiWi (approved):
LGTM

PR #862, opened 2023-04-26
feat: rename the compaction_config to compaction and adjust interval

ShiKaiWi:

Closes #

## Rationale for this change
After #822, the compaction triggered by flush will take place less frequently. And #861, the filter won't be built for level0 sst, so the query performance may drop a little if we don't adjust the compaction frequency.

- Rename the config name `compaction_config` to `compaction`, which is more elegant;
- Adjust the interval of periodic compaction from 30m to 5m;

The config name `compaction_config` is replaced with `compaction`.

## How does this change test
Manually.


jiacai2050 (approved):
LGTM

PR #863, opened 2023-04-26
feat: place table datas into manifest, update them together

Rachelint:

Closes #
Part of #799 

## Rationale for this change
 Now, we update `TableData` and store its wal seperately. The order of two operations above is maintained by developer, that will lead to a big bug if developer is not so careful when modifying related codes.

+ Place table datas into manifest.
+ Update it both in memory and storage in the same call.

None.

## How does this change test
Test by ut.


analytic_engine/src/instance/open.rs
```diff
@@ -61,8 +59,16 @@ impl Instance {
...
     ) -> Result<Arc<Self>> {
         let spaces: Arc<RwLock<Spaces>> = Arc::new(RwLock::new(Spaces::default()));
-        let table_snapshot_provider = Arc::new(TableSnapshotProviderImpl {
+        let default_runtime = ctx.runtimes.default_runtime.clone();
+        let file_purger = Arc::new(FilePurger::start(
+            &default_runtime,
+            store_picker.default_store().clone(),
+        ));
+
+        let table_snapshot_provider = Arc::new(TableMetaSetImpl {
```
ShiKaiWi, line 68: Shall we rename this variable name?

analytic_engine/src/table_meta_set_impl.rs
```diff
@@ -0,0 +1,326 @@
...
+                    table_data.set_table_options(options);
+
+                    Ok(())
+                });
+                find_space_and_apply_edit(space_id, alter_option)
+            }
+        }
+    }
+
+    fn apply_snapshot(
```
ShiKaiWi, line 197: Maybe `install_snapshot` is a better name?

analytic_engine/src/table_meta_set_impl.rs
```diff
@@ -0,0 +1,326 @@
...
+            dyn FnOnce(Arc<Space>) -> crate::manifest::details::Result<()>,
+        >| {
+            let spaces = self.spaces.read().unwrap();
+            let space = spaces
+                .get_by_id(space_id)
+                .with_context(|| ApplyUpdateToTableNoCause {
+                    msg: format!("space not found, space_id:{space_id}"),
+                })?;
+            apply_edit(space.clone())
+        };
```
ShiKaiWi, line 67: Why not define this as a method on the `TableMetaSetImpl` with apply_edit as a generic parameter?

PR #865, opened 2023-04-28
chore: proxy as a separate module

chunshao90:

Closes #

## Rationale for this change
The proxy module code is getting more and more complex, and it is logically inconsistent with the server module, so proxy is defined as a separate module.

* Proxy as a separate module.

No.

## How does this change test
CI.


Rachelint (approved):
LGTM

PR #867, opened 2023-05-04
refactor: rewrite UDF time_bucket using date_bin and date_trunc

### LINKED ISSUE #558: Deprecate custom UDF time_bucket in favor of date_bin
### Describe This Problem

In early version of CeresDB, there is a custom UDF `time_bucket` to align time of values, but now datafusion support it out of box, so I suggest deprecate our `time_bucket`, and use `data_bin` instead. Ex:

```sql
SELECT
      DATE_BIN(INTERVAL '15' minute, time, TIMESTAMP '2001-01-01T00:00:00Z') AS time,
      val
    FROM (
      VALUES
        (TIMESTAMP '2021-06-10 17:05:00Z', 0.5),
        (TIMESTAMP '2021-06-10 17:19:10Z', 0.3)
      ) as t (time, val)


        "+---------------------+-----+",
        "| time                | val |",
        "+---------------------+-----+",
        "| 2021-06-10 17:00:00 | 0.5 |",
        "| 2021-06-10 17:15:00 | 0.3 |",
        "+---------------------+-----+",
```

### Proposal

1. Completely remove it before 1.0 is out.
2. If cannot remove it, add a deprecation warning.

### Additional Context

https://github.com/apache/arrow-datafusion/issues/3015

dust1:

Closes #558 

Some time_bucket functions use an implementation of date_bin

none

## How does this change test
none


dust1:
@jiacai2050 hello. I have a problem, after I debug the `truncate_week` implementation of `time_bucket`. the ts `1659577423000 (2022-8-4 9:43:43)` will truncate to `1652918400000 (2022-5-19 8:0:0)`. So I modified some of the test cases

jiacai2050:
@dust1 Don't worry, I will review this PR this week soon.

integration_tests/cases/common/function/time_bucket.result
```diff
@@ -49,12 +49,12 @@ Timestamp(1659283200000),
 SELECT time_bucket(`timestamp`, 'P1W') FROM `02_function_time_bucket_table`;
 
 time_bucket(02_function_time_bucket_table.timestamp,Utf8("P1W")),
-Timestamp(1652918400000),
-Timestamp(1657756800000),
-Timestamp(1657756800000),
-Timestamp(1657756800000),
-Timestamp(1657756800000),
-Timestamp(1657756800000),
+Timestamp(1656259200000),
```
jiacai2050, line 52: I still have question about this week result.
The original ts is `1659577423000`, 2022-07-04 09:43:43+08

When truncate by week, it should be `2022-07-04 00:00:00+08`.

Tested aginst PostgreSQL
```
SELECT
    date_trunc('week', '2022-07-04T09:43:43+08'::timestamp with time zone),
    extract(epoch FROM date_trunc('week', '2022-07-04T09:43:43+08'::timestamp with time zone));
```
Output
```
       date_trunc       |      extract      
------------------------+-------------------
 2022-07-04 00:00:00+08 | 1656864000.000000
(1 row)
```

dust1: It should be a time zone problem.  I am currently dealing with it after `trunc`. this part should be placed before the function call

jiacai2050: 👍 

Also could you update those sql to something like
```
SELECT timestamp, time_bucket(`timestamp`, 'P1D') FROM `02_function_time_bucket_table` order by timestamp;
```

df_operator/src/udfs/time_bucket.rs
```diff
@@ -235,89 +232,87 @@ impl Period {
...
-            .unwrap();
-        let truncated_ts = truncated_datetime.timestamp_millis();
+    fn fix_offset(array: &DfColumnarValue) -> Result<DfColumnarValue> {
+        let list = match array {
+            DfColumnarValue::Array(array) => {
+                let array = as_timestamp_nanosecond_array(&array).context(TruncateTimestamp)?;
+                array
+                    .iter()
+                    .map(|ts| {
+                        ts.map(|t| Ok(t + DEFAULT_TIMEZONE_OFFSET_SECS as i64 * 1_000_000_000))
```
jiacai2050, line 308: I don't understand here.

Usually first you need to convert timestamp to local datetime, then truncate by week/month/year.

You can refer how datafusion implement similar functions:
https://github.com/apache/arrow-datafusion/blob/8ada7fd1e949b1ff2c9207ce2143bb19157e75cd/datafusion/physical-expr/src/datetime_expressions.rs#L218

jiacai2050:
Also this rewrite doesn't simplify time_bucket much, I expect it's should be much simpler than before.

Part of this PR was to convert timestamp precision, which was unnecessary in datafusion 24
- https://github.com/apache/arrow-datafusion/blob/main/dev/changelog/24.0.0.md

I plan to upgrade datafusion first, then implement this afterwards.

df_operator/src/udfs/time_bucket.rs
dust1: `date_trunc` is converted to UTC after receiving the timestamp, which subtracts the time zone, This will result in the wrong time zone. what do I think is wrong?😂

jiacai2050: If `date_trunc` convert UTC datetime, then we can't using it directly...

I will check this carefully later.

dust1:
Can we implement a `date_trunc` inside `time_bucket` first🤔

jiacai2050:
> Can we implement a date_trunc inside time_bucket first🤔

I think it make little sense to do this, the logic won't be simpler than before.

The original intention is to simplify time_bucket's implementation, so I think you can wait a little moment on this PR, after I bump datafusion to 24, much part of this PR could be removed.

jiacai2050:
After bump datafusion in #894, there are following issues:
- `date_trunc` will use UTC datetime to do truncate, which is different from our default values. https://github.com/apache/arrow-datafusion/issues/5976
- `date_trunc` still output nano timestamp
https://github.com/apache/arrow-datafusion/blob/063f99fd61aa8f2854dc138d16864ae485a7116e/datafusion/physical-expr/src/datetime_expressions.rs#L287

This PR cannot proceed until those two issues be fixed...

tanruixiang:
Hey, I noticed that the previous issue has been fixed by [datafusion](https://github.com/apache/arrow-datafusion/issues/5976), but this pr is already too far behind the master branch. Would you be willing to merge the master branch in this pr? Or I can open another pr and invite you to review and co-author. Whichever way you choose I will work with you to resolve this issue.

jiacai2050:
Kinds of stale, will open new PR to address this.

PR #868, opened 2023-05-05
chore!: rename table name in table_kv based wal

Rachelint:

Closes #

## Rationale for this change
The wal related table name in table_kv based wal is strange.

Rename related tables names.

  BREAKING CHANGE, users using obkv based wal will lose the wal data.

## How does this change test
Test by ut.


chunshao90 (approved):
LGTM

PR #869, opened 2023-05-05
fix: fix write partition table

chunshao90:

Closes #

## Rationale for this change
Fix the problem that grpc will repeatedly create a table when writing to the partition table.

* Open partition table when build write plan.

No.

## How does this change test
Manual test and CI.


ShiKaiWi (approved):
LGTM

PR #871, opened 2023-05-06
fix: drop partition table

chunshao90:

Closes #

## Rationale for this change
 Fix the problem that the partition table in the proxy memory was not dropped after the partition table was dropped.

 Fix the problem that the partition table in the proxy memory was not dropped after the partition table was dropped.

No.

## How does this change test
Manual test and CI.

### DIFF
--- a/proxy/src/lib.rs
+++ b/proxy/src/lib.rs
@@ -257,24 +257,34 @@ impl<Q: QueryExecutor + 'static> Proxy<Q> {
                 msg: format!("Failed to find table, table_name:{table_name}"),
             })?;
 
-        let partition_table_info = self
+        let partition_table_info_in_meta = self
             .router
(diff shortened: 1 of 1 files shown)


proxy/src/lib.rs
```diff
@@ -256,44 +248,65 @@ impl<Q: QueryExecutor + 'static> Proxy<Q> {
...
+                            schema.clone(),
+                            catalog_name.to_string(),
+                            schema_name.to_string(),
+                            table_name.to_string(),
+                        )
+                        .await?;
+                    }
+                    // No need to create non-partition table.
+                    return Ok(());
+                }
```
ShiKaiWi, line 300: Define this as a method called `maybe_drop_partition_table`?

ShiKaiWi (approved):
LGTM

PR #872, opened 2023-05-06
fix: fix evict logic in remote client

Rachelint:

Closes #

## Rationale for this change
 Some situations are not considered in remote client's evict logic.

See title.

None.

## How does this change test
Test by integration test.

### DIFF
--- a/remote_engine_client/src/client.rs
+++ b/remote_engine_client/src/client.rs
@@ -87,6 +87,8 @@ impl Client {
             }
         };
 
+        // When success to get the stream, table has been found in remote, not need to
+        // evict cache entry.
         let response = response.into_inner();
         let remote_read_record_batch_stream =
             ClientReadRecordBatchStream::new(table_ident, response, projected_schema);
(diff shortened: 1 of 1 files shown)


remote_engine_client/src/client.rs
```diff
@@ -116,27 +118,26 @@ impl Client {
...
-            Ok(response) => response,
-
-            Err(e) => {
-                // If occurred error, we simply evict the corresponding channel now.
-                // TODO: evict according to the type of error.
-                self.evict_route_from_cache(&[table_ident]).await;
-                return Err(e);
+        let result = result.and_then(|response| {
+            let response = response.into_inner();
+            if let Some(header) = &response.header && !status_code::is_ok(header.code) {
```
ShiKaiWi, line 123: Is it possible that the header is missing?

remote_engine_client/src/client.rs
```diff
@@ -203,40 +204,31 @@ impl Client {
...
+            let result = batch_result.and_then(|response| {
+                let response = response.into_inner();
+                if let Some(header) = &response.header && !status_code::is_ok(header.code) {
+                    Server {
+                        table_idents: table_idents.clone(),
+                        code: header.code,
+                        msg: header.error.clone(),
+                    }.fail().box_err()
+                } else {
+                    Ok(response.affected_rows)
```
ShiKaiWi, line 216: Shall we define a function to do this check to avoid the duplicate codes?

ShiKaiWi (approved):
LGTM

PR #875, opened 2023-05-08
feat: impl influxdb api with proxy

chunshao90:

Closes #

## Rationale for this change
 Implement influxdb api with proxy.

* Add `influxdb` in proxy module.
* Remove some useless code in http query.

No.

## How does this change test
CI.


proxy/src/http/query.rs
```diff
@@ -189,14 +140,14 @@ impl<Q: QueryExecutor + 'static> Proxy<Q> {
...
-            query_request
+            req
         );
 
         Ok(output)
     }
 }
 #[derive(Debug, Deserialize)]
-pub struct Request {
+pub struct SqlRequest {
```
jiacai2050, line 150: If we rename this file to sql.rs, then we don't need to rename this struct.

proxy/src/influxdb/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+    influxdb::types::{convert_write_request, InfluxqlRequest, WriteRequest, WriteResponse},
+    Proxy,
+};
+
+impl<Q: QueryExecutor + 'static> Proxy<Q> {
+    pub async fn handle_influxdb_query(
+        &self,
+        ctx: RequestContext,
+        req: InfluxqlRequest,
+    ) -> Result<Output> {
```
jiacai2050, line 37: How about return `Result<InfluxqlResponse>` directly?

jiacai2050 (approved):
LGTM

PR #876, opened 2023-05-09
feat: add metrics for http requests

jiacai2050:

Closes #

## Rationale for this change
 
Currently there are no metrics for http handler.

Add http handle metrics

No

## How does this change test

Manually

### DIFF
--- a/server/src/http.rs
+++ b/server/src/http.rs
@@ -43,7 +43,10 @@ use warp::{
     Filter,
 };
 
-use crate::{consts, error_util, metrics};
+use crate::{
+    consts, error_util,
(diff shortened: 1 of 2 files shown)


server/src/http.rs
```diff
@@ -193,6 +196,18 @@ impl<Q: QueryExecutor + 'static> Service<Q> {
             .or(self.cpu_profile())
             .or(self.server_config())
             .or(self.stats())
+            .with(warp::log("http_requests"))
```
jiacai2050, line 199: This will log something like
```
127.0.0.1:43550 "POST /prom/v1/write HTTP/1.1" 200 "-" "Prometheus/2.37.8" 23.71479ms
```

chunshao90 (approved):
LGTM

PR #877, opened 2023-05-09
chore: remove udeps in CI

chunshao90:

Closes #

## Rationale for this change
 The running time of `udeps` is too long, so it is suitable to put it in CI.

Remove udeps in CI.

No.

## How does this change test
No need.


jiacai2050 (approved):
LGTM

PR #878, opened 2023-05-09
feat: impl forwarding prom write

chunshao90:

Closes #

## Rationale for this change
 Implement forwarding prom write.

 Implement forwarding prom write.

No.

## How does this change test
Manual test and CI.

### DIFF
--- a/proxy/src/http/prom.rs
+++ b/proxy/src/http/prom.rs
@@ -63,6 +63,8 @@ impl<Q: QueryExecutor + 'static> Proxy<Q> {
             runtime: self.engine_runtimes.write_runtime.clone(),
             timeout: ctx.timeout,
         };
+        // TODO: Define a new public method instead of calling the grpc write method
+        // directly.
         let result = self.handle_write(ctx, table_request).await;
         if let Some(header) = result.header {
             if header.code != StatusCode::OK.as_u16() as u32 {
--- a/proxy/src/http/prom.rs
+++ b/proxy/src/http/prom.rs
@@ -13,8 +13,9 @@ use std::{
 use async_trait::async_trait;
 use catalog::consts::DEFAULT_CATALOG;
 use ceresdbproto::storage::{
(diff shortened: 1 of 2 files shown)


proxy/src/http/prom.rs
```diff
@@ -43,59 +44,36 @@ use crate::{
...
-    grpc::write::{execute_insert_plan, WriteContext},
-    Proxy,
+    Context as ProxyContext, Proxy,
 };
 
 impl reject::Reject for Error {}
 
 impl<Q: QueryExecutor + 'static> Proxy<Q> {
     /// Handle write samples to remote storage with remote storage protocol.
     async fn handle_prom_write(&self, ctx: RequestContext, req: WriteRequest) -> Result<()> {
```
jiacai2050, line 54: InfluxDB write should also updated.

proxy/src/http/prom.rs
```diff
@@ -43,59 +44,36 @@ use crate::{
...
+            context: Some(GrpcRequestContext {
+                database: ctx.schema.clone(),
+            }),
+            table_requests: write_table_requests,
+        };
+        let ctx = ProxyContext {
+            runtime: self.engine_runtimes.write_runtime.clone(),
+            timeout: ctx.timeout,
+        };
+        let result = self.handle_write(ctx, table_request).await;
```
jiacai2050, line 66: It's wired to do this special handling, isn't is more reasonable for`handle_write` to `Result<()>`

proxy/src/http/prom.rs
chunshao90: I will refactor all the writing process in another pr, and will not modify it in this pr for now.

proxy/src/http/prom.rs
chunshao90: I will refactor all the writing process in another pr, and will not modify it in this pr for now.

jiacai2050 (approved):
LGTM

PR #879, opened 2023-05-09
feat: support merge small write requests

ShiKaiWi:

Closes #

## Rationale for this change
 Currently, massive small write requests will lead to bad performance because of the lock contention for table writes. By buffering the write requests blocked by the write lock into a wait queue, these small write requests may be merged, leading to a much better performance.

- Buffer the write requests blocked by the write lock and merge and write them when another write worker acquires the write lock.

None.

## How does this change test
Existing tests and the newly added unit tests.

### DIFF
--- a/analytic_engine/src/table/mod.rs
+++ b/analytic_engine/src/table/mod.rs
@@ -13,7 +13,7 @@ use common_types::{
 use common_util::error::BoxError;
 use datafusion::{common::Column, logical_expr::Expr};
 use futures::TryStreamExt;
-use log::warn;
+use log::{error, warn};
 use snafu::{ensure, OptionExt, ResultExt};
 use table_engine::{
     partition::PartitionInfo,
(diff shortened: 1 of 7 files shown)


analytic_engine/src/table/mod.rs
```diff
@@ -112,19 +359,21 @@ impl Table for TableImpl {
     }
 
     async fn write(&self, request: WriteRequest) -> Result<usize> {
+        if self.instance.max_rows_in_write_queue > 0 {
```
jiacai2050, line 362: Add function `enable_pending_write()` to encapsulate this.

analytic_engine/src/table/mod.rs
```diff
@@ -112,19 +359,21 @@ impl Table for TableImpl {
...
+            return self.write_with_pending_queue(request).await;
+        }
+
         let mut serial_exec = self.table_data.serial_exec.lock().await;
```
jiacai2050, line 366: `try_lock` here? if failed, then `write_with_pending_queue`.

analytic_engine/src/table/mod.rs
```diff
@@ -81,6 +94,240 @@ impl fmt::Debug for TableImpl {
...
+        self.pending_writes.try_push(request)
+    }
+
+    #[inline]
+    fn is_full(&self) -> bool {
+        self.pending_writes.num_rows >= self.max_rows
+    }
+
+    /// Clear the pending writes and reset the number of rows.
+    fn reset(&mut self) -> PendingWrites {
```
jiacai2050, line 196: take?

jiacai2050:
Any benchmark result about this PR?

analytic_engine/src/table/mod.rs
ShiKaiWi: Add a method `should_queue_write_request` to do this check.

analytic_engine/src/table/mod.rs
ShiKaiWi: There is no need to submit the write request to the queue if the queue's capacity is smaller than the write requests.

Rachelint (approved):
LGTM

PR #884, opened 2023-05-09
fix: remove unused write_worker module

ShiKaiWi:

Closes #

## Rationale for this change
`write_worker` module is not used any more.
Remove the `write_worker` module.

None.
## How does this change test
No need.


chunshao90 (approved):
LGTM

PR #888, opened 2023-05-11
fix: new multiple write requests when tag names are different

jiacai2050:

Closes #

## Rationale for this change
For one metric, when tag name is different, it cannot be in same write request

Use `(metric, tag_names)` as group key when new write requests
No

## How does this change test

CI and manually


chunshao90 (approved):
LGTM

PR #891, opened 2023-05-11
chore: reflush memory tables after flush failed

MachaelLee:

Closes #

## Rationale for this change
Currently, when a flush table operation fails, write requests would aways fail. Perhaps we should consider flushing the memory table again when the next write request coming to address this issue.

* flush memory table again after flush failed

None

### DIFF
--- a/analytic_engine/src/instance/serial_executor.rs
+++ b/analytic_engine/src/instance/serial_executor.rs
@@ -7,7 +7,7 @@ use std::{
 
 use common_util::{runtime::Runtime, time::InstantExt};
 use futures::Future;
-use log::error;
+use log::{error, warn};
 use table_engine::table::TableId;
 use tokio::sync::{
     oneshot,
@@ -131,7 +131,10 @@ impl TableFlushScheduler {
                     }
                     FlushState::Flushing => (),
                     FlushState::Failed { err_msg } => {
-                        return BackgroundFlushFailed { msg: err_msg }.fail();
+                        // Mark the worker is flushing.
+                        *flush_state = FlushState::Flushing;
+                        warn!("Retry to flush memory tables after background flush failed:{err_msg}");
+                        break;
                     }
                 }
 
--- a/analytic_engine/src/instance/serial_executor.rs
+++ b/analytic_engine/src/instance/serial_executor.rs
@@ -44,17 +44,19 @@ struct ScheduleSync {
 impl ScheduleSync {
     #[inline]
     pub fn should_retry_flush(&self, max_retry_limit: usize) -> bool {
-        self.continuous_flush_failure_count.load(Ordering::Relaxed) < max_retry_limit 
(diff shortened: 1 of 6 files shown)


analytic_engine/src/instance/serial_executor.rs
```diff
@@ -131,7 +131,10 @@ impl TableFlushScheduler {
                     }
                     FlushState::Flushing => (),
                     FlushState::Failed { err_msg } => {
-                        return BackgroundFlushFailed { msg: err_msg }.fail();
+                        warn!("Re-flush memory tables after background flush failed:{err_msg}");
+                        // Mark the worker is flushing.
+                        *flush_state = FlushState::Flushing;
```
Rachelint, line 136: Maybe add an option?
For some storage, failure maybe unrecoverable and retry may be meaningless?

ShiKaiWi: I guess a better way is to support limited retry, that is to say, we can use a counter to record the times of continuous flush failures (the counter will be reset to 0 once a flush succeeds). If the counter reaches a limit, we can set the `flush_state` as `FlushState::Failed`.

analytic_engine/src/instance/serial_executor.rs
```diff
@@ -34,6 +38,25 @@ type ScheduleSyncRef = Arc<ScheduleSync>;
...
+        }
+        false
+    }
+
+    #[inline]
+    pub fn reset_retry_flush_count(&self) {
+        if self.retry_flush_count.load(Ordering::Relaxed) > 0 {
+            self.retry_flush_count.store(0, Ordering::Relaxed);
+        }
+    }
```
ShiKaiWi, line 59:
```suggestion
    pub fn reset_retry_flush_count(&self) {
        self.retry_flush_count.store(0, Ordering::Relaxed);
    }
```

ShiKaiWi, line 59:
```suggestion
    pub fn reset_retry_flush_count(&self) {
        self.retry_flush_count.store(0, Ordering::Relaxed);
    }
```

analytic_engine/src/instance/serial_executor.rs
```diff
@@ -34,6 +38,25 @@ type ScheduleSyncRef = Arc<ScheduleSync>;
 struct ScheduleSync {
     state: Mutex<FlushState>,
     notifier: Sender<()>,
+    retry_flush_count: AtomicUsize,
```
ShiKaiWi, line 41: I guess a better variable should be called `contionuous_flush_failure_count`. If so, there is no need to increment it in `should_retry_flush`. And add a method `inc_flush_failure_count` and call it when flush fails, which is symmetrical to the case where flush succeeds.

MachaelLee: done

ShiKaiWi (approved):
LGTM

PR #897, opened 2023-05-16
feat: add shard related methods to table engine

Rachelint:

Closes #
Part of #799 

## Rationale for this change
+ Add `open_shard` and `close_shard` methods into `TableEngine`.
+ Impl the methods above on demand.

See above.
None.
## How does this change test
Test by ut.


ShiKaiWi (approved):
LGTM

PR #901, opened 2023-05-18
fix: reserve column case when build plan

jiacai2050:

Closes #

## Rationale for this change
When build plan for prom query, it will normalize column, but in ceresdb column name is case-sensitive.

- Replace `col` with `ident`, which will not do normalization.
No

## How does this change test
CI


integration_tests/Makefile
```diff
@@ -37,21 +37,21 @@ build-ceresdb:
 build-test:
 	cargo build
 
-build: build-ceresdb build-test build-meta
+build: build-ceresdb build-test
```
jiacai2050, line 40: Only build meta when run cluster tests

Rachelint (approved):
LGTM

PR #910, opened 2023-05-22
fix: limit input sst size when compact for old bucket

jiacai2050:
## Related Issues
Closes #

## Detailed Changes

- Adjust logs to make it more readable.
- `trim_to_threshold` respect `max_input_sstable_size`.

## Test Plan 
Manually

### DIFF
--- a/analytic_engine/src/compaction/picker.rs
+++ b/analytic_engine/src/compaction/picker.rs
@@ -484,6 +484,7 @@ impl TimeWindowPicker {
 
         let all_keys: BTreeSet<_> = buckets.keys().collect();
 
+        // First compact latest buckets
         for key in all_keys.into_iter().rev() {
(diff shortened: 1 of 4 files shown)


analytic_engine/src/compaction/picker.rs
```diff
@@ -589,8 +590,8 @@ impl LevelPicker for TimeWindowPicker {
             &ctx.segment_duration,
             opts.timestamp_resolution,
         );
-        debug!("now {}, max_ts: {}", now, ts);
-        assert!(now >= ts);
+        debug!("now {}, max_ts: {}", now, max_bucket_ts);
```
Rachelint, line 593: Too brief the log here is, how about adding some context?

Rachelint (approved):
LGTM

Issue #914, opened 2023-05-23
Use partitioned lock to optimize disk cache


jiacai2050:
### Describe This Problem

Partitioned lock is a common trick to reduce lock contention, disk cache could use it to improve perf.

https://github.com/CeresDB/ceresdb/blob/da6899c4c97089d4fbd8aa8a01db98ab8366d5bc/components/object_store/src/disk_cache.rs#L114

### Proposal

Use PartitionedRwLock to replace `cache` in `DiskCache`

https://github.com/CeresDB/ceresdb/blob/da6899c4c97089d4fbd8aa8a01db98ab8366d5bc/common_util/src/partitioned_lock.rs#L17

tanruixiang:
please assign to me.

PR #919, opened 2023-05-23
refactor: use binary bits to determine the number of partitions

tanruixiang:
## Related Issues
related with #914 

## Detailed Changes
- Modify the type of `partitions` in `PartitionedMutex` and `PartitionedRwLock`.
- Fix the bug that multiple partitions use the same lock.
## Test Plan 
Unit tests under the same file.

### DIFF
--- a/common_util/src/partitioned_lock.rs
+++ b/common_util/src/partitioned_lock.rs
@@ -15,11 +15,8 @@ pub struct PartitionedRwLock<T> {
 }
 
 impl<T> PartitionedRwLock<T> {
-    // TODO: we should get the nearest 2^n of `partition_num` as real
-    // `partition_num`. By doing so, we can use "&" to get partition rather than
-    // "%".
-    pub fn new(t: T, partition_num: NonZeroUsize) -> Self {
-        let partition_num = partition_num.get();
+    pub fn new(t: T, partition_bit: NonZeroUsize) -> Self {
+        let partition_num = 1 << partition_bit.get();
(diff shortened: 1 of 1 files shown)


common_util/src/partitioned_lock.rs
```diff
@@ -15,11 +15,8 @@ pub struct PartitionedRwLock<T> {
 }
 
 impl<T> PartitionedRwLock<T> {
-    // TODO: we should get the nearest 2^n of `partition_num` as real
-    // `partition_num`. By doing so, we can use "&" to get partition rather than
-    // "%".
-    pub fn new(t: T, partition_num: NonZeroUsize) -> Self {
-        let partition_num = partition_num.get();
+    pub fn new(t: T, partition_bit: NonZeroUsize) -> Self {
```
jiacai2050, line 18: `usize` is enough

common_util/src/partitioned_lock.rs
```diff
@@ -41,9 +38,9 @@ impl<T> PartitionedRwLock<T> {
     fn get_partition<K: Eq + Hash>(&self, key: &K) -> &RwLock<T> {
         let mut hasher = DefaultHasher::new();
         key.hash(&mut hasher);
-        let partition_num = self.partitions.len();
+        let partition_mask = self.partitions.len() - 1;
```
jiacai2050, line 41: Save `mask` inside `PartitionedRwLock`, so we can use it directly here.

jiacai2050 (approved):
LGTM

PR #928, opened 2023-05-24
chore: replace DefaultHash with AHasher

### LINKED ISSUE #921: Implement a Hasher based on murmur3
### Describe This Problem

We now use `DefaultHasher` inside partition lock, maybe we can change to a different Hasher based on murmur3 to improve perf.

### Proposal

https://github.com/CeresDB/ceresdb/blob/27267b55faf1bce3c263b49d2fb61bcd5ce55c00/common_types/src/hash.rs#L7

### Additional Context

It's better to do a simple benchmark about those Hasher.

tanruixiang:
## Related Issues
Closes #921 

## Detailed Changes
Replace the hash algorithm from `DefaultHasher` to `ahash` for better performance in partition lock.

## Test Plan
Time cost to compute hash on a 10-char string:
- DefaultHash: 10: 33.6364 ns
- ahash: 10: 19.0412 ns
- murmur3: 10: 33.0394 ns

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1380,6 +1380,7 @@ dependencies = [
 name = "common_types"
 version = "1.2.0"
 dependencies = [
+ "ahash 0.8.3",
  "arrow 38.0.0",
  "arrow_ext",
  "byteorder",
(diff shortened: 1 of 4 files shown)


jiacai2050 (approved):
LGTM

PR #931, opened 2023-05-25
chore: add hint about `ahash` usage

### LINKED ISSUE #921: Implement a Hasher based on murmur3

tanruixiang:
## Related Issues
Closes #921 

## Detailed Changes
Be careful not to use `ahash` for non-memory scenarios.
Note that when an upstream dependency also references the same library, their `FEATURES` are merged.
Reference: 
[1] https://github.com/tkaitchuck/aHash/blob/master/README.md?plain=1#L22

### DIFF
--- a/common_util/src/partitioned_lock.rs
+++ b/common_util/src/partitioned_lock.rs
@@ -6,7 +6,7 @@ use std::{
     hash::{Hash, Hasher},
     sync::{Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
-
+/// Warning, do not use this hash in non-memory scenarios.
 use common_types::hash::AHasher;
 /// Simple partitioned `RwLock`
 pub struct PartitionedRwLock<T> {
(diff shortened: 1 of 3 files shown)


common_util/src/partitioned_lock.rs
```diff
@@ -7,6 +7,7 @@ use std::{
     sync::{Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
 
+/// Warning, do not use this hash in non-memory scenarios.
```
jiacai2050, line 10: This warning should be place in where AHasher is defined, not its usage.

Also you can link to its readme to explain its reason.
https://github.com/tkaitchuck/aHash/blob/master/README.md#goals-and-non-goals

jiacai2050 (approved):
LGTM

jiacai2050 (approved):
LGTM

PR #932, opened 2023-05-25
fix: http query support querying partitioned sub-table

chunshao90:
## Related Issues
Closes #

## Detailed Changes
The http api supports querying sub-tables of partition table.

## Test Plan 
Manual test.


Rachelint (approved):
LGTM

PR #936, opened 2023-05-26
refactor: use partition lock for memory cache

tanruixiang:
## Related Issues
Ralated #914 

## Detailed Changes
Add `build_fixed_seed_ahasher` to build fixed  seeds ahasher.
Use PartitionedMutex in `MemCache`.
## Test Plan 
UT.

### DIFF
--- a/common_util/src/partitioned_lock.rs
+++ b/common_util/src/partitioned_lock.rs
@@ -7,7 +7,7 @@ use std::{
     sync::{Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
 
-use common_types::hash::AHasher;
+use common_types::hash::{build_fixed_seed_ahasher, AHasher};
 /// Simple partitioned `RwLock`
 pub struct PartitionedRwLock<T> {
     partitions: Vec<RwLock<T>>,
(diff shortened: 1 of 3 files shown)


components/object_store/src/mem_cache.rs
```diff
@@ -98,44 +54,37 @@ impl MemCache {
...
     fn peek(&self, key: &str) -> Option<Bytes> {
-        let partition = self.locate_partition(key);
-        partition.peek(key)
+        self.inner.lock(&key).peek(key).cloned()
     }
 
     fn insert(&self, key: String, value: Bytes) {
-        let partition = self.locate_partition(&key);
-        partition.insert(key, value);
+        self.inner.lock(&key).put_with_weight(key, value).unwrap();
```
jiacai2050, line 72: This is what it does before
```rs
// don't care error now.
        _ = guard.put_with_weight(key, value);
```

This is what `put_with_weight`'s doc says
```
    /// Puts a key-value pair into cache taking it's weight into account.
    /// If the weight of the new element is greater than what the cache can hold,
    /// it returns the provided key-value pair as an error.

```

common_util/src/partitioned_lock.rs
```diff
@@ -91,6 +93,10 @@ where
     fn get_partition_by_index(&self, index: usize) -> &Mutex<T> {
         &self.partitions[index]
     }
+
+    pub fn get_all_partition(&self) -> &Vec<Mutex<T>> {
```
jiacai2050, line 97:
```
#[cfg(test)]
```

jiacai2050 (approved):
LGTM

PR #940, opened 2023-05-29
fix: write cancel when flush pending write queue

ShiKaiWi:
## Related Issues
Closes #

## Detailed Changes
Currently, the writer for the first request in the pending write queue is responsible for flushing all the writes in the queue, but it can't be ensured to always trigger flush because it may be cancelled when await for the serial exec lock. And this change set replaces the `try_join_all` in the rpc services to avoid such cancellation.

## Test Plan 
Test it manually.

### DIFF
--- a/server/src/grpc/remote_engine_service/mod.rs
+++ b/server/src/grpc/remote_engine_service/mod.rs
@@ -178,6 +178,8 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
                 code: StatusCode::Internal,
                 msg: "fail to run the join task",
             });
+            // The underlying write can't be cancelled, so just ignore the left write
+            // handles (don't abort them) if any error is encountered.
             match write_result {
                 Ok(res) => match res {
                     Ok(resp) => batch_resp.affected_rows += resp.affected_rows,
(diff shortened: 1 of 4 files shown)


server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -162,31 +159,41 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+                    Err(e) => {
+                        error!("Failed to write batches, err:{e}");
+                        batch_resp.header = Some(error::build_err_header(e));
+                        break;
+                    }
+                },
+                Err(e) => {
+                    error!("Failed to write batches, err:{e}");
+                    batch_resp.header = Some(error::build_err_header(e));
+                    break;
```
jiacai2050, line 193: Add comments why we can't abort remaining handles

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -162,31 +159,41 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+                Ok(res) => match res {
+                    Ok(resp) => batch_resp.affected_rows += resp.affected_rows,
+                    Err(e) => {
+                        error!("Failed to write batches, err:{e}");
+                        batch_resp.header = Some(error::build_err_header(e));
+                        break;
+                    }
+                },
+                Err(e) => {
+                    error!("Failed to write batches, err:{e}");
```
MachaelLee, line 193: Make this error log different with L185?

jiacai2050 (approved):
LGTM

server/src/grpc/remote_engine_service/mod.rs
jiacai2050: Doesn't make much sense, log with include line number.

PR #942, opened 2023-05-29
fix: the problem of inconsistent error message returned when auto_create_table fails

chunshao90:
## Related Issues
Closes #

## Detailed Changes
* Fix the problem of inconsistent error message returned when auto_create_table fails.
* Grpc returns more detailed error info.

## Test Plan 
Manual test and CI.


jiacai2050 (approved):
LGTM

PR #943, opened 2023-05-29
fix: too frequent flush

ShiKaiWi:
## Related Issues
Closes #

## Detailed Changes
The condition for checking whether to flush a table whose last flush time is too old is wrong.

### DIFF
--- a/analytic_engine/src/compaction/scheduler.rs
+++ b/analytic_engine/src/compaction/scheduler.rs
@@ -637,6 +637,8 @@ impl ScheduleWorker {
     }
 
     async fn flush_tables(&self) {
+        info!("Scheduled flush all tables begins");
+
         let mut tables_buf = Vec::new();
         self.space_store.list_all_tables(&mut tables_buf);
         let flusher = Flusher {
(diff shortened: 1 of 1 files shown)


analytic_engine/src/compaction/scheduler.rs
```diff
@@ -637,6 +637,8 @@ impl ScheduleWorker {
     }
 
     async fn flush_tables(&self) {
+        info!("Scheduled flush all tables begins");
```
jiacai2050, line 640: The schedule interval is very little, so this may cause lots of logs.

analytic_engine/src/compaction/scheduler.rs
```diff
@@ -648,8 +650,14 @@ impl ScheduleWorker {
         for table_data in &tables_buf {
             let last_flush_time = table_data.last_flush_time();
             if last_flush_time + self.max_unflushed_duration.as_millis_u64()
-                > common_util::time::current_time_millis()
+                < common_util::time::current_time_millis()
```
jiacai2050, line 653: Good catch 👍, no wonder there are so many little sst files.

jiacai2050:
```rs
let current = common_util::time::current_time_millis();
if current > last_flush_time + ... {
}
```
Does this make it more clear?

jiacai2050 (approved):
LGTM

PR #944, opened 2023-05-29
feat: ignore column filter when there are too few values

jiacai2050:
## Related Issues
Closes #

## Detailed Changes
Introduce a new options `min_value_num_when_build_filter`(default 20), which control column filter build.

Ignore build current column filter when distinct column values are less than it.

## Test Plan 
New UT `test_row_group_filter_builder_none_case`

### DIFF
--- a/analytic_engine/src/instance/mod.rs
+++ b/analytic_engine/src/instance/mod.rs
@@ -153,7 +153,7 @@ pub struct Instance {
     /// Write sst max buffer size
     pub(crate) write_sst_max_buffer_size: usize,
     /// Min rows required when build sst filter
-    pub(crate) min_rows_when_build_filter: usize,
+    pub(crate) min_value_num_when_build_filter: usize,
     /// Max retry limit to flush memtables
     pub(crate) max_retry_flush_limit: usize,
     /// Max bytes per write batch
@@ -273,7 +273,7 @@ impl Instance {
             // Do flush in write runtime
             runtime: self.runtimes.write_runtime.clone(),
             write_sst_max_buffer_size: self.write_sst_max_buffer_size,
-            min_rows_when_build_filter: self.min_rows_when_build_filter,
+            min_value_num_when_build_filter: self.min_value_num_when_build_filter,
         }
     }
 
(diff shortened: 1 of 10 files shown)


analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -117,23 +117,35 @@ impl Filter for Xor8Filter {
 
 pub struct RowGroupFilterBuilder {
     builders: Vec<Option<Xor8Builder>>,
+    column_values: Vec<HashSet<Vec<u8>>>,
+    min_value_num_when_build_filter: usize,
 }
 
 impl RowGroupFilterBuilder {
-    pub(crate) fn with_num_columns(num_col: usize) -> Self {
+    pub(crate) fn with_num_columns(num_col: usize, min_value_num_when_build_filter: usize) -> Self {
```
ShiKaiWi, line 125: With a new parameter, maybe we should change this method name to `new`?

analytic_engine/src/sst/parquet/meta_data.rs
```diff
@@ -117,23 +117,35 @@ impl Filter for Xor8Filter {
...
+            column_values: vec![HashSet::new(); num_col],
+            min_value_num_when_build_filter,
         }
     }
 
     pub(crate) fn add_key(&mut self, col_idx: usize, key: &[u8]) {
-        self.builders[col_idx].get_or_insert_default().insert(key)
+        self.builders[col_idx].get_or_insert_default().insert(key);
+        // TODO: remove to_owned?
+        self.column_values[col_idx].insert(key.to_owned());
```
ShiKaiWi, line 136: I guess Xor8Builder can provide an estimate for the number of unique keys, and we can use that instead of an extra `HashSet`.

And I guess a better way is to pre-compute this number of unique keys, and to avoid building xorfilter if it should not. However, maybe we can add metrics to figure out the proportion of the total which doesn't need bloom filter.

analytic_engine/src/instance/mod.rs
```diff
@@ -152,6 +152,8 @@ pub struct Instance {
     pub(crate) replay_batch_size: usize,
     /// Write sst max buffer size
     pub(crate) write_sst_max_buffer_size: usize,
+    /// Min rows required when build sst filter
+    pub(crate) min_value_num_when_build_filter: usize,
```
ShiKaiWi, line 156: After some thoughts, I guess what we should avoid is to build a bloom filter for too many keys, which costs too much storage but provides high false positive rate.

PR #946, opened 2023-05-30
feat: remove duplicated schema covert from arrow schema

Rachelint:
## Related Issues
Closes #

## Detailed Changes
+ Add a `RecordBatchBuilder` which will cache the schema coverted from arrow's.
+ Build inner `RecordBatch` using `RecordBatchBuilder`.

## Test Plan 
Test by exist tests.

### DIFF
--- a/common_types/src/record_batch.rs
+++ b/common_types/src/record_batch.rs
@@ -269,13 +269,21 @@ impl RecordBatch {
     }
 }
 
+/// Converter from arrow record batches to inner's
+///
+/// Converting arrow schema to inner's is expensive. For eliminate the
+/// duplicated converting, we define such converter to cache the first converted
+/// one and reuse it in later converting.
+///
+/// NOTE: must ensure the converted arrow record batches has the same schema
+/// before deciding to use this converter.
 #[derive(Default)]
-pub struct RecordBatchBuilder {
+pub struct CachedRecordBatchesConverter {
     schema_cache: Option<RecordSchema>,
 }
 
-impl RecordBatchBuilder {
-    pub fn build(&mut self, arrow_record_batch: ArrowRecordBatch) -> Result<RecordBatch> {
+impl CachedRecordBatchesConverter {
+    pub fn convert(&mut self, arrow_record_batch: ArrowRecordBatch) -> Result<RecordBatch> {
         let schema = match &self.schema_cache {
             Some(schema) => schema.clone(),
             None => {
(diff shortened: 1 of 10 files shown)


common_types/src/record_batch.rs
```diff
@@ -269,19 +269,30 @@ impl RecordBatch {
...
+    schema_cache: Option<RecordSchema>,
+}
 
-    fn try_from(arrow_record_batch: ArrowRecordBatch) -> Result<Self> {
-        let record_schema =
-            RecordSchema::try_from(arrow_record_batch.schema()).context(ConvertArrowSchema)?;
+impl RecordBatchBuilder {
+    pub fn build(&mut self, arrow_record_batch: ArrowRecordBatch) -> Result<RecordBatch> {
+        let schema = match &self.schema_cache {
+            Some(schema) => schema.clone(),
```
chunshao90, line 280: Is it possible that the schema is not consistent?

common_types/src/record_batch.rs
```diff
@@ -269,19 +269,30 @@ impl RecordBatch {
...
+#[derive(Default)]
+pub struct RecordBatchBuilder {
+    schema_cache: Option<RecordSchema>,
+}
 
-    fn try_from(arrow_record_batch: ArrowRecordBatch) -> Result<Self> {
-        let record_schema =
-            RecordSchema::try_from(arrow_record_batch.schema()).context(ConvertArrowSchema)?;
+impl RecordBatchBuilder {
+    pub fn build(&mut self, arrow_record_batch: ArrowRecordBatch) -> Result<RecordBatch> {
```
jiacai2050, line 278: It's wired to name this `build`, usually build is only called once.

`convert` maybe a better name.

common_types/src/record_batch.rs
Rachelint: Now,  such usecase not exist.
But it is needed to be considered actually, I will add some comments about it.

Rachelint: Some comments about are added.

common_types/src/record_batch.rs
Rachelint: Done.

common_types/src/record_batch.rs
```diff
@@ -269,27 +269,63 @@ impl RecordBatch {
...
+///
+/// Converting arrow schema to inner's is expensive. For eliminate the
+/// duplicated converting, we define such converter to cache the first converted
+/// one and reuse it in later converting.
+///
+/// NOTE: must ensure the converted arrow record batches has the same schema
+/// before deciding to use this converter.
+#[derive(Default)]
+pub struct CachedRecordBatchesConverter {
+    schema_cache: Option<RecordSchema>,
```
jiacai2050, line 282: How about init this field in `fn new()`, then `convert` could take `&self` directly.

PR #948, opened 2023-05-30
feat: make remote client configurable

MachaelLee:
## Related Issues
Closes #

## Detailed Changes

* add compression config in remote client config 
* add remote client config in server config

### DIFF
--- a/components/arrow_ext/src/ipc.rs
+++ b/components/arrow_ext/src/ipc.rs
@@ -62,7 +62,7 @@ impl Default for CompressOptions {
     fn default() -> Self {
         Self {
             compress_min_length: DEFAULT_COMPRESS_MIN_LENGTH,
-            method: CompressionMethod::Zstd,
+            method: CompressionMethod::None,
         }
     }
 }
--- a/components/arrow_ext/src/ipc.rs
+++ b/components/arrow_ext/src/ipc.rs
@@ -62,7 +62,7 @@ impl Default for CompressOptions {
     fn default() -> Self {
         Self {
             compress_min_length: DEFAULT_COMPRESS_MIN_LENGTH,
-            method: CompressionMethod::None,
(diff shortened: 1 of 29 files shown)


table_engine/src/remote/model.rs
```diff
@@ -159,15 +157,16 @@ impl TryFrom<ceresdbproto::remote_engine::WriteBatchRequest> for WriteBatchReque
     }
 }
 
-impl TryFrom<WriteBatchRequest> for ceresdbproto::remote_engine::WriteBatchRequest {
-    type Error = Error;
-
-    fn try_from(batch_request: WriteBatchRequest) -> std::result::Result<Self, Self::Error> {
+impl WriteBatchRequest {
+    pub fn batch_convert_to_pb(
```
Rachelint, line 161: convert_write_batch_to_pb?

MachaelLee: Ok, fixed.

components/arrow_ext/src/ipc.rs
```diff
@@ -48,13 +51,22 @@ pub struct RecordBatchesEncoder {
...
     /// The minimum length of the payload to be compressed.
     pub compress_min_length: usize,
     pub method: CompressionMethod,
 }
 
+impl Default for CompressOptions {
+    fn default() -> Self {
+        Self {
+            compress_min_length: DEFAULT_COMPRESS_MIN_LENGTH,
+            method: CompressionMethod::None,
```
jiacai2050, line 65: Those value default to zstd before, why change to none?

MachaelLee: In my opinion, in most scenarios, network traffic is not the limiting factor for a system, so there is no need to default to using zstd compression.

jiacai2050 (approved):
LGTM

PR #954, opened 2023-05-31
refactor: open shard impl of `TableEngine`

Rachelint:
## Related Issues
Closes #

## Detailed Changes
- Define a `ShardOpener` to manage shard's opening.
- Make use of it for support later shard based recovery.
- The responsibilities among `TableOperator`, `Instacne`, `ShardOpener` are like:
  - `TableOperator`: install the successfully opened tables to their schemas, print error log for the faileds.
  - `Instance`: insert the successfully opened tables to their spaces, and the faileds to spaces' failed maps.
  - `ShardOpener`: do the tables(on same shard) opening work.
## Test Plan 
Test by exist tests.


analytic_engine/src/instance/open.rs
```diff
@@ -138,71 +139,265 @@ impl Instance {
...
+            let space = &table_ctx.space;
+            let table_id = table_ctx.table_def.id;
+            let state = if let Some(table_data) = space.find_table_by_id(table_id) {
+                TableOpenState::Success(Some(SpaceAndTable::new(space.clone(), table_data)))
+            } else {
+                TableOpenState::RecoverTableMeta(RecoverTableMetaContext {
+                    table_def: table_ctx.table_def,
+                    space: table_ctx.space,
+                })
+            };
```
ShiKaiWi, line 226: Why does the tables to open have these two states? Which one state is normal? Shall we take the other state as a mistake?

analytic_engine/src/instance/open.rs
```diff
@@ -138,71 +139,265 @@ impl Instance {
...
     ///
     /// Return None if no meta data is found for the table.
-    async fn recover_table_meta(self: &Arc<Self>, request: &OpenTableRequest) -> Result<()> {
-        info!("Instance recover table:{} meta begin", request.table_id);
+    async fn recover_single_table_meta(
+        manifest: &dyn Manifest,
+        shard_id: ShardId,
+        table_def: &TableDef,
+    ) -> Result<()> {
+        info!("Instance recover table:{} meta begin", table_def.id);
```
ShiKaiWi, line 375: How about printing the table name too?

analytic_engine/src/instance/open.rs
```diff
@@ -138,71 +139,265 @@ impl Instance {
...
+    pub space: SpaceRef,
+}
+
+#[derive(Debug)]
+enum TableOpenState {
+    RecoverTableMeta(RecoverTableMetaContext),
+    RecoverTableData(RecoverTableDataContext),
+    Failed(crate::instance::engine::Error),
+    Success(Option<SpaceAndTable>),
+}
```
ShiKaiWi, line 179: How about putting it:
```suggestion
enum TableOpenStage {
    Ready(Option<SpaceAndTable>),
    TableMetaRecovered(RecoverTableMetaContext),
    TableDataRecovered(RecoverTableDataContext),
    Failed(crate::instance::engine::Error),
    Success(Option<SpaceAndTable>),
}
```

The reason why I'm in favor of `Stage` is that there is no complex transition between these states and `Stage` suits to the linear transition.

catalog/src/lib.rs
```diff
@@ -50,13 +50,16 @@ pub enum Error {
...
+        msg: Option<String>,
+        source: GenericError,
+    },
 
     // Fixme: Temporarily remove the stack information, otherwise you will encounter a
     // segmentation fault.
-    #[snafu(display("Failed to operate table, msg:{}.\n", msg))]
-    TableOperatorNoCause { msg: String },
+    #[snafu(display("Failed to operate table, msg:{:?}.\n", msg))]
+    TableOperatorNoCause { msg: Option<String> },
```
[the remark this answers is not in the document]
Rachelint: I think passing a empty `String` is strange.
In fact, I want to make it an auto generated `Package` but has no idea about how to impl now...

ShiKaiWi (approved):
LGTM

PR #960, opened 2023-06-01
fix: capacity should equal to total / part_num

jiacai2050:
## Rationale

Current mem cache will consume more memory than expected.

## Detailed Changes
Fix the wrong capacity of a partition cache.

## Test Plan
Add UT `test_mem_cache_capacity`


ShiKaiWi (approved):
LGTM

PR #961, opened 2023-06-01
chore: add log for stream read of remote engine service

ShiKaiWi:
## Rationale
Information of partition table query routed to a specific ceresdb server is missing.

## Detailed Changes
Add logs for the remote engine read.


Rachelint (approved):
LGTM

PR #964, opened 2023-06-02
fix: missing metrics when using chain iterator

ShiKaiWi:
## Rationale
The metrics about scan is missing when chain iterator is used underlying.

## Detailed Changes
Do metrics collection when using chain iterator.

## Test Plan
Manually.


jiacai2050 (approved):
LGTM

PR #975, opened 2023-06-07
refactor: optimize sst iterator and filter build to consume less CPU

zouxiang1993:
## Rationale
When doing benchmark, sst iterator and xor filter build cost too much CPU.

## Detailed Changes
- Change RowViewOnBatchColumnIter's item from Datum to DatumView

## Test Plan
I will do benchmark in my test env.

### DIFF
--- a/common_types/src/datum.rs
+++ b/common_types/src/datum.rs
@@ -1010,6 +1010,76 @@ impl<'a> DatumView<'a> {
             DatumView::Time(_) => DatumKind::Time,
         }
     }
+
+    pub fn do_with_bytes<F>(&self, mut f: F)
+        where
+            F: FnMut(&[u8]),
+    {
+        match self {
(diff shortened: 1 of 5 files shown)


common_types/src/datum.rs
```diff
@@ -1010,6 +1010,76 @@ impl<'a> DatumView<'a> {
             DatumView::Time(_) => DatumKind::Time,
         }
     }
+
+    pub fn do_with_bytes<F>(&self, mut f: F)
```
jiacai2050, line 1014: Does this mean Datum's do_with_bytes could be deleted?

jiacai2050: Looks other places could use this fn, so keep it for now.

jiacai2050 (approved):
LGTM

PR #977, opened 2023-06-08
refactor: add generic support to generate hasher

tanruixiang:
## Rationale
Add generic support to generate hasher 

## Detailed Changes
Add generic support to generate hasher  for all `PartitionedLock`.

## Test Plan
Ut.


jiacai2050 (approved):
LGTM

PR #986, opened 2023-06-09
chore: reduce kafka logs

jiacai2050:
## Rationale
There are too many logs.

## Test Plan
No need

### DIFF
--- a/wal/src/message_queue_impl/region.rs
+++ b/wal/src/message_queue_impl/region.rs
@@ -18,7 +18,7 @@ use util::*;
 use crate::{
     kv_encoder::CommonLogEncoding,
     log_batch::{LogEntry, LogWriteBatch},
-    manager::{self},
+    manager,
     message_queue_impl::{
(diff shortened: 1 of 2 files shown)


analytic_engine/Cargo.toml
```diff
@@ -56,4 +56,5 @@ common_util = { workspace = true, features = ["test"] }
 env_logger = { workspace = true }
 pin-project-lite = { workspace = true }
 rand = { workspace = true }
+tempfile = { workspace = true }
```
jiacai2050, line 59: By put `tempfile` here, we can test `analytic_engine` alone, otherwise it will throw errors when `cargo test` in analytic_engine dir:
```
error[E0433]: failed to resolve: use of undeclared crate or module `tempfile`
   --> analytic_engine/src/tests/util.rs:557:19
    |
557 |         let dir = tempfile::tempdir().unwrap();
    |                   ^^^^^^^^ use of undeclared crate or module `tempfile`
```

Rachelint (approved):
LGTM

PR #989, opened 2023-06-12
refactor: profiling

chunshao90:
## Detailed Changes
Refactor profiling api.

## Test Plan
Manual test and CI.

### DIFF
--- a/components/profile/src/lib.rs
+++ b/components/profile/src/lib.rs
@@ -1,6 +1,6 @@
 // Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
 
-//! Memory profiler for running application based on jemalloc features.
+//! Profiler for running application.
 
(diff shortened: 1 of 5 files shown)


components/profile/src/lib.rs
```diff
@@ -109,7 +111,7 @@ impl Profiler {
             .create(true)
             .write(true)
             .truncate(true)
-            .open(PROFILE_OUTPUT_FILE_PATH)
+            .open(PROFILE_MEM_OUTPUT_FILE_PATH)
```
Rachelint, line 114: Expose the path as config or env?

jiacai2050 (approved):
LGTM

components/profile/src/lib.rs
jiacai2050: Will do this if required next time.

PR #991, opened 2023-06-13
refactor: avoid grpc forwarding twice

### LINKED ISSUE #984: Prevent continuous forwarding between multiple nodes
### Describe This Problem

Now the proxy module does not check the number of forwarding times when forwarding, which may cause continuous forwarding between multiple nodes.

### Proposal

Add a parameter to the headers of grpc to mark that it has been forwarded.

### Additional Context

_No response_

baojinri:
## Rationale
Close #984 

## Detailed Changes
Add a parameter to the headers of grpc to mark that it has been forwarded.

## Test Plan
Existing tests

### DIFF
--- a/proxy/src/forward.rs
+++ b/proxy/src/forward.rs
@@ -21,7 +21,7 @@ use tonic::{
     transport::{self, Channel},
 };
 
-use crate::FORWARDED;
+use crate::FORWARDED_FROM;
 
 #[derive(Debug, Snafu)]
 pub enum Error {
@@ -71,10 +71,7 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
-    #[snafu(display(
-        "Request should not be forwarded multiple times, forward endpoint:{}",
-        endpoint
-    ))]
+    #[snafu(display("Request should not be forwarded twice, forward from:{}", endpoint))]
     ForwardedErr { endpoint: String },
 }
 
@@ -192,7 +189,7 @@ pub struct ForwardRequest<Req> {
     pub schema: String,
     pub table: String,
     pub req: tonic::Request<Req>,
-    pub forwarded: bool,
+    pub forwarded_from: Option<String>,
 }
 
 impl Forwarder<DefaultClientBuilder> {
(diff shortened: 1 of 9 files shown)


proxy/src/grpc/sql_query.rs
```diff
@@ -92,32 +98,32 @@ impl<Q: QueryExecutor + 'static> Proxy<Q> {
...
-                ForwardResult::Local => req,
-            },
-            None => req,
-        };
+
+        if let Some(ForwardResult::Forwarded(resp)) =
+            self.clone().maybe_forward_stream_sql_query(req).await
+        {
+            return resp;
+        }
```
baojinri, line 119: add  note

proxy/src/forward.rs
```diff
@@ -68,6 +70,12 @@ pub enum Error {
         source: tonic::transport::Error,
         backtrace: Backtrace,
     },
+
+    #[snafu(display(
+        "Request should not be forwarded multiple times, forward endpoint:{}",
```
ShiKaiWi, line 75:
```suggestion
        "Request should not be forwarded multiple times, forward from:{}",
```

proxy/src/forward.rs
```diff
@@ -310,6 +326,15 @@ impl<B: ClientBuilder> Forwarder<B> {
...
             endpoint, req,
         );
+
+        if forwarded {
+            let endpoint = endpoint.to_string();
+            return ForwardedErr { endpoint }.fail();
+        }
+
+        // mark forwarded
+        req.metadata_mut().insert(FORWARDED, "".parse().unwrap());
```
ShiKaiWi, line 336: Maybe we should insert the local endpoint as the value of `Forwarded-From`

proxy/src/lib.rs
```diff
@@ -22,6 +22,8 @@ pub mod schema_config_provider;
 mod util;
 mod write;
 
+pub const FORWARDED: &str = "forwarded";
```
ShiKaiWi, line 25: How about this name?
```suggestion
pub const FORWARDED_FROM: &str = "Forwarded-From";
```
And move this definition to `the forward.rs`.

ShiKaiWi (approved):
LGTM

PR #996, opened 2023-06-15
chore: add integration test about recovery

Rachelint:
## Rationale
Part of #799 
Now we run the test about recovery manually that is so tired, this pr add this into integration tests which will be run automatically in ci.

## Detailed Changes
+ Add  integration test about recovery.
+ Add above test to ci.

## Test Plan
None.


ShiKaiWi (approved):
LGTM

PR #1000, opened 2023-06-19
fix: add page index for metadata

MachaelLee:
## Rationale
Now there is no page index in the `meta_data`, we should build page index if we want to use row selection.

## Detailed Changes
* build page index for `meta_data`
* add some debug log

### DIFF
--- a/analytic_engine/src/sst/meta_data/cache.rs
+++ b/analytic_engine/src/sst/meta_data/cache.rs
@@ -1,4 +1,4 @@
-// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+// Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
 
 use std::{
     fmt::Debug,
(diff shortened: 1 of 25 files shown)


analytic_engine/src/sst/meta_data/cache.rs
```diff
@@ -22,8 +22,8 @@ pub type MetaCacheRef = Arc<MetaCache>;
 pub struct MetaData {
     /// The extended information in the parquet is removed for less memory
     /// consumption.
-    parquet: parquet_ext::ParquetMetaDataRef,
-    custom: ParquetMetaDataRef,
+    pub parquet: parquet_ext::ParquetMetaDataRef,
+    pub custom: ParquetMetaDataRef,
```
ShiKaiWi, line 26: I guess two methods called `parquet()` and `custom` has provided read-only access already.

analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -281,13 +278,23 @@ impl<'a> Reader<'a> {
...
+                parquet_metadata.clone(),
+            );
             let mut builder = ParquetRecordBatchStreamBuilder::new(object_store_reader)
                 .await
                 .with_context(|| ParquetError)?;
+
             let row_selection =
                 self.build_row_selection(arrow_schema.clone(), &chunk, parquet_metadata)?;
+
+            debug!(
```
jiacai2050, line 293: Delete this debug?

jiacai2050 (approved):
LGTM

PR #1004, opened 2023-06-19
fix: arrow meta data is lost when decode custom meta data

ShiKaiWi:
## Rationale
The metadata for arrow schema is encoded into the parquet file. However, this part is lost when building our custom metadata.

## Detailed Changes
Keep the other metadata in the parquet metadata after extracting our custom metadata.

## Test Plan
Add unit test `test_arrow_meta_data` for it.

### DIFF
--- a/analytic_engine/src/sst/meta_data/cache.rs
+++ b/analytic_engine/src/sst/meta_data/cache.rs
@@ -40,9 +40,8 @@ impl MetaData {
             .key_value_metadata()
             .context(KvMetaDataNotFound)?;
 
-        ensure!(!kv_metas.is_empty(), KvMetaDataNotFound,);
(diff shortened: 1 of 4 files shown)


analytic_engine/src/sst/meta_data/cache.rs
```diff
@@ -62,7 +70,7 @@ impl MetaData {
                 file_meta_data.num_rows(),
                 file_meta_data.created_by().map(|v| v.to_string()),
                 // Remove the key value metadata.
-                None,
+                Some(other_kv_metas),
```
tanruixiang, line 73: The comment can be changed.
Here if `other_kv_metas` is empty, do we need to pass a `None`?

tanruixiang:
LGTM

tanruixiang (approved):

PR #1005, opened 2023-06-20
feat: use instead forked `rskafka` to support limited retry

Rachelint:
## Rationale
Part of #799 
We use `rskafka` as our kafka client.
However I found it will retry without limit even though kafka service is unavailable...
(see [https://github.com/influxdata/rskafka/issues/65](https://github.com/influxdata/rskafka/issues/65))
Worse, I found `rskafka` is almostis no longer maintained...

For quick fix, I forked it to support limited retry.

## Detailed Changes
+ Use the instead forked `rskafka`(supporting limited retry).
+ Add more logs in recovery path for better debugging.

## Test Plan
Test manually.

### DIFF
--- a/analytic_engine/src/manifest/details.rs
+++ b/analytic_engine/src/manifest/details.rs
@@ -265,10 +265,6 @@ where
         }
 
         if has_logs {
-            debug!(
-                "Manifest recover with only logs, table_id:{}, space_id:{}",
-                self.table_id, self.space_id
-            );
             Ok(Some(Snapshot {
                 end_seq: latest_seq,
                 data: manifest_data_builder.build(),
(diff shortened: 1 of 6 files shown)


analytic_engine/src/manifest/details.rs
```diff
@@ -505,6 +527,8 @@ impl Manifest for ManifestImpl {
             self.table_meta_set.apply_edit_to_table(request)?;
         }
 
+        info!("Manifest recover finish, request:{:?}", load_req);
```
ShiKaiWi, line 530:
```suggestion
        info!("Manifest recover finish, request:{load_req:?}");
```

analytic_engine/src/manifest/details.rs
```diff
@@ -253,11 +265,19 @@ where
         }
 
         if has_logs {
+            debug!(
+                "Manifest recover with only logs, table_id:{}, space_id:{}",
+                self.table_id, self.space_id
+            );
```
ShiKaiWi, line 271: Maybe we can just print one log instead of L268 and L277:
```
            debug!(
                "Manifest finish recover without prev snapshot, has_logs:{has_logs}, table_id:{}, space_id:{}",
                self.table_id, self.space_id
            );
```

ShiKaiWi (approved):
LGTM

PR #1006, opened 2023-06-20
feat: fork `Column` and `ColumnData` from influxdb_iox

chunshao90:
## Detailed Changes
Introduce `Column` which is a columnar storage to optimize the writing process later.

## Test Plan
CI.


Rachelint (approved):
LGTM

PR #1013, opened 2023-06-21
fix: Cherry-pick commits removed in PR:1000

MachaelLee:
## Rationale
In PR:https://github.com/CeresDB/ceresdb/pull/1000, some commits were removed. 

## Detailed Changes
* cherry-pick these commits


ShiKaiWi:
@MachaelLee I guess we should do revert on #1000, cherry-pick in this PR seems not perfect, and I'm not sure whether any other problems will be introduced.

PR #1014, opened 2023-06-21
chore: use pr branch to actions

### LINKED ISSUE #1012: The current CI process merges the code from the Pr branch into the main branch for testing
### Describe this problem

The current CI process merges the code in the pr branch with the main branch. This doesn't reasonable and can lead to extremely confusing results when multiple contributors make changes to related code at the same time.
We need to use the code from the pr branch for CI and the project maintainer to ensure that the pr code is newer than the main branch code before it can be merged.
This has the advantage that each pr is isolated and will ensure that the pr does not produce different ci results due to changes in the main branch.
I recommend using this setting:
https://i.stack.imgur.com/doAEL.png
### Server version

9a9c0f79116dff6df2a09fc8766e6952fd5f4432

### Steps to reproduce
Any CI before 9a9c0f79116dff6df2a09fc8766e6952fd5f4432

### Expected behavior

The CI process runs only on pr's branch.

### Additional Information

_No response_

tanruixiang:
## Rationale
close #1012

## Detailed Changes
[Checkout pull request HEAD commit instead of merge commit](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token)

## Test Plan
At the end of checkout@3 the commit id will be printed,  can check if it is correct by this id.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
+        with:
           submodules: true
       - run: |
           rustup set auto-self-update disable
@@ -90,6 +92,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
+        with:
           submodules: true
       - run: |
           rustup set auto-self-update disable
@@ -129,7 +133,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
         with:
           submodules: true
       - name: Cache Rust Dependencies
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -40,6 +40,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
jiacai2050, line 43: What this would be if we trigger CI manually?

jiacai2050:
```
${{ github.event.pull_request.head.sha || github.sha}}
```
This may more accurate.
- https://github.com/apache/arrow-datafusion/blob/main/.github/workflows/dev.yml#L22

ShiKaiWi:
I guess current behavior (pick the commits from PR branch to main and then run the ci) is reasonable. If only run the ci on the PR branch, the merged main branch can't be ensured correctness.

tanruixiang:
> I guess current behavior (pick the commits from PR branch to main and then run the ci) is reasonable. If only run the ci on the PR branch, the merged main branch can't be ensured correctness.

Yes, as the issue says, it needs to be paired with setting:`require branches to be up to date before mergin`.
Maybe we need to find a better way to solve this problem.

PR #1021, opened 2023-06-25
chore: bump ceresdbproto

jiacai2050:
## Rationale
See https://github.com/CeresDB/ceresdbproto/pull/81

After this, developers can set `export CERESDBPROTO_ENABLE_VENDORED=false` to use protoc on their host.

## Test Plan
No need.


Rachelint (approved):
LGTM

PR #1023, opened 2023-06-25
cherry-pick: arrow meta data is lost when decode custom meta data (#1004)

chunshao90:
## Rationale
The metadata for arrow schema is encoded into the parquet file. However, this part is lost when building our custom metadata.

## Detailed Changes
Keep the other metadata in the parquet metadata after extracting our custom metadata.

## Test Plan
Add unit test `test_arrow_meta_data` for it.


ShiKaiWi:
In favor of #1026

PR #1028, opened 2023-06-26
fix: `test_suggest_duration_and_ranges()` occasional fail.

zouxiang1993:
## Rationale
Timestamp::now() produces a random timestamp which leads to occasional test fail.

## Detailed Changes
Use a fixed timestamp.

### DIFF
--- a/analytic_engine/src/sampler.rs
+++ b/analytic_engine/src/sampler.rs
@@ -382,7 +382,7 @@ mod tests {
             &[(0, 2 * HOUR_MS as i64)],
         );
 
-        let now = 1687747471000i64;
+        let now = 1672502400000i64;
         let now_ts = Timestamp::new(now);
         let sec_ms_i64 = SEC_MS as i64;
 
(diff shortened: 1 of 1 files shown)


analytic_engine/src/sampler.rs
```diff
@@ -382,8 +382,8 @@ mod tests {
             &[(0, 2 * HOUR_MS as i64)],
         );
 
-        let now_ts = Timestamp::now();
-        let now = now_ts.as_i64();
+        let now = 1687747471000i64;
+        let now_ts = Timestamp::new(now);
```
tanruixiang, line 386:
```suggestion
        let now = 1672502400000i64;
        let now_ts = Timestamp::new(now);
```

tanruixiang (commented):
Thank you for your contribution. I recommend using a more debug-friendly timestamp, like `2023-1-1 00:00:00`

tanruixiang (approved):
LGTM

PR #1030, opened 2023-06-26
feat: support hex literal

ShiKaiWi:
## Rationale
Currently, the underlying storage supports binary data type. However, during the parsing and planning, the binary data is not supported.

## Detailed Changes
Support binary data type during parsing and planning stage.

## Test Plan
New unit tests and integration tests.


chunshao90 (approved):
LGTM

PR #1033, opened 2023-06-26
feat: expose more rocksdb options

zouxiang1993:
## Rationale
Now rocksdb as the wal, it is easy to become the bottleneck of write.

## Detailed Changes
1. expose more rocksdb options to avoid write stall
2. introduce rocksdb's FIFO compaction style, which makes rocksdb looks like a message queue. 
    (FIFO is more suitable for time-series data, maybe it will become the default option in the future?)
## Test Plan
I will test it in my test env.

### DIFF
--- a/wal/src/rocks_impl/config.rs
+++ b/wal/src/rocks_impl/config.rs
@@ -9,6 +9,11 @@ use serde::{Deserialize, Serialize};
 pub struct Config {
     pub max_background_jobs: i32,
     pub enable_statistics: bool,
+    pub write_buffer_size: u64,
+    pub max_write_buffer_number: i32,
+    pub level_zero_file_num_compaction_trigger: i32,
+    pub level_zero_slowdown_writes_trigger: i32,
+    pub level_zero_stop_writes_trigger: i32,
 }
 
 impl Default for Config {
@@ -18,6 +23,11 @@ impl Default for Config {
             // https://github.com/facebook/rocksdb/blob/v6.4.6/include/rocksdb/options.h#L537
             max_background_jobs: 2,
             enable_statistics: false,
+            write_buffer_size: 64 << 20,
+            max_write_buffer_number: 2,
+            level_zero_file_num_compaction_trigger: 4,
+            level_zero_slowdown_writes_trigger: 20,
+            level_zero_stop_writes_trigger: 36,
         }
     }
 }
--- a/wal/src/rocks_impl/config.rs
+++ b/wal/src/rocks_impl/config.rs
@@ -15,6 +15,7 @@ pub struct Config {
     pub level_zero_file_num_compaction_trigger: i32,
     pub level_zero_slowdown_writes_trigger: i32,
     pub level_zero_stop_writes_trigger: i32,
+    pub fifo_compaction_max_table_files_size: u64,
 }
 
 impl Default for Config {
@@ -30,6 +31,7 @@ impl Default for Config {
             level_zero_file_num_compaction_trigger: 4,
             level_zero_slowdown_writes_trigger: 20,
             level_zero_stop_writes_trigger: 36,
+            fifo_compaction_max_table_files_size: 0, // default is 1G, use 0 to disable fifo
         }
     }
 }
--- a/wal/src/rocks_impl/config.rs
+++ b/wal/src/rocks_impl/config.rs
@@ -2,6 +2,7 @@
 
 //! RocksDB Config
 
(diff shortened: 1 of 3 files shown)


wal/src/rocks_impl/config.rs
```diff
@@ -7,17 +7,31 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(default)]
 pub struct Config {
+    pub max_subcompactions: u32,
     pub max_background_jobs: i32,
     pub enable_statistics: bool,
+    pub write_buffer_size: u64,
```
ShiKaiWi, line 13: In favor of `ReadableSize` here.

wal/src/rocks_impl/config.rs
```diff
@@ -7,17 +7,31 @@ use serde::{Deserialize, Serialize};
...
+    pub max_write_buffer_number: i32,
+    pub level_zero_file_num_compaction_trigger: i32,
+    pub level_zero_slowdown_writes_trigger: i32,
+    pub level_zero_stop_writes_trigger: i32,
+    pub fifo_compaction_max_table_files_size: u64,
```
ShiKaiWi, line 18: In favor of `ReadableSize` here.

wal/src/rocks_impl/manager.rs
```diff
@@ -566,7 +621,34 @@ impl Builder {
...
+        if let Some(v) = self.fifo_compaction_max_table_files_size {
+            if v > 0 {
+                let mut fifo_opts = FifoCompactionOptions::new();
+                fifo_opts.set_max_table_files_size(v);
+                cf_opts.set_fifo_compaction_options(fifo_opts);
+                cf_opts.set_compaction_style(DBCompactionStyle::Fifo);
+            }
+        }
+
+        let db = DB::open_cf(rocksdb_config, &self.wal_path, vec![("default", cf_opts)])
```
chunshao90, line 651: Define a const or use the const in rocksdb?

wal/src/rocks_impl/config.rs
```diff
@@ -7,17 +7,31 @@ use serde::{Deserialize, Serialize};
...
 #[serde(default)]
 pub struct Config {
+    pub max_subcompactions: u32,
     pub max_background_jobs: i32,
     pub enable_statistics: bool,
+    pub write_buffer_size: u64,
+    pub max_write_buffer_number: i32,
+    pub level_zero_file_num_compaction_trigger: i32,
+    pub level_zero_slowdown_writes_trigger: i32,
+    pub level_zero_stop_writes_trigger: i32,
```
ShiKaiWi, line 17: Can these fields be negative? If they can, maybe we should add some comments to explain the negative values meaning. Otherwise, how about using usize here?

wal/src/rocks_impl/manager.rs
zouxiang1993: fixed

wal/src/rocks_impl/config.rs
zouxiang1993: fixed

wal/src/rocks_impl/config.rs
zouxiang1993: fixed

wal/src/rocks_impl/config.rs
zouxiang1993: The datatypes is same as rocksdb itself. I add some comments.

wal/src/rocks_impl/manager.rs
```diff
@@ -566,11 +623,42 @@ impl Builder {
...
+                cf_opts.set_fifo_compaction_options(fifo_opts);
+                cf_opts.set_compaction_style(DBCompactionStyle::Fifo);
+            }
+        }
+
+        let db = DB::open_cf(
+            rocksdb_config,
+            &self.wal_path,
+            vec![(ROCKSDB_DEFAULT_COLUMN_FAMILY, cf_opts)],
+        )
```
ShiKaiWi, line 657: Maybe we can use `ColumnFamilyDescriptor::default()` to create the descriptor of default column family:
```suggestion
        let default_cfd = {
            let mut cfd = ColumnFamilyDescriptor::default();
            cfd.options = cf_opts;
            cfd
         };
        let db = DB::open_cf(
            rocksdb_config,
            &self.wal_path,
            vec![default_cfd],
        )
```

ShiKaiWi (approved):
LGTM

Issue #987, opened 2023-06-12
Improve compaction


Rachelint:
### Describe This Problem

Now the design of compaction in ceresdb is still so rough, we should make more efforts in it.
There are several improvements that can be made in the following areas:
+ Compaction strategy. Now we just impl TWSC actually, we define a level 1 but do nothing special for it.
+ How to do compaction more efficiently. Speed of compaction may important equally important to strategy.
+ Metrics and tests. We should have ways to check the correctness and effectiveness(especially in query improvement) about our compaction strategy.

### 1. Compaction strategy
- [ ] Introduce `score mechanism` to integrate multiple rules.
- [ ] Consider sequence(wal) when picking compacting files to ensure the correctness.
- [ ] Eliminate time range overlap of ssts in level 1.
- [ ] Take priority of respective table in consideration.

### 2. Performace of compaction
- [ ] Optimize sst iterator and filter build to consume less CPU. #975 

### 3. Metrics and tests
- [ ] Emulator for compaction strategy inspired by [iox](https://github.com/influxdata/influxdb_iox/issues/7687) 
- [ ] Add metrics (like `read amplification`, `write  amplification`, `space  amplification`) to check the effectiveness of the strategy.

jiacai2050:
>  Add metrics (like read amplification, write  amplification, space  amplification) to check the effectiveness of the strategy.

Current codebase already have basic metrics for compact:
1. Input sst size/row num
2. Output sst size/row num

PR #1041, opened 2023-06-29
fix: compaction support pick by max_seq

jiacai2050:
## Rationale
Part of #987.

Current implementation will compact by file size, max_seq is not considered, this may cause data corruption
in corner case, eg:
- sst1, max_seq:10, PK1=10
- sst2, max_seq:11, PK1=9
- sst3, max_seq:12, no PK1

If compact pick sst1 and sst3, and output sst4, its max_seq will be 12, now PK1 exists in two files:
- sst2, max_seq:11, PK1=9
- sst4, max_seq:12, PK1=10

That's to say, PK1's value is 10 now, which is wrong value(9 is right).

## Detailed Changes
When do compaction, first sort sst by max_seq desc, then only pick adjacent ssts, the original issue is fixed in this way.
At the same time picked ssts are ensured to meet other requirements such as `min_threshold`, `max_threshold`, `max_input_size`.

## Test Plan
UT and manually.

### DIFF
--- a/analytic_engine/src/compaction/picker.rs
+++ b/analytic_engine/src/compaction/picker.rs
@@ -176,28 +176,28 @@ fn trim_to_threshold(
         .collect()
 }
 
-// Remove this when pick_by_seq is stable.
+/// Origin solution will only consider file size, but this will cause data
+/// corrupt, see https://github.com/CeresDB/ceresdb/pull/1041
+///
+/// So we could only compact files with adjacent seq, or ssts without
+/// overlapping key range among them. Currently solution is relative simple,
+/// only pick adjacent sst. Maybe a better, but more complex solution could be
+/// introduced later.
+// TODO: Remove this function when pick_by_seq is stable.
 fn prefer_pick_by_seq() -> bool {
-    std::env::var("CERESDB_SIZE_PICKER_PREFER_BY_SEQ").unwrap_or_else(|_| "true".to_string())
-        == "true"
+    std::env::var("CERESDB_COMPACT_PICK_BY_SEQ").unwrap_or_else(|_| "true".to_string()) == "true"
 }
 
 /// Size tiered compaction strategy
 /// See https://github.com/jeffjirsa/twcs/blob/master/src/main/java/com/jeffjirsa/cassandra/db/compaction/SizeTieredCompactionStrategy.java
 pub struct SizeTieredPicker {
-    // Origin solution will only consider file size, but this will cause data corrupt, see
-    // https://github.com/CeresDB/ceresdb/pull/1041
-    //
-    // So we could only compact files with adjacent seq, or ssts without overlapping key range
-    // among them. Currently solution is relative simple, only pick adjacent sst.
-    // Maybe a better, but more complex solution could be introduced later.
-    prefer_seq: bool,
+    pick_by_seq: bool,
 }
 
 impl Default for SizeTieredPicker {
     fn default() -> Self {
         Self {
-            prefer_seq: prefer_pick_by_seq(),
+            pick_by_seq: prefer_pick_by_seq(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


analytic_engine/src/compaction/picker.rs
```diff
@@ -241,51 +262,87 @@ impl LevelPicker for SizeTieredPicker {
...
+                return files;
             }
+            debug!("No compaction necessary for segment, idx:{idx}, segment_key:{segment_key:?}");
         }
 
         None
     }
 }
 
 impl SizeTieredPicker {
```
Rachelint, line 280: It's not STCS now, I think maybe we should define a trait having the `pick_ssts` method and place it into TWCS?

analytic_engine/src/compaction/picker.rs
```diff
@@ -420,8 +477,17 @@ impl SizeTieredPicker {
 
 /// Time window compaction strategy
 /// See https://github.com/jeffjirsa/twcs/blob/master/src/main/java/com/jeffjirsa/cassandra/db/compaction/TimeWindowCompactionStrategy.java
-#[derive(Default)]
-pub struct TimeWindowPicker {}
+pub struct TimeWindowPicker {
+    pick_by_seq: bool,
+}
```
Rachelint, line 482: Define the trait, and refactor like this?
```
struct TimeWindowPicker {
    new_bucket_picker: Box<dyn Picker>,
    old_bucket_picker: Box<dyn Picker>,
}
```

analytic_engine/src/compaction/picker.rs
```diff
@@ -420,8 +477,17 @@ impl SizeTieredPicker {
 
 /// Time window compaction strategy
 /// See https://github.com/jeffjirsa/twcs/blob/master/src/main/java/com/jeffjirsa/cassandra/db/compaction/TimeWindowCompactionStrategy.java
```
Rachelint, line 479: Modify comment?

analytic_engine/src/compaction/picker.rs
```diff
@@ -176,10 +176,31 @@ fn trim_to_threshold(
...
+/// overlapping key range among them. Currently solution is relative simple,
+/// only pick adjacent sst. Maybe a better, but more complex solution could be
+/// introduced later.
+// TODO: Remove this function when pick_by_seq is stable.
+fn prefer_pick_by_seq() -> bool {
+    std::env::var("CERESDB_COMPACT_PICK_BY_SEQ").unwrap_or_else(|_| "true".to_string()) == "true"
+}
+
 /// Size tiered compaction strategy
 /// See https://github.com/jeffjirsa/twcs/blob/master/src/main/java/com/jeffjirsa/cassandra/db/compaction/SizeTieredCompactionStrategy.java
```
Rachelint, line 192: Modify comment?

analytic_engine/src/compaction/picker.rs
jiacai2050: Not necessary, old implementation will be removed when new is stable.

analytic_engine/src/compaction/picker.rs
jiacai2050: Although it's not strict size tiered, but I would like to keep the structure as it's now, current implementation is relative simple, we may need to group sst by size when there are no overlapping. See comment at L182.

analytic_engine/src/compaction/picker.rs
jiacai2050: Already comment at L182 to explain current behavior, we can update this when new strategy is stable.

analytic_engine/src/compaction/picker.rs
jiacai2050: Moved here.

Rachelint (approved):
LGTM

PR #1042, opened 2023-06-29
feat: sst-metadata support sort

jiacai2050:
## Rationale

When debugging SST, it's useful to check sst ordered by time/max_seq/size.

## Detailed Changes
- add a option `sort`


chunshao90 (approved):
LGTM

PR #1046, opened 2023-07-04
feat: memtable bitmap for null

ShiKaiWi:
## Rationale
For a table which contains many nullable columns, many nulls may cause huge memory consumption in the memtable, leading to small ssts where null values will be compressed.

## Detailed Changes
- Introduce a bitset for continuous row encoding
- Use u32 to encode the offest in the encoded bytes rather than u64

## Test Plan
Existing unit test. And test it with tsbs's data, here is result with write buffer 32MB:
|          description                  |   version  | encoded size of one row | number of rows in memtable | sst's size |
| --------------------------- | -------- | ------------------------- | ---------------------------- | --------- |
| 20 col (not null)                    | main      | 325B                                   | 70000+                                    | 1.8M        |
| 20 col (not null)                    | current | 295B                                   | 78000                                       | 2M           |
| 20 col + 200 null col            | main      | 2135B                                  | 12100                                      | 460K       |
| 20 col + 200 null col            | current | 332B                                   | 70000                                      | 1.5M         |

### DIFF
--- a/common_types/src/bitset.rs
+++ b/common_types/src/bitset.rs
@@ -63,31 +63,6 @@ impl BitSet {
         Some(set)
     }
 
-    /// Compute the number of the set bits before the bit at `index`.
-    pub fn num_unset_bits_before(&self, index: usize) -> Option<usize> {
-        self.num_set_bits_before(index).map(|num_set_bits| {
-            let num_bits = 1 + index;
-            num_bits - num_set_bits
-        })
-    }
-
-    /// Compute the number of the set bits before the bit at `index`.
-    pub fn num_set_bits_before(&self, index: usize) -> Option<usize> {
-        if index >= self.num_bits {
-            return None;
-        }
-        let (byte_index, bit_index) = Self::compute_byte_bit_index(index);
-
-        let mut num_set_bits = 0;
-        for b in &self.buffer[0..byte_index] {
-            num_set_bits += count_set_bits(*b) as u32;
-        }
-
-        num_set_bits += count_set_bits_before(self.buffer[byte_index], bit_index) as u32;
-
-        Some(num_set_bits as usize)
-    }
-
     #[inline]
     pub fn as_bytes(&self) -> &[u8] {
         &self.buffer
@@ -101,21 +76,6 @@ impl BitSet {
     fn compute_byte_bit_index(index: usize) -> (usize, usize) {
         (index >> 3, index & 7)
     }
-}
-
-fn count_set_bits(b: u8) -> u8 {
-    static LOOKUP: [u8; 16] = [0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4];
-
-    LOOKUP[(b & 0x0F) as usize] + LOOKUP[(b >> 4) as usize]
-}
-
-fn count_set_bits_before(b: u8, idx: usize) -> u8 {
-    let mut num_bits = 0;
-    for i in 0..idx {
-        num_bits += b >> i & 1;
-    }
-
-    num_bits
 }
 
 #[cfg(test)]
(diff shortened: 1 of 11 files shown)


common_types/src/bitset.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Tells whether the bit at the `index` is set.
+    pub fn is_set(&self, index: usize) -> Option<bool> {
+        if index >= self.num_bits {
+            return None;
+        }
+        let (byte_index, bit_index) = Self::compute_byte_bit_index(index);
+        let set = (self.buffer[byte_index] & (1 << bit_index)) != 0;
+        Some(set)
+    }
```
tanruixiang, line 64: Can another `unset` interface be provided?

common_types/src/row/contiguous.rs
```diff
@@ -50,47 +65,135 @@ pub trait ContiguousRow {
...
+        ensure!(inner.len() >= 4, NumNullColsMissing);
+        let num_bits = u32::from_ne_bytes(inner[0..4].try_into().unwrap()) as usize;
+        if num_bits > 0 {
+            ContiguousRowReaderWithNulls::try_new(inner, schema, num_bits).map(Self::WithNulls)
+        } else {
+            let reader = ContiguousRowReaderNoNulls {
+                inner,
+                byte_offsets,
+                datum_offset: 4,
+            };
```
tanruixiang, line 97: Can 4 be expressed as a constant?

common_types/src/bitset.rs
ShiKaiWi: I guess there is no need to do so because after initialization all the bits are unset.

common_types/src/row/contiguous.rs
```diff
@@ -280,45 +390,144 @@ impl<'a, T: RowBuffer + 'a> ContiguousRowWriter<'a, T> {
...
+        if num_null_cols > 0 {
+            self.write_row_with_nulls(row)
+        } else {
+            self.write_row_without_nulls(row)
+        }
+    }
+
+    fn write_row_with_nulls(&mut self, row: &Row) -> Result<()> {
+        let mut encoded_len = 0;
+        let mut num_bytes_of_string = 0;
```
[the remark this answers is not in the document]
ShiKaiWi: Done

common_types/src/row/contiguous.rs
```diff
@@ -50,47 +65,135 @@ pub trait ContiguousRow {
...
+            InvalidBitSetBytes {
+                expect_len: bit_set_size,
+                given_len: bit_set_buf.len()
+            }
+        );
+
+        let nulls_bit_set =
+            BitSet::try_from_raw(bit_set_buf[..bit_set_size].to_vec(), num_bits).unwrap();
+
+        let mut col_offsets = Vec::with_capacity(schema.num_columns());
```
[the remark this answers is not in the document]
ShiKaiWi: How about renaming it as `fixed_byte_offsets`?

common_types/src/row/contiguous.rs
ShiKaiWi: Updated.

common_types/src/bitset.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+//! BitSet supports counting set/unset bits.
+
+#[derive(Debug, Default, Clone)]
+pub struct BitSet {
+    /// The bits are stored as bytes in the least significant bit order.
+    buffer: Vec<u8>,
```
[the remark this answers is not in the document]
ShiKaiWi: Is there any advantage?

common_types/src/row/contiguous.rs
```diff
@@ -22,21 +23,35 @@ use crate::{
...
 
-/// Size to store the offset of string buffer.
-type OffsetSize = usize;
+/// Offset used in row's encoding
+type Offset = u32;
 
 /// Max allowed string length of datum to store in a contiguous row (16 MB).
 const MAX_STRING_LEN: usize = 1024 * 1024 * 16;
+/// Max allowed length of total bytes in a contiguous row (1 GB).
+const MAX_ROW_LEN: usize = 1024 * 1024 * 1024;
```
jiacai2050, line 54: This is aligned with WAL?

common_types/src/row/contiguous.rs
```diff
@@ -50,47 +65,197 @@ pub trait ContiguousRow {
...
+/// follows the type.
+///
+/// Here is the payload for variable length type:
+/// ```plaintext
+/// +-----------------+-------------------+
+/// |  var_len(u32)   |   var payload...  |
+/// +-----------------+-------------------+
+/// ```
+///
+/// # Row without null columns
```
jiacai2050, line 102: It seems this layout is same with above, so in fact we have only one layout, this is a special case when bit number is 0.

If my understanding is right, I think we can delete `Row without null columns` section, it feels there are two different layout in current writing style, maybe a note is enough, something like this:
> When there is no null value in row, then num_bits is 0, and bit_set is empty.

common_types/src/bitset.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// The number of real bits in the `buffer`
+    num_bits: usize,
+}
+
+impl BitSet {
+    /// Initialize a unset [`BitSet`].
+    pub fn new(num_bits: usize) -> Self {
```
jiacai2050, line 16: I think may be we can provide an another constructor with all bits set to 1, since in some case, there is hardly no null values at all.

common_types/src/row/contiguous.rs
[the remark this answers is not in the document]
jiacai2050: Sounds better 👍

common_types/src/row/contiguous.rs
ShiKaiWi: It has nothing with the WAL.

jiacai2050 (approved):
LGTM

PR #1049, opened 2023-07-05
feat: sql support dictionary column

### LINKED ISSUE #965: Use dictionary type to store tags column
### Describe This Problem

When we store data ordered by primary key =tsid, timestamp=, tags with same value will likely be put together, in current design we use string to represent it, in theory we could use more efficient datatype like dictionary.

### Proposal

Use dictionary to store tags column.

### Additional Context

https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Dictionary

tanruixiang:
## Rationale

Close #965 #1050

## Test Plan
CI


jiacai2050 (approved):
LGTM

PR #1056, opened 2023-07-07
feat: split operations of `Cluster` and `Shard`, and serialize operations of `Shard`

### LINKED ISSUE #1053: Relation between `Cluster` and `Shard` seems messy
### Describe This Problem

As I see, `Cluster` can  be seemed as `Space` in storage engine, and `Shard` can be seemed as `Table`.
So call relationship between `Cluster` and `Shard` should be similar as `Space` and `Table`'s, like:

However the operation levels seems messy in current design.

### Proposal

Refactor the call path between `Cluster` and `Shard` to be similar as `Space` and `Table`.

### Additional Context

_No response_

Rachelint:
## Rationale
Close #1053 

Now, write operations for shard will be performed parallelly in ceresdb node(e.g. recover, alter tables in shard).
Obviously, those operations should be performed sequentially like what about table.

## Detailed Changes
+ introduce `Shard` like `Table`.
+ make write operations for shard performing sequentially by adding a mutex.

## Test Plan
Test manually.


cluster/src/shard_set.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    fn insert(&mut self, shard_id: ShardId, shard: ShardRef) {
+        self.shards.insert(shard_id, shard);
+    }
+}
+
+#[derive(Debug)]
+pub struct Shard {
+    pub data: ShardData,
+    pub serializing_lock: tokio::sync::Mutex<()>,
```
ShiKaiWi, line 67: How about this name?
```suggestion
    pub serial_lock: tokio::sync::Mutex<()>,
```

cluster/src/shard_set.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// Insert the tables of one shard.
+    pub fn insert(&self, shard_id: ShardId, shard: ShardRef) {
+        self.inner.write().unwrap().insert(shard_id, shard)
+    }
+}
+
+#[derive(Debug, Default)]
+struct ShardSetInner {
```
ShiKaiWi, line 40: I guess an alias is enough for this hashmap?

cluster/src/shard_set.rs
```diff
@@ -0,0 +1,233 @@
...
+            data,
+            serializing_lock: tokio::sync::Mutex::new(()),
+        }
+    }
+}
+
+pub type ShardRef = Arc<Shard>;
+
+#[derive(Debug, Clone)]
+pub struct TableWithShards {
```
ShiKaiWi, line 91: It seems `TableWithShards` is not used in this crate. Shall we remove it or move it where it should be?

cluster/src/lib.rs
```diff
@@ -132,19 +129,21 @@ pub struct ClusterNodesResp {
...
+    async fn open_shard(&self, shard_info: &ShardInfo) -> Result<ShardRef>;
+
+    /// Get shard.
     ///
-    /// Return error if the shard is not found.
-    async fn close_shard(&self, req: ShardId) -> Result<TablesOfShard>;
-    /// Freeze the shard to reject create/drop table on the shard.
+    /// If target shard has opened in cluster, return it. Otherwise, return
+    /// None.
+    fn get_shard(&self, shard_id: ShardId) -> Option<ShardRef>;
```
ShiKaiWi, line 140:
```suggestion
    fn shard(&self, shard_id: ShardId) -> Option<ShardRef>;
```

cluster/src/shard_set.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    fn insert(&mut self, shard_id: ShardId, shard: ShardRef) {
+        self.shards.insert(shard_id, shard);
+    }
+}
+
+#[derive(Debug)]
+pub struct Shard {
+    pub data: ShardData,
```
ShiKaiWi, line 66: The nested sub-structs in `ShardData` are too verbose. How about remove current `ShardData` and renaming `ShardDataInner` as `ShardData`, and wrapping the `ShardData` in lock directly here:
```
pub data: std::sync::RwLock<ShardData>
```

server/src/grpc/meta_event_service/mod.rs
```diff
@@ -259,10 +295,14 @@ async fn do_open_shard(ctx: HandlerContext, shard_info: ShardInfo) -> Result<()>
             msg: "fail to open shards in cluster",
         })?;
 
+    // Lock the shard in local, and then recover it.
+    let _guard = shard.serializing_lock.lock().await;
```
ShiKaiWi, line 299: I guess we can define a struct called `ShardSerialExec` which holding the `serial_lock` guard in its inner. And all the update operations on the shard are defined on the `ShardSerialExec`. After this, the user won't be able to operate the shard without holding the shard's `serial_lock` anymore.

server/src/grpc/meta_event_service/mod.rs
```diff
@@ -317,21 +360,25 @@ async fn handle_open_shard(ctx: HandlerContext, request: OpenShardRequest) -> Re
...
+        .with_context(|| ErrNoCause {
+            code: StatusCode::Internal,
+            msg: format!("shard not found when closing shard, shard_id:{shard_id}",),
+        })?;
+
+    // Lock for serializing the write operation.
+    let _guard = shard.serializing_lock.lock().await;
+    info!("Shard close sequentially begin, shard_id:{shard_id}");
+
+    shard.data.freeze();
```
ShiKaiWi, line 375: If the `serial_lock` is introduced, I guess the freeze is not necessary anymore.

server/src/grpc/meta_event_service/mod.rs
Rachelint: Has refactor all related logic.

cluster/src/shard_operator.rs
```diff
@@ -0,0 +1,404 @@
...
+            })?;
+
+        info!("ShardOperator close sequentially finish, shard_info:{shard_info:?}");
+
+        Ok(())
+    }
+
+    pub async fn create_table(&self, ctx: CreateTableContext) -> Result<()> {
+        let shard_info = ctx.updated_table_info.shard_info.clone();
+        let table_info = ctx.updated_table_info.table_info.clone();
```
ShiKaiWi, line 213: Are these two clones necessary?

cluster/src/shard_set.rs
```diff
@@ -0,0 +1,227 @@
...
+        inner.insert(shard_id, shard);
+    }
+}
+
+/// Shard
+///
+/// NOTICE: all write operations on a shard will be performed sequentially.
+pub struct Shard {
+    pub data: ShardDataRef,
+    pub operator: tokio::sync::Mutex<ShardOperator>,
```
ShiKaiWi, line 55: This field may be private.

cluster/src/shard_operator.rs
Rachelint: Maybe necessary now? Becasue we need to use these two inner struct, after moving `ctx.updated_table_info`.

ShiKaiWi (approved):
LGTM

PR #1061, opened 2023-07-11
fix: fix fully flushed region open in kakfa wal

Rachelint:
## Rationale
Now when opening a fully flushed region, it may hang forever...
This pr fix this problem on wal level, but it is the problem in kafka client that is planned to fix in later.

## Detailed Changes
See title.

## Test Plan
Test manually.


ShiKaiWi (approved):
LGTM

PR #1062, opened 2023-07-11
fix: directly return when found no table datas to replay

Rachelint:
## Rationale
Now it will still scan logs from wal when no table datas needed to replay, it should directly return actually.

## Detailed Changes
Add a fast path to return directly when no table datas needed to replay.

## Test Plan
Test manually.


ShiKaiWi (approved):
LGTM

PR #1065, opened 2023-07-11
fix: avoid write queue full block

baojinri:
## Rationale
If the first write request is canceled, the write request in the queue will never be executed, until the queue is full, then an error will be returned.

## Detailed Changes
Use `CancellationSafeFuture` to  execute the write requests, see #1071 

## Test Plan
Existing tests.

### DIFF
--- a/analytic_engine/src/table/mod.rs
+++ b/analytic_engine/src/table/mod.rs
@@ -292,7 +292,7 @@ impl TableImpl {
 
                 match CancellationSafeFuture::new(
                     Self::write_requests(write_requests),
-                    self.instance.read_runtime().clone(),
+                    self.instance.write_runtime().clone(),
                 )
                 .await
                 {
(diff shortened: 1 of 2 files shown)


analytic_engine/src/table/mod.rs
```diff
@@ -250,35 +277,38 @@ impl TableImpl {
...
+                    self.table_data.clone(),
+                    self.pending_writes.clone(),
                 );
-                let merged_write_request =
-                    merge_pending_write_requests(pending_writes.writes, pending_writes.num_rows);
-                (merged_write_request, serial_exec, pending_writes.notifiers)
+
+                match CancellationSafeFuture::new(
+                    Self::write_requests(write_requests),
+                    self.instance.read_runtime().clone(),
```
jiacai2050, line 295: It should use write runtime

PR #1068, opened 2023-07-12
fix: add missing `dict_id`

jiacai2050:
## Rationale
When schema has more than one dict fields, datafusion may lose the `dict_id`, which will cause an error in the record batch encode, and the client decoded the record batch via ipc will throw following errors:
```
DecodeArrowPayload(InvalidArgumentError("Value at position 0 out of bounds: 0 (should be in [0, -1])"))
```
More context: https://github.com/apache/arrow-datafusion/issues/6784
## Detailed Changes
- Assign unique dict id when there are more than one dict fields. 

## Test Plan
UT and integration tests.

### DIFF
--- a/components/arrow_ext/src/ipc.rs
+++ b/components/arrow_ext/src/ipc.rs
@@ -49,8 +49,8 @@ const ZSTD_LEVEL: i32 = 3;
 pub struct RecordBatchesEncoder {
     stream_writer: Option<StreamWriter<Vec<u8>>>,
     num_rows: usize,
-    /// Whether the writer has dict fields.
-    has_dict: bool,
+    /// Whether the writer has more than one dict fields.
+    more_than_one_dict: bool,
     compress_opts: CompressOptions,
 }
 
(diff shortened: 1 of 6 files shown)


integration_tests/cases/env/local/ddl/alter_table.sql
```diff
@@ -13,15 +13,15 @@ INSERT INTO TABLE `05_alter_table_t0`(a, b, t, dic) values (2, '2', 2, "d2");
 SELECT * FROM `05_alter_table_t0`;
 
 -- waiting for datafusion's bug fix
```
tanruixiang, line 15:
```suggestion
```

components/arrow_ext/src/ipc.rs
```diff
@@ -116,20 +120,82 @@ impl RecordBatchesEncoder {
...
+            let stream_writer = StreamWriter::try_new(buffer, &schema).context(ArrowError)?;
+            self.more_than_one_dict = schema.is_owned();
             self.stream_writer = Some(stream_writer);
             self.stream_writer.as_mut().unwrap()
         };
 
-        stream_writer.write(batch).context(ArrowError)?;
+        if self.more_than_one_dict {
+            let schema = batch.schema();
+            let schema = Self::convert_schema(&schema);
```
Rachelint, line 192: Maybe we can cache the converted schema to avoid duplicated converts?

Rachelint (approved):
LGTM

PR #1071, opened 2023-07-12
feat: support cancellation safe future

ShiKaiWi:
## Rationale
In some case, the future cancel safety must be guaranteed, but there is no such a mechanism to make it now.

## Detailed Changes
Introduce `CancellationSafeFuture` to wrap a future to ensure it must be executed.

## Test Plan
New unit tests.

## Additional
This implementation refers to: https://github.com/influxdata/influxdb_iox/blob/885767aa0a6010de592bde9992945b01389eb994/cache_system/src/cancellation_safe_future.rs

### DIFF
--- a/common_util/src/future_cancel.rs
+++ b/common_util/src/future_cancel.rs
@@ -1,35 +1,97 @@
 // Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
 
-/// A guard to detect whether a future is cancelled.
-pub struct FutureCancelGuard<F: FnMut()> {
-    cancelled: bool,
-    on_cancel: F,
+//! A future wrapper to ensure the wrapped future must be polled.
+//!
+//! This implementation is forked from: https://github.com/influxdata/influxdb_iox/blob/885767aa0a6010de592bde9992945b01389eb994/cache_system/src/cancellation_safe_future.rs
+//! Here is the copyright and license disclaimer:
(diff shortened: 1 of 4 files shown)


common_util/src/future_cancel.rs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                let mut state = cloned_state.lock().unwrap();
+                *state = State::Processing;
+            }
+            let mut cancel_guard = FutureCancelGuard::new_cancelled(|| {
+                let mut state = cloned_state.lock().unwrap();
+                *state = State::Cancelled;
+            });
+
+            // It may be cancelled at this await point.
```
tanruixiang, line 74:
```suggestion
            // It will be cancelled at this await point.
```

common_util/src/future_cancel.rs
```diff
@@ -0,0 +1,91 @@
...
+            cancelled: true,
+            on_cancel,
+        }
+    }
+
+    /// Set the inner state is uncancelled to ensure the `on_cancel` callback
+    /// won't be executed.
+    pub fn uncancel(&mut self) {
+        self.cancelled = false;
+    }
```
tanruixiang, line 26: Personally, I think it feels more appropriate to call it finish, and it looks like the function will only be used when the feature is complete.

ShiKaiWi: Sounds reasonable.

Rachelint (approved):
LGTM

PR #1073, opened 2023-07-13
refactor: remove type encode

baojinri:
## Rationale
In order to reduce the memory consumption of encoding row.

## Detailed Changes
- Remove type encode.
- Use schema to obtain type.

## Test Plan
Existing tests.


ShiKaiWi (approved):
LGTM

Issue #1075, opened 2023-07-14
Failed to create the table because of creation of the snapshot


chunshao90:
### Describe this problem

Failed to create the table because of creation of the snapshot.
```
{
    "code": 400,
[...]

PR #1076, opened 2023-07-14
feat: use manifest updates stats in `TableData` to trigger snapshot

Rachelint:
## Rationale
Now we stats manifest updates of all tables in `Manifest` to trigger snapshot. 
However when snapshot is triggered, it will just do snapshot of latest table storing update.
It is obvious unreasonable...

Worse, this mechanism lead to bug #1075 ...

So In this pr, I place the manifest updates stats into `TableData` , and each table can only trigger snapshot of itself rather than others.

## Detailed Changes
+ Place the manifest updates stats into `TableData`.
+ Use it to trigger manifest snapshot.

## Test Plan
Test by new ut.

### DIFF
--- a/analytic_engine/src/instance/open.rs
+++ b/analytic_engine/src/instance/open.rs
@@ -4,7 +4,6 @@
 
 use std::{
     collections::HashMap,
-    num::NonZeroUsize,
     sync::{Arc, RwLock},
 };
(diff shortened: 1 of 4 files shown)


jiacai2050 (approved):
LGTM

PR #1080, opened 2023-07-17
fix: ensure shard is opened once

jiacai2050:
## Rationale
In current implementation, open shard process have following issue:
1. When tables of shard is not fully opened, it will report to meta the shard is opened. It should wait all tables are opened.
2. Open shard may get executed multiple times, since meta may keep sending open shard request

## Detailed Changes
1. Introduce ShardStatus, which will be set to `Opened` when all tables are opened, and only after this, shard can be sent to meta.
2. Before open shard, check status, if it's already in opening, return an Error.

## Test Plan
WIP

### DIFF
--- a/cluster/src/shard_set.rs
+++ b/cluster/src/shard_set.rs
@@ -10,7 +10,8 @@ use crate::{
         CloseContext, CloseTableContext, CreateTableContext, DropTableContext, OpenContext,
         OpenTableContext, ShardOperator,
     },
-    Result, ShardVersionMismatch, TableAlreadyExists, TableNotFound, UpdateFrozenShard,
+    OpenShardNoCause, Result, ShardVersionMismatch, TableAlreadyExists, TableNotFound,
+    UpdateFrozenShard,
 };
 
 /// Shard set
@@ -66,7 +67,7 @@ impl Shard {
         let data = Arc::new(std::sync::RwLock::new(ShardData {
             shard_info: tables_of_shard.shard_info,
             tables: tables_of_shard.tables,
-            frozen: false,
+            status: ShardStatus::default(),
         }));
 
         let operator = tokio::sync::Mutex::new(ShardOperator { data: data.clone() });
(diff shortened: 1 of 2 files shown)


cluster/src/shard_set.rs
```diff
@@ -86,7 +87,35 @@ impl Shard {
...
+        }
+
+        let ret = operator.open(ctx).await;
+
+        {
+            let mut data = self.data.write().unwrap();
+            if ret.is_ok() {
+                data.finish_open();
+            } else {
+                data.status = ShardStatus::NewlyCreated;
```
Rachelint, line 109: How about just keep the `Opening` state when failed?

cluster/src/shard_set.rs
```diff
@@ -124,6 +153,19 @@ pub struct UpdatedTableInfo {
...
 
+#[derive(Debug, Default, PartialEq)]
+pub enum ShardStatus {
+    /// Not allowed report to ceresmeta
+    #[default]
+    NewlyCreated,
+    /// Not allowed report to ceresmeta
+    Opening,
+    /// Healthy
+    Opened,
```
Rachelint, line 164: How about Ready?

cluster/src/shard_set.rs
```diff
@@ -124,6 +153,19 @@ pub struct UpdatedTableInfo {
     pub table_info: TableInfo,
 }
 
+#[derive(Debug, Default, PartialEq)]
+pub enum ShardStatus {
+    /// Not allowed report to ceresmeta
+    #[default]
+    NewlyCreated,
```
Rachelint, line 160: How about Init?

cluster/src/shard_set.rs
```diff
@@ -157,7 +225,7 @@ impl ShardData {
         } = updated_info;
 
         ensure!(
-            !self.frozen,
+            !matches!(self.status, ShardStatus::Frozen),
```
Rachelint, line 228: How about define `is_frozen` like `is_opened` ?

cluster/src/shard_set.rs
```diff
@@ -194,7 +290,7 @@ impl ShardData {
...
```
Rachelint, line 293: how about using `is_frozen` here?

cluster/src/shard_set.rs
```diff
@@ -124,6 +161,36 @@ pub struct UpdatedTableInfo {
...
+///   └──┬─┘
+///   ___▽___
+///  ╱       ╲     ┌─────┐
+/// ╱ Opening ╲____│Ready│
+/// ╲         ╱yes └──┬──┘
+///  ╲_______╱    ┌───▽──┐
+///               │Frozen│
+///               └──────┘
+/// ```
+/// When a open request comes in, shard can only be opened when it's in
```
Rachelint, line 178: an open request

Rachelint (approved):
LGTM

cluster/src/shard_set.rs
jiacai2050: Will fix in next PR

PR #1082, opened 2023-07-18
feat: add shard status when heartbeat to meta

jiacai2050:
## Rationale
After #1080, ceresdb will not heartbeat to meta until all table open finished, that is to say that all tables inside a shard will not be used for write/read, which is unreasonable.

## Detailed Changes
- Introduce PartialOpen status sent to meta, and it will send this status when part of tables are opened, and meta will route table in this status.
- <del>Flush table when open successfully to reduce WAL size.</del>
## Test Plan

Manually

### DIFF
--- a/catalog/src/table_operator.rs
+++ b/catalog/src/table_operator.rs
@@ -5,10 +5,7 @@ use std::time::Instant;
 use generic_error::BoxError;
 use log::{error, info, warn};
 use snafu::{OptionExt, ResultExt};
-use table_engine::{
-    engine,
-    table::{FlushRequest, TableRef},
-};
+use table_engine::{engine, table::TableRef};
 use time_ext::InstantExt;
 
 use crate::{
@@ -77,12 +74,6 @@ impl TableOperator {
 
             match table_result {
                 Ok(Some(table)) => {
-                    // When table open successfully, try flush to reduce WAL size, so when reopen
-                    // this shard the WAL required to fetch is reduced.
-                    if let Err(e) = table.flush(FlushRequest { sync: false }).await {
-                        warn!("Try flush after open table failed, err:{e}");
-                    }
-
                     schema.register_table(table);
                     success_count += 1;
                 }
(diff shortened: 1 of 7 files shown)


meta_client/src/types.rs
```diff
@@ -162,18 +162,58 @@ pub struct NodeInfo {
...
 impl ShardInfo {
     #[inline]
     pub fn is_leader(&self) -> bool {
         self.role == ShardRole::Leader
     }
+
+    #[inline]
+    pub fn is_opened(&self) -> bool {
+        matches!(self.status, ShardStatus::Ready | ShardStatus::Frozen)
+    }
```
Rachelint, line 216: I think the logic is a bit strange.
+ `ShardInfo` is defined in meta_client.
+ However, it can only be modify in `cluster`.
+ And it is read in `meta_client` from `cluster`.

How about define a struct such as `ShardState` in `cluster` and use it?
And we convert it to `ShardInfo` when call `all_shards`, during the convert, we map `ShardStatus` inner `ceresdb` to which sent to `ceresmeta`?
And the strange calling of `is_opened` in pb encode can be removed.

meta_client/src/types.rs
```diff
@@ -162,18 +162,58 @@ pub struct NodeInfo {
...
+    /// In opening
+    Opening,
+    /// Healthy
+    Ready,
+    /// Further updates are prohibited
+    Frozen,
+}
+
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize)]
 pub struct ShardInfo {
```
Rachelint, line 196: I guess it may better to move it back to `cluster`, and we define a struct just for sending in `meta_client`.

catalog/src/table_operator.rs
```diff
@@ -74,6 +77,12 @@ impl TableOperator {
 
             match table_result {
                 Ok(Some(table)) => {
+                    // When table open successfully, try flush to reduce WAL size, so when reopen
+                    // this shard the WAL required to fetch is reduced.
+                    if let Err(e) = table.flush(FlushRequest { sync: false }).await {
+                        warn!("Try flush after open table failed, err:{e}");
+                    }
```
Rachelint, line 84: If we report shard has opened before flush finished, maybe it will have no effect about reduce wal size in retry.
It is possible that, the retry request will be executed before none of the flush works is finished.

Rachelint:
LGTM

PR #1083, opened 2023-07-19
fix: quote column name in prom query filter

jiacai2050:
## Rationale
PromQL query should be case-sensitive for  column name

## Detailed Changes
- Use `ident` when construct filter.

## Test Plan
Update integration tests and UT


ShiKaiWi (approved):
LGTM

PR #1085, opened 2023-07-19
feat: cache record batch

ShiKaiWi:
## Rationale
Currently, there is no cache layer for the decoded record batch, and it will result in that the extra memory consumption for the query involving the similar record batches. With the record batch cache, the decoded record batches will be stored, and multiple queries can share the same record batches, which brings us less memory usage.

## Detailed Changes
Support cache layer for row groups in the record batch.

## Test Plan
Existing and newly-added unit tests.


ShiKaiWi:
Currently, the index can't filter out most of the row groups, which leads to high eviction rate of the cache.

PR #1089, opened 2023-07-20
fix:  out of range error in module ObjectStore base on OBKV

MachaelLee:
## Rationale
In current version, there are two bugs in module `ObjectStore` based on OBKV.
* `Out of range` error occurs when calling  method `get_range` 
*  Save meta info error when putting an existing file.

## Detailed Changes
* Fix these bugs
* Add some unit tests

## Test Plan
* Test by unit tests

### DIFF
--- a/components/object_store/src/obkv/meta.rs
+++ b/components/object_store/src/obkv/meta.rs
@@ -178,8 +178,8 @@ impl ObkvObjectMeta {
         let start_index = range.start / batch_size;
         let start_offset = range.start % batch_size;
 
-        // if `range.end` is zero, it means the range is empty
-        let inclusive_end = if range.end == 0 {
+        // if the range is empty, return empty parts
+        let inclusive_end = if range.is_empty() {
             return Ok(ConveredParts {
                 part_keys: &self.parts[0..0],
                 start_offset: 0,
--- a/components/object_store/src/obkv/meta.rs
+++ b/components/object_store/src/obkv/meta.rs
@@ -387,7 +387,7 @@ mod test {
 
         let range1 = Range { start: 0, end: 0 };
         let expect = meta.compute_covered_parts(range1).unwrap();
-        assert!(expect.part_keys.len() == 0);
+        assert!(expect.part_keys.is_empty());
         assert!(expect.start_offset == 0);
         assert!(expect.end_offset == 0);
     }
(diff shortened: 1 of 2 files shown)


components/object_store/src/obkv/meta.rs
```diff
@@ -375,4 +399,16 @@ mod test {
             version: String::from("123456fsdalfkassa;l;kjfaklasadffsd"),
         }
     }
+
+    fn build_test_meta1() -> ObkvObjectMeta {
```
ShiKaiWi, line 403: If name it in this style, how about renaming `build_test_meta` to `build_test_meta0` at L382?

components/object_store/src/obkv/meta.rs
```diff
@@ -177,8 +177,13 @@ impl ObkvObjectMeta {
         let batch_size = self.part_size;
         let start_index = range.start / batch_size;
         let start_offset = range.start % batch_size;
-        let end_index = range.end / batch_size;
-        let end_offset = range.end % batch_size;
+        let mut end_index = range.end / batch_size;
+        let mut end_offset = range.end % batch_size;
+
+        if end_offset == 0 {
+            end_index -= 1;
```
ShiKaiWi, line 184: If `range.end == 0`, then this may overflow? And I guess this bug results from the `range.end` is exclusive, so how about putting it as:
```rust
let inclusive_end =. if range.end == 0 {
   return Ok(...)
} else {
   range.end - 1
};

let end_index = inclusive_end / batch_size;
let end_offset = inclusive_end / batch_size;
```

components/object_store/src/obkv/meta.rs
MachaelLee: ok

components/object_store/src/obkv/meta.rs
MachaelLee: Oops, yes, it may overflow here. And I will fix it.

components/object_store/src/obkv/meta.rs
```diff
@@ -177,8 +177,20 @@ impl ObkvObjectMeta {
         let batch_size = self.part_size;
         let start_index = range.start / batch_size;
         let start_offset = range.start % batch_size;
-        let end_index = range.end / batch_size;
-        let end_offset = range.end % batch_size;
+
+        // if `range.end` is zero, it means the range is empty
+        let inclusive_end = if range.end == 0 {
```
jiacai2050, line 182: `range.is_empty`?

MachaelLee: good idea.

PR #1098, opened 2023-07-22
feat: refactor stats method in wal

Rachelint:
## Rationale
Although `get_statistic` method exist in `WalManager`, it print so less message and just impl in RocksDB wal.
In this pr, I impl this method in all wal impls, and print the detail inner infos through it.

The output example(RocksDB wal):
```
[Data wal stats]:
#RocksDB stats:
#Wal stats:
TableUnit { id: 2199023255724, next_sequence_num: 59 }
TableUnit { id: 2199023255777, next_sequence_num: 50 }
TableUnit { id: 2199023255627, next_sequence_num: 50 }
...

------------------------------------------------------

[Manifest wal stats]:
#RocksDB stats:
#Wal stats:
TableUnit { id: 2199023255739, next_sequence_num: 2 }
TableUnit { id: 2199023255669, next_sequence_num: 2 }
TableUnit { id: 2199023255740, next_sequence_num: 2 }
...

```

## Detailed Changes
Impl and improve the `get_statistic` method in `WalManager`.

## Test Plan
Test manually.

### DIFF
--- a/wal/src/manager.rs
+++ b/wal/src/manager.rs
@@ -318,7 +318,7 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
     async fn scan(&self, ctx: &ScanContext, req: &ScanRequest) -> Result<BatchLogIteratorAdapter>;
 
     /// Get statistics
-    fn get_statistics(&self) -> Option<String> {
+    async fn get_statistics(&self) -> Option<String> {
         None
     }
 }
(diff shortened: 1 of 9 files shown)


wal/src/manager.rs
```diff
@@ -318,7 +318,7 @@ pub trait WalManager: Send + Sync + fmt::Debug + 'static {
     async fn scan(&self, ctx: &ScanContext, req: &ScanRequest) -> Result<BatchLogIteratorAdapter>;
 
     /// Get statistics
-    fn get_statistics(&self) -> Option<String> {
+    async fn get_statistics(&self) -> Option<String> {
         None
     }
```
tanruixiang, line 323:
```suggestion
    async fn get_statistics(&self) -> Option<String>
```
Is it possible to make the method have no default implementation?

ShiKaiWi (approved):
LGTM

wal/src/manager.rs
Rachelint: It seems reasonable.

PR #1100, opened 2023-07-24
feat: query requests dedup

baojinri:
## Rationale
In order to reduce the execution of the same query requests，then reduce memory consumption when querying.

## Detailed Changes
- Use hashmap to dedup query requests.
- Only the first query request will be execute, and other same query requests will wait for the execution result of the first request.

## Test Plan
Exsiting tests


server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,39 +62,105 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+        );
+
+        {
+            let mut dedup_map = self.dedup_map.write().await;
+            match dedup_map.get_notifiers(request_key.clone()) {
+                Some(notifiers) => {
+                    let txs = notifiers.get_txs();
+                    let mut notifiers = Notifiers::new(txs.clone());
+                    notifiers.add_tx(tx);
+                    dedup_map.add_notifiers(request_key.clone(), notifiers);
```
tanruixiang, line 89:
```suggestion
                    notifiers.add_tx(tx);
```
It looks like the code is duplicated here.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,39 +62,103 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+        })?;
+
+        let request_key = RequestKey::new(
+            table.table,
+            read_request.predicate.clone(),
+            read_request.projected_schema.projection(),
+            read_request.order,
+        );
+
+        let mut dedup_map = self.dedup_map.write().await;
```
Rachelint, line 82: Should not use async lock.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,40 +65,130 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+            let dedup_map = self.dedup_map.read().unwrap();
+            if dedup_map.contains_key(&request_key) {
+                let notifiers = dedup_map.get_notifiers(&request_key).unwrap();
+                notifiers.add_notifier(tx);
+
+                REMOTE_ENGINE_GRPC_HANDLER_DURATION_HISTOGRAM_VEC
+                    .stream_read
+                    .observe(instant.saturating_elapsed().as_secs_f64());
+                return Ok(ReceiverStream::new(rx));
+            } else {
```
ShiKaiWi, line 95: Maybe we can remove this else block to make it simple.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,40 +65,130 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+                }
+                Err(_) => {
+                    for tx in &notifiers {
+                        if let Err(e) = tx
+                            .send(
+                                ErrNoCause {
+                                    code: StatusCode::Internal,
+                                    msg: "failed to handler request".to_string(),
+                                }
+                                .fail(),
```
ShiKaiWi, line 177: Declare the error on another line to make code clean.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,40 +65,130 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+        let request = request.into_inner();
+        let table_engine::remote::model::ReadRequest {
+            table,
+            read_request,
+        } = request.clone().try_into().box_err().context(ErrWithCause {
+            code: StatusCode::BadRequest,
+            msg: "fail to convert read request",
+        })?;
+
+        let request_key = RequestKey::new(
```
ShiKaiWi, line 78: I'd like to make this feature optional because it introduces some overhead, that is to say, if this feature is disabled, the procedure should be kept as before.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,40 +65,130 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+                                ErrNoCause {
+                                    code: StatusCode::Internal,
+                                    msg: "failed to handler request".to_string(),
+                                }
+                                .fail(),
+                            )
+                            .await
+                        {
+                            error!("Failed to send handler result, err:{}.", e);
+                        }
```
ShiKaiWi, line 182: If the error happens, shall we break the forloop?

server/src/grpc/remote_engine_service/dedup_request.rs
```diff
@@ -0,0 +1,80 @@
...
+impl DedupMap {
+    pub fn get_notifiers(&self, key: &RequestKey) -> Option<&Notifiers> {
+        self.inner.get(key)
+    }
+
+    pub fn add_notifiers(&mut self, key: RequestKey, value: Notifiers) {
+        self.inner.insert(key, value);
+    }
+
+    pub fn delete_notifiers(&mut self, key: RequestKey) -> Option<Notifiers> {
```
ShiKaiWi, line 73:
```suggestion
    pub fn take_notifiers(&mut self, key: RequestKey) -> Option<Vec<Notifier>> {
```

server/src/grpc/remote_engine_service/dedup_request.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.notifiers.into_inner().unwrap()
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct DedupMap {
+    inner: HashMap<RequestKey, Notifiers>,
+}
+
+impl DedupMap {
```
ShiKaiWi, line 64: `DedupMap` is not a good name. How about naming it as `RequestNotifiers`?

server/src/grpc/remote_engine_service/dedup_request.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    pub fn add_notifiers(&mut self, key: RequestKey, value: Notifiers) {
+        self.inner.insert(key, value);
+    }
+
+    pub fn delete_notifiers(&mut self, key: RequestKey) -> Option<Notifiers> {
+        self.inner.remove(&key)
+    }
+
+    pub fn contains_key(&self, key: &RequestKey) -> bool {
```
ShiKaiWi, line 77:
```suggestion
    pub fn contains(&self, key: &RequestKey) -> bool {
```

server/src/grpc/remote_engine_service/dedup_request.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct DedupMap {
+    inner: HashMap<RequestKey, Notifiers>,
+}
+
+impl DedupMap {
+    pub fn get_notifiers(&self, key: &RequestKey) -> Option<&Notifiers> {
```
ShiKaiWi, line 65: Maybe we should remove `get_notifiers` and `add_notifiers`, and only provide a method called `insert` to do only a notifier insertion?

server/src/grpc/remote_engine_service/dedup_request.rs
```diff
@@ -0,0 +1,80 @@
...
+            projection,
+            order,
+        }
+    }
+}
+
+type Notifier = Sender<Result<RecordBatch>>;
+
+#[derive(Debug, Default)]
+pub struct Notifiers {
```
ShiKaiWi, line 38: I guess this struct is not necessary to be exposed.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -57,40 +62,110 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+        })?;
+
+        let request_key = RequestKey::new(
+            table.table,
+            read_request.predicate.clone(),
+            read_request.projected_schema.projection(),
+            read_request.order,
+        );
+
+        if let Some(request_notifiers) = self.request_notifiers.clone() {
```
ShiKaiWi, line 82: I guess we can provide two implementations: `stream_read_internal` & `deduped_stream_read_internal` and  which should be used depends on whether the `request_notifiers` is set.

server/src/grpc/remote_engine_service/dedup_requests.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2022-2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+use std::{collections::HashMap, hash::Hash, sync::RwLock};
+
+use tokio::sync::mpsc::Sender;
+
+type Notifier<T, E> = Sender<Result<T, E>>;
```
ShiKaiWi, line 7: Maybe there is no need to require the notification item to be a result.

server/src/config.rs
```diff
@@ -121,6 +121,9 @@ pub struct ServerConfig {
 
     /// Config of remote engine client
     pub remote_client: remote_engine_client::Config,
+
+    /// Whether to deduplicate requests
+    pub dedup_requests: bool,
```
ShiKaiWi, line 126:
```suggestion
    pub enable_query_dedup: bool,
```

server/src/grpc/remote_engine_service/dedup_requests.rs
ShiKaiWi, line 1: I guess we should put this module under the server module because it can not be used for only remote_engine_service.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -16,37 +16,95 @@ use ceresdbproto::{
...
+        Self {
+            table,
+            predicate,
+            projection,
+            order,
+        }
+    }
+}
+
+struct RequestNotifierGuard<F: FnOnce()>(Option<F>);
```
ShiKaiWi, line 77: Maybe it'll be better if we put it like:
```suggestion
struct ExecutionGuard<F> {
   do: F,
   cancelled: bool,
}

impl<F> for ExecutionGuard<F> {
   fn cancel(&mut self) {
       self.cancelled = true;
   }
}

impl<F: FnOnce()> Drop for ExecutionGuard<F> {...}
```

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -97,6 +155,140 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+        })?;
+
+        let request_key = RequestKey::new(
+            table.table,
+            read_request.predicate.clone(),
+            read_request.projected_schema.projection(),
+            read_request.order,
+        );
+
+        let mut _guard = RequestNotifierGuard::default();
```
ShiKaiWi, line 182: The guard is used so there is no need to prefix it with `_`.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -97,6 +155,140 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+        match self
+            .request_notifiers
+            .as_ref()
+            .unwrap()
+            .insert_notifier(request_key.clone(), tx)
+        {
+            // The first request, need to handle it, and then notify the other requests.
+            RequestResult::First => {
+                // The _guard is used to remove key when future is cancelled.
+                _guard.set_some(|| {
```
ShiKaiWi, line 192: There is no need to use the pattern declare and update for the guard. A declaration is enough:
```rust
let guard = match ... {
    First => ExecutionGuard::new(..),
    .... => return ...,
}
```

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -97,6 +155,139 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+            table.table,
+            read_request.predicate.clone(),
+            read_request.projected_schema.projection(),
+            read_request.order,
+        );
+
+        let mut guard = match self
+            .request_notifiers
+            .as_ref()
+            .unwrap()
```
ShiKaiWi, line 185: Maybe we can declare the `request_notifiers` as the params of `deduped_stream_read_internal` to avoid call `as_ref().unwrap()` again and again.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -97,6 +155,139 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+                        .fail();
+                        if let Err(e) = notifier.send(err).await {
+                            error!("Failed to send handler result, err:{}.", e);
+                        }
+                    }
+                    break;
+                }
+            }
+        }
+
```
ShiKaiWi, line 280: After the succeeds in notifying, how about adding a metric to record the number of `dedupped_stream_query`?

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -16,37 +16,95 @@ use ceresdbproto::{
...
 };
 
-mod error;
+pub mod error;
 
 const STREAM_QUERY_CHANNEL_LEN: usize = 20;
 const DEFAULT_COMPRESS_MIN_LENGTH: usize = 80 * 1024;
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct RequestKey {
```
ShiKaiWi, line 54: How about renaming it as `StreamReadReqKey`?

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -97,6 +155,131 @@ impl<Q: QueryExecutor + 'static> RemoteEngineServiceImpl<Q> {
...
+            // The first request, need to handle it, and then notify the other requests.
+            RequestResult::First => {
+                // This is used to remove key when future is cancelled.
+                ExecutionGuard::new(|| {
+                    request_notifiers.take_notifiers(&request_key);
+                })
+            }
+            // The request is waiting for the result of first request.
+            RequestResult::Wait => {
+                REMOTE_ENGINE_GRPC_HANDLER_DURATION_HISTOGRAM_VEC
```
ShiKaiWi, line 193: This observation is wrong because the query is not finished at this point. A better way to collect the metrics when `rx` is dropped.

server/src/dedup_requests.rs
```diff
@@ -0,0 +1,89 @@
...
+where
+    K: PartialEq + Eq + Hash,
+{
+    /// Insert a notifier for the given key.
+    pub fn insert_notifier(&self, key: K, notifier: Notifier<T>) -> RequestResult {
+        // First try to read the notifiers, if the key exists, add the notifier to the
+        // notifiers.
+        let notifiers = self.inner.read().unwrap();
+        if notifiers.contains_key(&key) {
+            notifiers.get(&key).unwrap().add_notifier(notifier);
```
ShiKaiWi, line 56: The check on L55 may be wrong when L56 is reached. Maybe we should a method called insert_if_key_exists for notifiers.

server/src/config.rs
```diff
@@ -140,6 +143,7 @@ impl Default for ServerConfig {
             route_cache: router::RouteCacheConfig::default(),
             hotspot: hotspot::Config::default(),
             remote_client: remote_engine_client::Config::default(),
+            enable_query_dedup: true,
```
ShiKaiWi, line 146: Make it false by default.

server/src/dedup_requests.rs
ShiKaiWi, line 56:
```rust
if let Some(notifiers) = notifiers.get(&key) {
   notifiers.add_notifier(notifier);
}
```

server/src/dedup_requests.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.notifiers.write().unwrap().push(notifier);
+    }
+}
+
+#[derive(Debug)]
+pub struct RequestNotifiers<K, T>
+where
+    K: PartialEq + Eq + Hash,
+{
+    inner: RwLock<HashMap<K, Notifiers<T>>>,
```
ShiKaiWi, line 32: `notifiers_by_key`.

ShiKaiWi (approved):
LGTM

PR #1101, opened 2023-07-24
feat: support compact table concurrently

jiacai2050:
## Rationale
When table with many small l0 sst, it will hurt query perf, so we should support compact same table concurrently.

## Detailed Changes
- Reschdule another compact req before do compact.

## Test Plan
WIP


Rachelint (approved):
LGTM

PR #1104, opened 2023-07-25
fix: ensure shard lock is release in corner case

### LINKED ISSUE #1099: Shard lock granted, but it's not sent to ceresmeta via heartbeat
**Describe this problem**
In current open shard process, it will first try grant shard lock, and then do real `open_shard` logic.

But there is one corner case in implementation of cluster_impl.rs
- https://github.com/CeresDB/ceresdb/blob/035080c90f3ec8f2489254a105f6cd24e979da6a/cluster/src/cluster_impl.rs#L233

It will only heartbeat to meta after it's inserted into `shard_set`, this is not guaranteed, and if `get_tables_of_shards` operation failed, it will not insert to `shard_set`

**Steps to reproduce**
Hard to reproduce, it required both ceresdb and ceresmeta to restart at the same time.

**Expected behavior**
When open shard failed, it should release its shard lock, so meta can assign other node for it.

jiacai2050:
## Rationale
Close #1099

## Detailed Changes
- Add `retry_async`, retry in following cases, if retry still fails, panic.
  - `cluster.open_shard` when open shard
- Rename module `future_cancel` to `future_ext`

## Test Plan
UT


components/future_ext/src/retry.rs
```diff
@@ -0,0 +1,95 @@
...
+    F: Fn() -> Fut,
+    Fut: Future<Output = Result<T, E>>,
+{
+    for _ in 0..config.max_retries - 1 {
+        let result = f().await;
+
+        if result.is_ok() {
+            return result;
+        }
+        tokio::time::sleep(config.interval).await;
```
ShiKaiWi, line 34: Shall we support enlarge the interval with the increased retries?

server/src/grpc/meta_event_service/mod.rs
```diff
@@ -362,10 +365,24 @@ async fn do_close_shard(ctx: &HandlerContext, shard_id: ShardId) -> Result<()> {
             msg: "fail to close shards in cluster",
         })?;
 
-    ctx.release_shard_lock(shard_id).await.map_err(|e| {
-        error!("Failed to release shard lock, shard_id:{shard_id}, err:{e}");
-        e
-    })?;
+    if future_ext::retry_async(
```
ShiKaiWi, line 368: Shall we do retry for the whole `do_close_shard`?

components/future_ext/src/retry.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+}
+
+pub async fn retry_async<F, Fut, T, E>(f: F, config: &RetryConfig) -> Fut::Output
+where
+    F: Fn() -> Fut,
+    Fut: Future<Output = Result<T, E>>,
+{
+    for _ in 0..config.max_retries - 1 {
```
ZuLiangWang, line 28: What will happen if `max_retries` less than 0? do we need to validate it?

jiacai2050: when max_retries is 0, the for loop won't get executed, so no retry happens.

server/src/grpc/meta_event_service/mod.rs
jiacai2050: Not required, we only need to ensure shard lock is released after cluster.close_shard succeed.

ShiKaiWi (approved):

PR #1111, opened 2023-07-27
chore: remove unused codes about the page index

ShiKaiWi:
## Rationale
The code about the page index is not used for now and remove it.

## Detailed Changes
Remove the unused codes about the page index.

## Test Plan
No need.


Rachelint (approved):
LGTM

PR #1113, opened 2023-07-28
fix: logic of keep scheduling compaction

baojinri:
## Rationale
See title

## Detailed Changes
- Remove ! in logic of keep scheduling compaction

## Test Plan
- Existing tests
- mannually tests


jiacai2050 (approved):
LGTM

PR #1116, opened 2023-07-28
chore: remove the codes about the reverse reading

ShiKaiWi:
## Rationale
Currently, the read in a reverse order of primary key is not supported yet. Remove the codes about that.

## Detailed Changes
Remove the codes about the reverse read during plan stage and reading sst stage, while the ability to scan memtable in reverse order is kept for possible future use.

## Test Plan
Existing tests.


jiacai2050 (approved):
LGTM

PR #1118, opened 2023-07-31
fix: enable query sub-table by grpc

baojinri:
## Rationale
See title

## Detailed Changes
Change `enable_partition_table_access` to true in grpc query

## Test Plan
- Existing tests
- manually tests


integration_tests/prom/nohup.out
```diff
@@ -0,0 +1 @@
+nohup: ./prometheus-2.43.0.linux-amd64/prometheus: No such file or directory
```
tanruixiang, line 1: Was this submitted accidentally?

PR #1119, opened 2023-08-01
chore: check response header when query failed in proxy

chunshao90:
## Detailed Changes
Check response header when query failed in proxy.

### DIFF
--- a/proxy/src/http/sql.rs
+++ b/proxy/src/http/sql.rs
@@ -12,6 +12,7 @@ use common_types::{
     record_batch::RecordBatch,
 };
 use generic_error::BoxError;
+use http::StatusCode;
 use interpreters::interpreter::Output;
 use query_engine::executor::{Executor as QueryExecutor, RecordBatchVec};
 use serde::{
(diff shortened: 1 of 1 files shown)


proxy/src/http/sql.rs
```diff
@@ -162,6 +163,14 @@ fn convert_records(records: RecordBatchVec) -> Response {
 }
 
 fn convert_sql_response_to_output(sql_query_response: SqlQueryResponse) -> Result<Output> {
```
jiacai2050, line 165: A better fix is to change SqlQueryRespones to Result<_>

proxy/src/http/sql.rs
```diff
@@ -162,6 +163,14 @@ fn convert_records(records: RecordBatchVec) -> Response {
...
+    if let Some(header) = sql_query_response.header {
+        if header.code as u16 != StatusCode::OK.as_u16() {
+            return InternalNoCause {
```
jiacai2050, line 168: ErrNoCause

jiacai2050 (approved):
LGTM

PR #1121, opened 2023-08-01
fix: meta service change to meta_runtime

jiacai2050:
## Rationale
Currently meta service is running inside default runtime, which may cause some operation of meta service take very long time.

## Test Plan
No need.


ShiKaiWi (approved):
LGTM

PR #1122, opened 2023-08-01
feat: add obkv operation metrics

jiacai2050:
## Rationale
In one of our deployment,  we see lots of obkv timeout requests, so we need to add some metrics for trouble shooting.

## Detailed Changes
Add metrics for different obkv operations.

### DIFF
--- a/components/table_kv/src/metrics.rs
+++ b/components/table_kv/src/metrics.rs
@@ -4,12 +4,12 @@ use lazy_static::lazy_static;
 use prometheus::{exponential_buckets, register_histogram_vec, HistogramVec};
 
 lazy_static! {
-    // Buckets: 0.001, .., 0.01 * 2^10
+    // Buckets: 0.001, .., 0.001 * 2^15 = 32.7s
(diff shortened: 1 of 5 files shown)


components/table_kv/src/metrics.rs
```diff
@@ -0,0 +1,15 @@
...
+use lazy_static::lazy_static;
+use prometheus::{exponential_buckets, register_histogram_vec, HistogramVec};
+
+lazy_static! {
+    // Buckets: 0.001, .., 0.01 * 2^10
+    pub static ref OBKV_OP_DURATION_HISTOGRAM: HistogramVec = register_histogram_vec!(
+        "obkv_op_duration",
+        "Histogram for duration of different obkv operations",
+        &["type"],
+        exponential_buckets(0.001, 2.0, 10).unwrap()
```
ShiKaiWi, line 12: The maximum value is 1.024s and I guess this is too small for some requests with long latency.

jiacai2050: fixed

ShiKaiWi (approved):
LGTM

PR #1124, opened 2023-08-02
chore: add metric for manifest recover

Rachelint:
## Rationale
Now, no way to inspect process of manifest recover, we should make it more observable.

## Detailed Changes
Add metric for manifest recover.

## Test Plan
Not need.


jiacai2050 (approved):
LGTM

PR #1125, opened 2023-08-02
feat: dedup requests in proxy

baojinri:
## Rationale
In order to reduce the execution of the same query requests，then reduce memory consumption when querying.

## Detailed Changes
Add dedup requests in proxy

## Test Plan
Existing tests


ShiKaiWi (approved):
LGTM

PR #1126, opened 2023-08-03
feat: remove filter plan node in pipeline

### LINKED ISSUE #341: Remove filter plan node in pipeline
### Describe This Problem

For filter pushdown, we only pushdown partial filters to table scan now, that means the filter plan node is still preserved. For example:

`Query`:
```
SELECT  * FROM system.public.tables where table_name='1';
```

`Plan`:
```
| logical_plan  | Projection: #system.public.tables.timestamp, #system.public.tables.catalog, #system.public.tables.schema, #system.public.tables.table_name, #system.public.tables.table_id, #system.public.tables.engine
  Filter: #system.public.tables.table_name = Utf8("1")
    TableScan: system.public.tables projection=[timestamp, catalog, schema, table_name, table_id, engine], partial_filters=[#system.public.tables.table_name = Utf8("1")] |
| physical_plan | ProjectionExec: expr=[timestamp@0 as timestamp, catalog@1 as catalog, schema@2 as schema, table_name@3 as table_name, table_id@4 as table_id, engine@5 as engine]
  CoalesceBatchesExec: target_batch_size=4096
    FilterExec: table_name@3 = 1
      ScanTable: table=tables, parallelism=8, order=None,                                                                                                                         
```

### Proposal

We have three derives of `Table` trait:
https://github.com/CeresDB/ceresdb/blob/28be2ed754a8dc404499b01797d0e45a412b3f65/analytic_engine/src/table/mod.rs#L84

https://github.com/CeresDB/ceresdb/blob/28be2ed754a8dc404499b01797d0e45a412b3f65/system_catalog/src/lib.rs#L91

https://github.com/CeresDB/ceresdb/blob/28be2ed754a8dc404499b01797d0e45a412b3f65/table_engine/src/memory.rs#L78

In #326, we will support filters pushdown fully in `TableImpl`. If we want to remove filter plan node in pipeline, we need add same filter logic to the other two derives.  

### Additional Context

_No response_

dust1:
## Rationale
close #341 

## Detailed Changes
I  just change `supports_filter_pushdown` in `TableProvider` to return `TableProviderFilterPushDown::Exact` to fix this

## Test Plan
pass

### DIFF
--- a/table_engine/src/provider.rs
+++ b/table_engine/src/provider.rs
@@ -219,7 +219,6 @@ impl TableProviderAdapter {
 
     fn pushdown_inner(&self, filters: &[&Expr]) -> Vec<TableProviderFilterPushDown> {
         let unique_keys = self.read_schema.unique_keys();
-
         // TODO: add pushdown check in table trait
         let options = &self.table.options();
         let is_append = matches!(options.get(UPDATE_MODE), Some(mode) if mode == "APPEND");
--- a/table_engine/src/provider.rs
+++ b/table_engine/src/provider.rs
@@ -238,7 +238,7 @@ impl TableProvider for TableProviderAdapter {
     }
 
     fn supports_filter_pushdown(&self, _filter: &Expr) -> Result<TableProviderFilterPushDown> {
-        Ok(TableProviderFilterPushDown::Exact)
+        Ok(TableProviderFilterPushDown::Inexact)
     }
 
     /// Get the type of this table for metadata/catalog purposes.
--- a/table_engine/src/provider.rs
+++ b/table_engine/src/provider.rs
@@ -23,7 +23,9 @@ use std::{
 
 use arrow::datatypes::SchemaRef;
 use async_trait::async_trait;
-use common_types::{projected_schema::ProjectedSchema, request_id::RequestId, schema::Schema, UPDATE_MODE};
+use common_types::{
+    projected_schema::ProjectedSchema, request_id::RequestId, schema::Schema, UPDATE_MODE,
+};
 use datafusion::{
     config::{ConfigEntry, ConfigExtension, ExtensionOptions},
     datasource::TableProvider,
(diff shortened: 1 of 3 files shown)


table_engine/src/provider.rs
```diff
@@ -253,7 +253,7 @@ impl TableSource for TableProviderAdapter {
     /// Tests whether the table provider can make use of a filter expression
     /// to optimize data retrieval.
     fn supports_filter_pushdown(&self, _filter: &Expr) -> Result<TableProviderFilterPushDown> {
-        Ok(TableProviderFilterPushDown::Inexact)
```
jiacai2050, line 256: L229 also need to update.

dust1: it will no pass this test case, I may need to find other things that need fix🤔
https://github.com/CeresDB/ceresdb/blob/62d38355311af987ad40e8fd5dbca4d163e19d70/integration_tests/cases/common/dml/select_filter.sql#L24-L32

jiacai2050: Only this case will failed?

dust1: and this use case is wrong
https://github.com/CeresDB/ceresdb/blob/7a701c6df8f687babeb5602dbc88c7ead07fa4a8/integration_tests/cases/env/local/system/system_tables.sql#L14-L23

jiacai2050: OK.

I will also investigate why those cases fails next week, you can try other issue first in you think this is too hard for you.

dust1: maybe I can try to fix the problem this weekend. 😥

dust1:
This issue seems to have been fixed in other pr🤔

jiacai2050:
Don't worry, the original issue isn't fixed, I will look into why we can't change to `Exact` directly.

jiacai2050:
I have found the cause why those case failed, in https://github.com/CeresDB/ceresdb/issues/605 we only allow pushdown primary key filters, so we cannot directly change this to Exact here.

dust1:
ok, I'll check other places to fix the problem

jiacai2050:
If when want to return Exact here, we need to ensure following cases are met:
1. The table is in overwrite mode, and the filter only contains primary key
2. The table is in append mode

For case 2, you need to update `check_and_build_predicate_from_filters` to pushdown all filters.
https://github.com/CeresDB/ceresdb/blob/ae17b1f6322fbf3abcc23386d685128b28c1a532/table_engine/src/provider.rs#L186

jiacai2050:
@dust1 Your implementation is fine, I will fix CI failed based on your code, please don't push again.

dust1:
Well, I was gonna fix it later😂

jiacai2050:
You can retry other issue first, I want to verify this in one of my deployment first.

jiacai2050 (approved):
LGTM

PR #1127, opened 2023-08-03
feat: hotspot record remote engine requests

jiacai2050:
## Rationale
Currently hotspot will not records remote engine requests 

## Detailed Changes
Add hotspot for remote engine

## Test Plan
No need


ShiKaiWi (approved):
LGTM

PR #1129, opened 2023-08-03
feat: make obkv wal opening more parallelly

Rachelint:
## Rationale
When initing obkv wal, we need to check and create bucket table one by one...
However this process can be runned in parallel.

## Detailed Changes
Check and create tables in parallel when opening obkv wal.

## Test Plan
Test manually.

### DIFF
--- a/wal/src/table_kv_impl/namespace.rs
+++ b/wal/src/table_kv_impl/namespace.rs
@@ -1237,7 +1237,7 @@ impl TableOperator {
         let wal_shard_num_per_group = wal_shard_num / wal_shard_group_num;
         let wal_shard_groups = wal_shards
             .chunks(wal_shard_num_per_group)
-            .map(|a| a.to_owned())
+            .map(|group| group.to_owned())
             .collect::<Vec<_>>();
         let (tx, rx) = std::sync::mpsc::channel();
         let stop = Arc::new(AtomicBool::new(false));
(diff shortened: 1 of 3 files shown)


wal/src/table_kv_impl/namespace.rs
```diff
@@ -1202,28 +1221,132 @@ impl<T: TableKv> Namespace<T> {
...
+                }
+                Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {
+                    info!(
+                        "TableOperator monitor tables creating periodically, namespace:{namespace}"
+                    );
+                    continue;
+                }
+            };
+
+            cur_running_tasks -= 1;
```
ShiKaiWi, line 1277: Shall we move this line to L1256?

wal/src/table_kv_impl/namespace.rs
```diff
@@ -1202,28 +1221,132 @@ impl<T: TableKv> Namespace<T> {
...
+        let stop = Arc::new(AtomicBool::new(false));
+
+        // Create tables in parallel.
+        Self::do_create_table_in_parallel(namespace, table_kv, wal_shard_groups, tx, stop.clone());
+
+        // Wait for all tables created.
+        let mut cur_running_tasks = wal_shard_group_num;
+        loop {
+            if cur_running_tasks == 0 {
+                break;
```
ShiKaiWi, line 1252: Maybe we should add a log for time cost?

ShiKaiWi (approved):
LGTM

PR #1130, opened 2023-08-03
fix: avoid panic when the file is corrupted in disk cache

ShiKaiWi:
## Rationale
- The corrupted file in the disk cache will lead to the panic of the server
- The whole cache page file is fetched when only a range of the file is accessed, leading to high memory consumption

## Detailed Changes
- Store the file size in the memory for file integrity check
- Omit the corrupted file in the disk cache rather than panic
- Do IO operation outside the meta data cache lock
- Read only the involved range of a cached paged file instead of the whole file

## Test Plan
Existing tests and a newly designed test.

### DIFF
--- a/components/object_store/src/disk_cache.rs
+++ b/components/object_store/src/disk_cache.rs
@@ -27,7 +27,7 @@ use chrono::{DateTime, Utc};
 use crc::{Crc, CRC_32_ISCSI};
 use futures::stream::BoxStream;
 use hash_ext::SeaHasherBuilder;
-use log::{debug, error, info, warn};
+use log::{debug, info, warn};
 use lru::LruCache;
 use partitioned_lock::PartitionedMutex;
 use serde::{Deserialize, Serialize};
@@ -242,7 +242,7 @@ impl DiskCache {
 
         let do_persist = || async {
             if let Err(e) = self.persist_bytes(&filename, value).await {
-                error!("Failed to persist cache, file:{filename}, err:{e}");
+                warn!("Failed to persist cache, file:{filename}, err:{e}");
             }
         };
 
@@ -256,16 +256,8 @@ impl DiskCache {
             do_persist().await;
 
             // Remove the evicted file.
-            let evicted_file_path = std::path::Path::new(&self.root_dir)
-                .join(evicted_file)
-                .into_os_string()
-                .into_string()
-                .unwrap();
-
-            debug!("Evicted file:{evicted_file_path} is to be removed");
-            if let Err(e) = tokio::fs::remove_file(&evicted_file_path).await {
-                error!("Failed to remove evicted file:{evicted_file_path}, err:{e}");
-            }
+            debug!("Evicted file:{evicted_file} is to be removed");
+            self.remove_file_by_name(&evicted_file).await;
         } else {
             do_persist().await;
         }
@@ -289,7 +281,7 @@ impl DiskCache {
             Ok(ReadBytesResult::Corrupted {
                 file_size: real_file_size,
             }) => {
-                error!(
+                warn!(
                     "File:{filename} is corrupted, expect:{file_size}, got:{real_file_size}, and it will be removed",
                 );
 
@@ -303,7 +295,7 @@ impl DiskCache {
                 None
             }
             Ok(ReadBytesResult::OutOfRange) => {
-                error!(
+                warn!(
                     "File:{filename} is not enough to read, range:{range:?}, file_size:{file_size}, and it will be removed",
                 );
 
@@ -317,7 +309,7 @@ impl DiskCache {
                 None
             }
             Err(e) => {
-                error!("Failed to read file:{filename} from the disk cache, err:{e}");
+                warn!("Failed to read file:{filename} from the disk cache, err:{e}");
                 None
             }
         }
@@ -333,7 +325,7 @@ impl DiskCache {
             .unwrap();
 
         if let Err(e) = tokio::fs::remove_file(&file_path).await {
-            error!("Failed to remove evicted file:{file_path}, err:{e}");
+            warn!("Failed to remove file:{file_path}, err:{e}");
         }
     }
 
@@ -468,7 +460,7 @@ impl DiskCacheStore {
             assert!(cap_per_part > 0);
             Ok(LruCache::new(cap_per_part))
         };
-        let size_cache = PartitionedMutex::try_new(
+        let meta_cache = PartitionedMutex::try_new(
             init_size_lru,
             FILE_SIZE_CACHE_PARTITION_BITS,
             SeaHasherBuilder,
@@ -478,7 +470,7 @@ impl DiskCacheStore {
             cache,
             cap,
             page_size,
-            meta_cache: size_cache,
+            meta_cache,
             underlying_store,
         })
     }
@@ -568,7 +560,7 @@ impl DiskCacheStore {
             let file_size = match entry.metadata().await {
                 Ok(metadata) => metadata.len() as usize,
                 Err(e) => {
-                    error!("Failed to get the size of file:{file_name}, and it will be skipped for recover, err:{e}");
+                    warn!("Failed to get the size of file:{file_name}, and it will be skipped for recover, err:{e}");
                     // TODO: Maybe we should remove this file.
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


components/object_store/src/disk_cache.rs
```diff
@@ -296,40 +453,72 @@ impl DiskCacheStore {
...
+
+        let cache = DiskCache::try_new(cache_dir.clone(), page_num, partition_bits)?;
         Self::recover_cache(&cache_dir, &cache).await?;
 
-        let size_cache = Arc::new(Mutex::new(LruCache::new(cap / page_size)));
+        let init_size_lru = |partition_num| -> Result<_> {
+            let cap_per_part = FILE_SIZE_CACHE_CAP / partition_num;
+            assert!(cap_per_part > 0);
+            Ok(LruCache::new(cap_per_part))
+        };
```
tanruixiang, line 471: For more precise memory control, need div FILE_MATA_DATA_SIZE. `FILE_SIZE_CACHE_CAP / partition_num / FILE_MATA_DATA_SIZE`

components/object_store/src/disk_cache.rs
```diff
@@ -108,19 +100,111 @@ impl From<Error> for ObjectStoreError {
...
+/// The manifest for describing the meta of the disk cache.
 #[derive(Debug, Serialize, Deserialize)]
 struct Manifest {
     create_at: String,
     page_size: usize,
     version: usize,
 }
 
-// TODO: support partition to reduce lock contention.
+impl Manifest {
```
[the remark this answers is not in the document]
ShiKaiWi: The file name for manifest has been persisted as `manifest.json`, so I guess we have to keep this name.

components/object_store/src/disk_cache.rs
ShiKaiWi: The `FILE_SIZE_CACHE_CAP` here is the number of the cache entry rather than the size. Besides, the size of an entry in this cache is dynamic but small enough, that is to say, there is no need and no way to impose control over the memory usage by that.

components/object_store/src/disk_cache.rs
```diff
@@ -145,138 +216,214 @@ impl DiskCache {
...
+        let evicted_file = self.insert_page_meta(filename.clone(), page_meta);
+
+        let do_persist = || async {
+            if let Err(e) = self.persist_bytes(&filename, value).await {
+                error!("Failed to persist cache, file:{filename}, err:{e}");
+            }
+        };
+
+        if let Some(evicted_file) = evicted_file {
+            if evicted_file == filename {
```
jiacai2050, line 250: LRU cache will not evict the key as you just insert, if you want avoid persist, you can use `cache.contains`.

components/object_store/src/disk_cache.rs
```diff
@@ -145,138 +216,214 @@ impl DiskCache {
...
+                .join(evicted_file)
                 .into_os_string()
                 .into_string()
                 .unwrap();
 
-            debug!("Remove disk cache, filename:{}.", &file_path);
-            if let Err(e) = tokio::fs::remove_file(&file_path).await {
-                error!("Remove disk cache failed, file:{}, err:{}.", file_path, e);
+            debug!("Evicted file:{evicted_file_path} is to be removed");
+            if let Err(e) = tokio::fs::remove_file(&evicted_file_path).await {
```
jiacai2050, line 266: Replace this delete with `remove_file_by_name` fn.

components/object_store/src/disk_cache.rs
```diff
@@ -145,138 +216,214 @@ impl DiskCache {
...
+    async fn insert_data(&self, filename: String, value: Bytes) {
+        let page_meta = {
+            let file_size = PageFileEncoder::encoded_size(value.len());
+            PageMeta { file_size }
+        };
+        let evicted_file = self.insert_page_meta(filename.clone(), page_meta);
+
+        let do_persist = || async {
+            if let Err(e) = self.persist_bytes(&filename, value).await {
+                error!("Failed to persist cache, file:{filename}, err:{e}");
```
jiacai2050, line 245: Since disk cache is not a hard dependency, I suggest we replace all error log with warn.

components/object_store/src/disk_cache.rs
```diff
@@ -308,112 +452,180 @@ impl DiskCacheStore {
...
+        let cache = DiskCache::try_new(cache_dir.clone(), page_num, partition_bits)?;
         Self::recover_cache(&cache_dir, &cache).await?;
 
-        let size_cache = Arc::new(Mutex::new(LruCache::new(cap / page_size)));
+        let init_size_lru = |partition_num| -> Result<_> {
+            let cap_per_part = FILE_SIZE_CACHE_CAP / partition_num;
+            assert!(cap_per_part > 0);
+            Ok(LruCache::new(cap_per_part))
+        };
+        let size_cache = PartitionedMutex::try_new(
```
jiacai2050, line 471: meta_cache

components/object_store/src/disk_cache.rs
```diff
@@ -120,19 +110,100 @@ impl From<Error> for ObjectStoreError {
...
 struct DiskCache {
     root_dir: String,
-    // Cache key is used as filename on disk.
-    cache: PartitionedMutexAsync<LruCache<String, ()>, SeaHasherBuilder>,
+    meta_cache: PartitionedMutex<PageMetaCache, SeaHasherBuilder>,
+}
+
+#[derive(Debug, Clone)]
+struct FileMeta {
+    last_modified: DateTime<Utc>,
```
jiacai2050, line 196: When will this be used?

components/object_store/src/disk_cache.rs
```diff
@@ -308,112 +452,180 @@ impl DiskCacheStore {
...
-
-        (start..end.min(max_size))
-            .step_by(self.page_size)
-            .map(|start| start..(start + self.page_size).min(max_size))
-            .collect::<Vec<_>>()
-    }
-
-    fn cache_key(location: &Path, range: &Range<usize>) -> String {
+    /// Generate the filename of a page file.
+    fn page_filename(location: &Path, range: &Range<usize>) -> String {
```
jiacai2050, line 585: `page_cache_name` is better than this name.

components/object_store/src/disk_cache.rs
```diff
@@ -445,83 +657,123 @@ impl ObjectStore for DiskCacheStore {
...
+                page_start += self.page_size;
+                page_idx += 1;
+            }
         }
 
-        // there are multiple aligned ranges for one request, such as
-        // range = [3, 33), page_size = 16, then aligned ranges will be
-        // [0, 16), [16, 32), [32, 48)
-        // we need to combine those ranged bytes to get final result bytes
+        // Fetch all the missing pages from the underlying store.
```
jiacai2050, line 727: I think original comments are helpful to understand logic here.

components/object_store/src/disk_cache.rs
```diff
@@ -445,83 +657,123 @@ impl ObjectStore for DiskCacheStore {
...
+        if num_missing_pages > 0 {
+            let mut missing_ranges = Vec::with_capacity(num_missing_pages);
```
jiacai2050, line 729: I think we can early return when missing pages is zero, when we can remove this big if block.

components/object_store/src/disk_cache.rs
tanruixiang:
```
    /// Pushes a key-value pair into the cache. If an entry with key `k` already exists in
    /// the cache or another cache entry is removed (due to the lru's capacity),
    /// then it returns the old entry's key-value pair. Otherwise, returns `None`.
```

components/object_store/src/disk_cache.rs
ShiKaiWi: When `head` method is called.

components/object_store/src/disk_cache.rs
ShiKaiWi: After this `if` block, there's a common procedure to concat the paged_bytes, and the concat codes will be called twice if so.

components/object_store/src/disk_cache.rs
ShiKaiWi: Agreed.

jiacai2050 (approved):
LGTM

Issue #1112, opened 2023-07-28
Support remote scan in physical plan level


Rachelint:
### Describe This Problem

As the first step of #1108 , I prefer to refactor our impl for remote scanning.The refactor maybe includes following domain:
+ Make `RemoteEngine` able to execute physical plan in remote nodes.
+ Identify and rewrite the physical plan nodes to a remote executing ones(In the first step, just rewrite the table scan).

### Proposal
- [x] Impl `UnresolvedPartitionedScan`, `ResolvedPartitionedScan`, `UnresolvedSubTableScan`, `ResolvedSubTableScan`.
- [x] Generate `UnresolvedPartitionedScan` when scan partitioned table.
   - Define a dedicated `TableProviderAdapter` for `PartitionedTable`.
   - Judge and decide to generate which `TableProviderAdapter`(according to if `PartitionInfo` exists).
- [x] Define `Resolver` to convert `UnresolvedPartitionedScan` to  `ResolvedPartitionedScan`.
- [ ] Send the `UnresolvedSubTableScan` to remote and use the `SubTableScanExecutor` to run it.

### Additional Context
- Some possible optimizations:
  - Can we generate `ResolvedPartitionedScan` directly rather than generating `UnresolvedPartitionedScan`?
  - More elegant way(physical optimization rule?) to resolve the `UnresolvedPartitionedScan`?

- Communication between remote plan execution client side and server side.
  ![remote plan](https://raw.githubusercontent.com/Rachelint/drawio-store/main/remote_plan.drawio.svg)

_No response_

PR #1137, opened 2023-08-05
refactor: refactor query engine

### LINKED ISSUE #1136: The originazation of query engine crate is too messy
### Describe This Problem

Contrary to datafusion's design, leading to the strange impl for current datafusion based query engine. 
In fact, datafusion just expose `create_logical_plan` and `create_physical_plan` through `SessionContext`, and works inlucding `logical plan optimization`, `initial physical plan creation`, `phyiscal plan optimization` are hidden behind `create_physical_plan`. 
However, we want to expose hidden works in `create_physical_plan`, and the result is the messy design and impl now.

For example:
+ We optimize the `logical plan` twice... We firstly call the `optimize`, and when we call datafusion's `create_physical_plan`, `optimize` will be called interanlly again...
+ We use `physical optimizer` to build physical plan from logical plan,  and the input is `logical plan`??? We use `physical optimizer` to optimize `logical plan`???
+ We define internal `LogicalOptimizer` and `PhysicalOptimizer`, but related `optimize rule`s are not internal defined `OptimizeRule` and use ones from datafusion directly...

### Proposal

+ Follow the design of datafusion, remove `LogicalOptimzer` and `PhysicalOptimizer`, define new `PhysicalPlanner` trait to carry out the those works.
+ Split the query engine traits and detailed datafusion based impls.

### Additional Context

_No response_

Rachelint:
## Rationale
Close #1136 
Part of #1112 

The query engine impl now is too messy, you can #1136 . Worse, it can't support physical plan's remote execution, which is necessary for #1112.

In this pr, I refactor it for solving problems mentioned above.

## Detailed Changes
+ Split `physical plan` and `physical planner`'s trait definition and impl.
+ Split `physical plan`'s creation and execution.
+ Modify the call path for making it runable again.

## Test Plan
Test by exist tests.


tanruixiang:
It's strange that even though I retried CI, it still skipped the unit test.

ShiKaiWi:
@Rachelint This PR looks good to me. Let's merge it after the ci is fixed.

ShiKaiWi (approved):
LGTM

PR #1140, opened 2023-08-07
fix: record remote engine requests

jiacai2050:
## Rationale
#1127 add support for remote engine, but in client, it should be for server.

## Detailed Changes
- Remove remote client record, add add for remote server.

## Test Plan
Manually

### DIFF
--- a/server/src/grpc/remote_engine_service/mod.rs
+++ b/server/src/grpc/remote_engine_service/mod.rs
@@ -31,7 +31,10 @@ use common_types::record_batch::RecordBatch;
 use futures::stream::{self, BoxStream, FuturesUnordered, StreamExt};
 use generic_error::BoxError;
 use log::{error, info};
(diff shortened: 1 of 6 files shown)


server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -196,6 +200,9 @@ impl<Q: QueryExecutor + 'static, P: PhysicalPlanner> RemoteEngineServiceImpl<Q,
                 REMOTE_ENGINE_GRPC_HANDLER_DURATION_HISTOGRAM_VEC
                     .stream_read
                     .observe(instant.saturating_elapsed().as_secs_f64());
+
+                record_read(&ctx.hotspot_recorder, &table).await;
```
ShiKaiWi, line 204: Shall we do recording at L436?

ShiKaiWi (approved):
LGTM

PR #1150, opened 2023-08-14
refactor: add support_pushdown in table trait

jiacai2050:
## Rationale
Followup PR of #1126, remove hard code condition check .

## Detailed Changes
- Add `support_pushdown` 
- Replace unique_keys with is_unique_column to avoid unnecessary Vec allocate.
- Remove old memory table pushdown tests, all memory table shouldn't support pushdown. Pushdown check is ensured in `integration_tests/cases/common/dml/issue-341.sql`

## Test Plan
Existing tests


common_types/src/schema.rs
```diff
@@ -884,34 +884,33 @@ impl Schema {
...
+        self.columns()
+            .iter()
+            .filter_map(|column| {
+                if column.is_tag {
+                    Some(column.name.as_str())
+                } else {
+                    None
+                }
+            })
+            .any(|tag_column| tag_column == column)
```
ShiKaiWi, line 913:
```suggestion
            .any(|col| col.is_tag && col.name == column)
```

common_types/src/schema.rs
```diff
@@ -884,34 +884,33 @@ impl Schema {
...
-                .filter_map(|column| {
-                    if column.is_tag {
-                        Some(column.name.as_str())
-                    } else {
-                        None
-                    }
-                })
-                .collect::<Vec<_>>();
-            keys.extend([&self.tsid_column().unwrap().name, self.timestamp_name()]);
+    pub fn is_unique_column(&self, column: &str) -> bool {
```
ShiKaiWi, line 887: `col_name` may be a better name here:
```suggestion
    pub fn is_unique_column(&self, col_name: &str) -> bool {
```

table_engine/src/table.rs
```diff
@@ -487,6 +487,11 @@ pub trait Table: std::fmt::Debug {
     /// Get table's statistics.
     fn stats(&self) -> TableStats;
 
+    /// Whether the columns used in filter expr can be pushdown
+    fn support_pushdown(&self, _read_schema: &Schema, _columns: &[String]) -> bool {
```
ShiKaiWi, line 491: `read_schema` seems useless here, and the column name should be converted if there is any column alias.

And shall we use `Iterator<&str>` as the type for `columns`  to make it more general?

common_types/src/schema.rs
jiacai2050: fixed

table_engine/src/table.rs
jiacai2050: Add comments explaining why.

table_engine/src/table.rs
```diff
@@ -487,6 +487,11 @@ pub trait Table: std::fmt::Debug {
     /// Get table's statistics.
     fn stats(&self) -> TableStats;
 
+    /// Whether the columns used in filter expr can be pushdown....
+    /// `read_schema` is used here to avoid upper layer see different schema
+    /// during one query.
```
ShiKaiWi, line 492:
```suggestion
    /// Whether the columns used in filter expr can be pushdown.
    ///
    /// `read_schema` is used here to avoid upper layer see different schema
    /// during one query.
```

ShiKaiWi (approved):
LGTM

PR #1153, opened 2023-08-15
refactor: add request id in context

jiacai2050:
## Rationale
Currently when query failed, the error log doesn't contain request id, this makes it's very hard to tell how long the error query takes.

## Detailed Changes
- Add `request_id` in context, and print this ctx when query failed.
- Remove `enable_partition_table_access`, `runtime` from context to avoid unnecessary clone.
- Remove unnecessary param's move, change to reference.

### DIFF
--- a/proxy/src/context.rs
+++ b/proxy/src/context.rs
@@ -16,6 +16,7 @@
 
 use std::time::Duration;
 
+use common_types::request_id::RequestId;
 use macros::define_result;
 use snafu::{ensure, Backtrace, Snafu};
 
@@ -42,15 +43,16 @@ define_result!(Error);
 /// Context for request, may contains
 /// 1. Request context and options
 /// 2. Info from http headers
+#[derive(Debug)]
 pub struct RequestContext {
     /// Catalog of the request
     pub catalog: String,
     /// Schema of request
     pub schema: String,
-    /// Enable partition table_access flag
-    pub enable_partition_table_access: bool,
     /// Request timeout
     pub timeout: Option<Duration>,
+    /// Request id
+    pub request_id: RequestId,
 }
 
 impl RequestContext {
@@ -63,7 +65,6 @@ impl RequestContext {
 pub struct Builder {
     catalog: String,
     schema: String,
-    enable_partition_table_access: bool,
     timeout: Option<Duration>,
 }
 
@@ -75,11 +76,6 @@ impl Builder {
 
     pub fn schema(mut self, schema: String) -> Self {
         self.schema = schema;
-        self
-    }
-
-    pub fn enable_partition_table_access(mut self, enable_partition_table_access: bool) -> Self {
-        self.enable_partition_table_access = enable_partition_table_access;
         self
     }
 
@@ -95,8 +91,8 @@ impl Builder {
         Ok(RequestContext {
             catalog: self.catalog,
             schema: self.schema,
-            enable_partition_table_access: self.enable_partition_table_access,
             timeout: self.timeout,
+            request_id: RequestId::next_id(),
         })
     }
 }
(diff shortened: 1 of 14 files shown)


proxy/src/read.rs
```diff
@@ -108,7 +111,7 @@ impl<Q: QueryExecutor + 'static, P: PhysicalPlanner> Proxy<Q, P> {
             !stmts.is_empty(),
             ErrNoCause {
                 code: StatusCode::BAD_REQUEST,
-                msg: format!("No valid query statement provided, sql:{sql}",),
+                msg: "No valid query statement provided",
```
jiacai2050, line 114: sql in appended in upper layer.

proxy/src/grpc/sql_query.rs
```diff
@@ -144,13 +141,12 @@ impl<Q: QueryExecutor + 'static, P: PhysicalPlanner> Proxy<Q, P> {
...
         let (tx, rx) = mpsc::channel(STREAM_QUERY_CHANNEL_LEN);
-        let runtime = ctx.runtime.clone();
         let resp_compress_min_length = self.resp_compress_min_length;
         let output = self
             .as_ref()
-            .fetch_sql_query_output(ctx, &schema, &req.sql)
+            .fetch_sql_query_output(ctx, schema, &req.sql, false)
             .await?;
-        runtime.spawn(async move {
+        Handle::current().spawn(async move {
```
ShiKaiWi, line 149: Current time is the `ctx.runtime`? I guess using the `ctx.runtime` is better for clear.

proxy/src/lib.rs
```diff
@@ -554,10 +553,19 @@ impl<Q: QueryExecutor + 'static, P: PhysicalPlanner> Proxy<Q, P> {
...
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Context {
-    pub timeout: Option<Duration>,
-    pub runtime: Arc<Runtime>,
-    pub enable_partition_table_access: bool,
-    pub forwarded_from: Option<String>,
+    request_id: RequestId,
+    timeout: Option<Duration>,
+    forwarded_from: Option<String>,
```
ShiKaiWi, line 560: If the runtime is removed from the context, how to switch to a different runtime when the workload is taken on.

proxy/src/grpc/sql_query.rs
jiacai2050: Currently there is no this requirements, and if we want to support, we can add explicit runtime to specific API.

Context is used in all request type, like route/query/write/stream, and only stream API have this requirement, I think put runtime in context is little heavy.

ShiKaiWi: That's ok to separate the runtime from the context, but for some cases, switching to a new runtime to execute the task is necessary and this mechanism should be kept.

proxy/src/context.rs
```diff
@@ -42,15 +43,16 @@ define_result!(Error);
 /// Context for request, may contains
 /// 1. Request context and options
 /// 2. Info from http headers
+#[derive(Debug)]
 pub struct RequestContext {
```
Rachelint, line 47: What's the relationship between `RequestContext` and `Context`(in proxy/src/lib.rs)?
I found the fileds in them are similar, can we just keep one?

proxy/src/http/prom.rs
```diff
@@ -171,7 +164,7 @@ impl<Q: QueryExecutor + 'static, P: PhysicalPlanner> Proxy<Q, P> {
             .await?;
 
         let cost = begin_instant.saturating_elapsed().as_millis();
-        debug!("Query handler finished, request_id:{request_id}, cost:{cost}ms, query:{query:?}");
+        info!("Handle prom remote query sucess, ctx:{ctx:?}, cost:{cost}ms");
```
Rachelint, line 167: Will info level lead to too many logs?

proxy/src/context.rs
jiacai2050: RequestContext is mainly used in HTTP handler, and convert to Context to use.

Maybe we can combine them into one. I can try this in another PR.

proxy/src/http/prom.rs
jiacai2050: Will add slow query threshold to reduce log, for now print all query logs

ShiKaiWi (approved):
LGTM

PR #1158, opened 2023-08-16
feat: support columar encoding for datums

ShiKaiWi:
## Rationale
The encoding format for wal log is not compact, which imposes too much overhead on the underlying wal service.

## Detailed Changes
- Introduce the columnar encoding framework sub-module in the codec module.
- Provide some basic examples of implementation for ValuesEncoder/ValuesDecoder.

## Test Plan
New unit tests.

## Future works
Implement specific encoders for every data type.

### DIFF
--- a/components/codec/src/columnar/mod.rs
+++ b/components/codec/src/columnar/mod.rs
@@ -21,12 +21,17 @@ use bytes_ext::{Buf, BufMut};
 use common_types::{
     datum::{Datum, DatumKind, DatumView},
     row::bitset::{BitSet, RoBitSet},
+    string::StringBytes,
 };
 use macros::define_result;
 use snafu::{self, ensure, Backtrace, OptionExt, ResultExt, Snafu};
 
-use self::int::I32ValuesDecoder;
-use crate::{columnar::int::I32ValuesEncoder, Decoder};
+use self::{
+    float::{F64ValuesDecoder, F64ValuesEncoder},
+    int::{I32ValuesDecoder, I64ValuesDecoder, I64ValuesEncoder},
+    string::{StringValuesDecoder, StringValuesEncoder},
+};
+use crate::{columnar::int::I32ValuesEncoder, varint, Decoder};
 
 mod float;
 mod int;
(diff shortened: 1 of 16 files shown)


components/codec/src/columnar/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+
+/// The implementation for [`ValuesEncoder`].
+struct ValuesEncoderImpl;
+
+/// The implementation for [`ValuesDecoder`].
+struct ValuesDecoderImpl;
+
+#[derive(Clone, Debug)]
+pub struct ColumnarEncoder {
+    column_idx: u32,
```
jiacai2050, line 105: ColumnId is better than idx.

components/codec/src/columnar/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+    }
+
+    fn compute_num_datums<'a, I>(&mut self, datums: &I) -> usize
+    where
+        I: Iterator<Item = DatumView<'a>> + Clone,
+    {
+        if let Some(v) = self.num_datums {
+            v
+        } else {
+            let num_datums = datums.clone().count();
```
jiacai2050, line 136: Invoke Iterator's size_hint?

components/codec/src/columnar/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// Refer to the [encode](ColumnarEncoder::encode) method.
+    #[inline]
+    const fn header_size() -> usize {
+        1 + 1 + 4 + 4 + 4
+    }
+
+    /// The layout of the final serialized bytes:
+    /// ```plaintext
+    /// +-------------+----------------+-----------------+-----------------+----------------+---------------+---------------------+
+    /// | version(u8) | datum_kind(u8) | column_idx(u32) | num_datums(u32) | num_nulls(u32) | nulls_bit_set | non-null data block |
```
jiacai2050, line 162: How about put nulls_bit_set at the end of buf, then when decode, we  will combine decode_without_nulls and decode_with_nulls into one function.

components/codec/src/columnar/mod.rs
ShiKaiWi: What we need is an exact count of the iterator, and I guess `size_hint` can't meet this requirement.

components/codec/src/columnar/mod.rs
ShiKaiWi: 👍

components/codec/src/columnar/mod.rs
ShiKaiWi: I guess it's good to separate the two methods according to whether null values are contained.

jiacai2050 (approved):
LGTM

PR #1160, opened 2023-08-17
chore: define `QueryEngine` and wrap all things into it

Rachelint:
## Rationale
When I add `PhysicalPlanEncoder`, I found there have been too many sub-modules in `query_engine` crate now... And I foudn in fact we always use them together, so I decide to define `QueryEngine` trait to integrate them.

## Detailed Changes
+ Add `PhysicalPlanEncoder` and its datafusion impl.
+ Refactor `Executor`.
+ Define `QueryEngine` to integrate `PhysicalPlanner`, `PhysicalPlanEncoder`, `Executor`.

## Test Plan
Test by exist tests.


query_engine/src/encoding.rs
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{fmt, sync::Arc};
+
+use prost::bytes::Bytes;
+
+use crate::{error::Result, physical_planner::PhysicalPlanPtr};
+
+pub trait PhysicalPlanEncoder: fmt::Debug + Send + Sync + 'static {
```
ShiKaiWi, line 21: Shall we reuse `Encoder/Decoder`?

query_engine/src/datafusion_impl/encoding.rs
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{fmt, sync::Arc};
+
+use datafusion::execution::{runtime_env::RuntimeEnv, FunctionRegistry};
+use prost::bytes::Bytes;
```
ShiKaiWi, line 18: Replace it with `bytes_ext::Bytes`?

ShiKaiWi (commented):
This PR looks very great! Lots of messy codes are deleted. I've left some comments for some trivial issues, and let's move on after they are resolved.

query_engine/src/encoding.rs
Rachelint: Encoder/Decoder is not trait object safe, I have no thought about using it here.

query_engine/src/datafusion_impl/encoding.rs
Rachelint: It is better actually for reducing the dependency.
I modify the bytes to `bytes = "1"` folling what prost does.

ShiKaiWi (approved):
LGTM

PR #1162, opened 2023-08-21
fix: add order by for integration test case

ShiKaiWi:
## Rationale
Currently, on the main branch, the integration test has failed resulting from a sql without the order by clause.

## Detailed Changes
Add order by clause for the wrong case.

## Test Plan
By the CI.


Rachelint (approved):
LGTM

PR #1163, opened 2023-08-21
refactor: pass the `TaskContext` when execute physical plan rather than holding it

Rachelint:
## Rationale
Part of #1112 
I found we holding the execution context in `DataFusionPhysicalPlanImpl`, rather than passing it as a  param when executing like what datafusion suggests to do.

It lead to problems when I plan to deserialize bytes to `DataFusionPhysicalPlanImpl`...

## Detailed Changes
+ Remove `SessionContext` from `DataFusionPhysicalPlanImpl`.
+ Pass it as a param when executing the plan.
+ Wrap all common part to `DfContextBuilder` and use it to build `SessionContext` in `DatafusionExecutorImpl` and `DatafusionPhysicalPlannerImpl`.

## Test Plan
Test by exist tests.

### DIFF
--- a/query_engine/src/physical_planner.rs
+++ b/query_engine/src/physical_planner.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{any::Any, fmt, sync::Arc};
+use std::{fmt, sync::Arc};
 
 use async_trait::async_trait;
+use datafusion::execution::TaskContext as DfTaskContext;
 use query_frontend::plan::QueryPlan;
 use table_engine::stream::SendableRecordBatchStream;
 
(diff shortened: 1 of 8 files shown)


query_engine/src/physical_planner.rs
```diff
@@ -36,10 +36,14 @@ pub type PhysicalPlannerRef = Arc<dyn PhysicalPlanner>;
...
-    fn execute(&self) -> Result<SendableRecordBatchStream>;
+    fn execute(&self, task_ctx: &dyn TaskContext) -> Result<SendableRecordBatchStream>;
 
     /// Convert internal metrics to string.
     fn metrics_to_string(&self) -> String;
 }
 
 pub type PhysicalPlanPtr = Box<dyn PhysicalPlan>;
+
+pub trait TaskContext: fmt::Debug {
```
ShiKaiWi, line 47: Maybe a solid struct is better than a trait here.

PR #1165, opened 2023-08-21
feat: implement compression encoding for columar bytes values

ShiKaiWi:
## Rationale
Current encoding of columnar bytes is naive, without any consideration about compression.

## Detailed Changes
Introduce the compression encoding based on a designed threshold for columnar bytes.

## Test Plan
New unit tests.

### DIFF
--- a/components/codec/src/columnar/bytes.rs
+++ b/components/codec/src/columnar/bytes.rs
@@ -15,8 +15,8 @@
 use std::io::{Read, Write};
 
 use bytes_ext::{Buf, BufMut, Bytes, WriterOnBufMut};
+use lz4_flex::frame::{FrameDecoder as Lz4Decoder, FrameEncoder as Lz4Encoder};
 use snafu::{ensure, ResultExt};
-use zstd::{Decoder as ZstdStreamDecoder, Encoder as ZstdStreamEncoder};
 
 use crate::{
     columnar::{
@@ -35,7 +35,7 @@ use crate::{
 ///
 /// Currently, the `compression` has two optional values:
 /// - 0: No compression over the data block
-/// - 1: the data block is compressed by ZSTD
+/// - 1: the data block will be compressed if it is too long
 ///
 /// And the lengths in the `length block` are encoded in varint.
 /// And the reason to put `length_block_len` and `compression` at the footer is
@@ -44,16 +44,13 @@ struct Encoding;
 
 impl Encoding {
     const COMPRESSION_SIZE: usize = 1;
-    /// If the data block exceeds this threshold, it will be compressed.
-    const COMPRESS_BYTES_THRESHOLD: usize = 256;
     const LENGTH_BLOCK_LEN_SIZE: usize = 4;
     const VERSION: u8 = 0;
     const VERSION_SIZE: usize = 1;
-    const ZSTD_LEVEL: i32 = 3;
 
-    fn decide_compression(data_block_len: usize) -> Compression {
-        if data_block_len > Self::COMPRESS_BYTES_THRESHOLD {
-            Compression::Zstd
+    fn decide_compression(data_block_len: usize, threshold: usize) -> Compression {
+        if data_block_len > threshold {
+            Compression::Lz4
         } else {
             Compression::NoCompression
         }
@@ -62,7 +59,7 @@ impl Encoding {
     fn decode_compression(v: u8) -> Result<Compression> {
         let version = match v {
             0 => Compression::NoCompression,
-            1 => Compression::Zstd,
+            1 => Compression::Lz4,
             _ => InvalidCompression { flag: v }.fail()?,
         };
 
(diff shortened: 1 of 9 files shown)


components/codec/src/columnar/bytes.rs
```diff
@@ -12,37 +12,119 @@
...
+        let compression =
+            Encoding::decide_compression(data_block_len, self.bytes_compress_threshold);
+        match compression {
+            Compression::NoCompression => {
+                for v in values {
+                    buf.put_slice(v);
+                }
+            }
+            Compression::Lz4 => encode_with_compress(buf, values).context(Compress)?,
         }
```
Rachelint, line 112: How about just put the these logic into `Encoding` ?

components/codec/src/columnar/bytes.rs
```diff
@@ -52,19 +134,90 @@ impl<'a> ValuesEncoder<&'a [u8]> for ValuesEncoderImpl {
...
+    reused_buf: &mut Vec<u8>,
+    f: F,
+) -> Result<()>
+where
+    F: FnMut(Bytes) -> Result<()>,
+{
+    let mut decoder = Lz4Decoder::new(compressed_data_block_buf);
+    decoder.read_to_end(reused_buf).context(Decompress)?;
+    decode_without_compression(&mut length_block_buf, &reused_buf[..], f)
+}
```
Rachelint, line 184: Why not defining as `Encoding`'s internal method? And just expose the `encode` method?

components/codec/src/columnar/bytes.rs
```diff
@@ -12,37 +12,119 @@
...
     {
-        for v in values {
-            debug_assert!(v.len() < u32::MAX as usize);
+        // Encode the `version`.
+        buf.put_u8(Encoding::VERSION);
+
+        // Encode the `length_block`.
+        let mut data_block_len = 0;
+        let mut length_block_len = 0;
+        for v in values.clone() {
```
jiacai2050, line 95: why clone is required here?

components/codec/src/columnar/bytes.rs
```diff
@@ -12,37 +12,119 @@
...
+/// ```
+///
+/// Currently, the `compression` has two optional values:
+/// - 0: No compression over the data block
+/// - 1: the data block will be compressed if it is too long
+///
+/// And the lengths in the `length block` are encoded in varint.
+/// And the reason to put `length_block_len` and `compression` at the footer is
+/// for friendly decode.
+struct Encoding;
```
Rachelint, line 43: How about name it DataBlockEncoding or DataEncoding?

components/codec/src/columnar/bytes.rs
```diff
@@ -12,37 +12,119 @@
...
+/// The layout for the string/bytes:
+/// ```plaintext
+/// +-------------+--------------+------------+-----------------------+-----------------+
+/// | version(u8) | length_block | data_block | length_block_len(u32) | compression(u8) |
+/// +-------------+--------------+------------+-----------------------+-----------------+
+/// ```
+///
+/// Currently, the `compression` has two optional values:
+/// - 0: No compression over the data block
+/// - 1: the data block will be compressed if it is too long
```
jiacai2050, line 38:
```suggestion
/// - 1: the data block will be compressed using lz4 if it is too long
```

components/codec/src/columnar/bytes.rs
```diff
@@ -52,19 +134,90 @@ impl<'a> ValuesEncoder<&'a [u8]> for ValuesEncoderImpl {
...
+        let compression_offset = chunk.len() - Encoding::COMPRESSION_SIZE;
+        let compression = Encoding::decode_compression(chunk[compression_offset])?;
+
+        let length_block_len_offset = chunk.len() - footer_len;
+        let length_block_end = {
+            let mut len_buf = &chunk[length_block_len_offset..compression_offset];
+            len_buf.get_u32() as usize + Encoding::VERSION_SIZE
+        };
+        let mut length_block = &chunk[Encoding::VERSION_SIZE..length_block_end];
+        let data_block = &chunk[length_block_end..length_block_len_offset];
```
Rachelint, line 214: Add some comments here?

components/codec/src/columnar/bytes.rs
ShiKaiWi: The iterator will be consumed if not clone.

components/codec/src/columnar/bytes.rs
ShiKaiWi: The data block can be compressed by not only lz4.

components/codec/src/columnar/bytes.rs
ShiKaiWi: This encoding is for the whole layout, not just data block.

Rachelint (approved):
LGTM

PR #1166, opened 2023-08-22
chore: add metrics for write logs in wal

ShiKaiWi:
## Rationale
Currently, there is no metrics to report the distribution of wal log's size, which helps to observe the encoding results of the wal entries.

## Detailed Changes
Introduce a metrics for wal log size.

## Test Plan
Write some data, and query metrics api to check the result.


jiacai2050 (approved):
LGTM

PR #1167, opened 2023-08-22
feat: implement compression encoding for columar timestamp values 

ZuLiangWang:
## Rationale
The columnar compression encoding for timestamp is not supported.

## Detailed Changes
Supported compression for the encoding for timestamp.

## Test Plan
New unit tests.

### DIFF
--- a/components/codec/src/bits/buffered_read.rs
+++ b/components/codec/src/bits/buffered_read.rs
@@ -53,9 +53,9 @@ impl<'a> BufferedReader<'a> {
         let byte = self.get_byte()?;
 
         let bit = if byte & BIT_MASKS[self.bit_idx as usize] == 0 {
-            Bit::Zero
+            Bit(0)
         } else {
-            Bit::One
+            Bit(1)
         };
 
         self.bit_idx += 1;
@@ -105,7 +105,7 @@ impl<'a> BufferedReader<'a> {
 
         while num > 0 {
             self.next_bit()
-                .map(|bit| bits = bits.wrapping_shl(1) | bit.to_u64())?;
+                .map(|bit| bits = bits.wrapping_shl(1) | bit.0 as u64)?;
 
             num -= 1;
         }
@@ -123,23 +123,23 @@ mod tests {
         let bytes = vec![0b01101100, 0b11101001];
         let mut b = BufferedReader::new(&bytes);
 
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
 
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
 
         assert_eq!(b.next_bit().err().unwrap(), Error::Eof);
     }
@@ -155,10 +155,10 @@ mod tests {
 
         // read some individual bits we can test `read_byte` when the position in the
         // byte we are currently reading is non-zero
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
-        assert_eq!(b.next_bit().unwrap(), Bit::One);
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
+        assert_eq!(b.next_bit().unwrap(), Bit(1));
 
         assert_eq!(b.next_byte().unwrap(), 15);
 
@@ -182,7 +182,7 @@ mod tests {
         let bytes = vec![0b01101101, 0b01101101];
         let mut b = BufferedReader::new(&bytes);
 
-        assert_eq!(b.next_bit().unwrap(), Bit::Zero);
+        assert_eq!(b.next_bit().unwrap(), Bit(0));
         assert_eq!(b.next_bits(3).unwrap(), 0b110);
         assert_eq!(b.next_byte().unwrap(), 0b11010110);
         assert_eq!(b.next_bits(2).unwrap(), 0b11);
--- a/components/codec/src/bits/buffered_read.rs
+++ b/components/codec/src/bits/buffered_read.rs
@@ -1,4 +1,18 @@
-use crate::bits::{Bit, Error, Read};
+// Copyright 2023 The CeresDB Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::bits::{Bit, Error};
 
 /// BufferedReader
 ///
@@ -25,8 +39,8 @@ impl<'a> BufferedReader<'a> {
     }
 }
 
-impl<'a> Read for BufferedReader<'a> {
-    fn read_bit(&mut self) -> Result<Bit, Error> {
+impl<'a> BufferedReader<'a> {
+    pub fn read_bit(&mut self) -> Result<Bit, Error> {
         if self.pos == 8 {
             self.index += 1;
             self.pos = 0;
@@ -45,7 +59,7 @@ impl<'a> Read for BufferedReader<'a> {
         Ok(bit)
     }
 
-    fn read_byte(&mut self) -> Result<u8, Error> {
+    pub fn read_byte(&mut self) -> Result<u8, Error> {
         if self.pos == 0 {
             self.pos += 8;
             return self.get_byte();
@@ -71,7 +85,7 @@ impl<'a> Read for BufferedReader<'a> {
 
     // example: read_bits(4): read data:u64 0000 0000 0000 000f
     // datastore in low position
-    fn read_bits(&mut self, mut num: u32) -> Result<u64, Error> {
+    pub fn read_bits(&mut self, mut num: u32) -> Result<u64, Error> {
         // can't read more than 64 bits into a u64
         if num > 64 {
             num = 64;
@@ -94,7 +108,8 @@ impl<'a> Read for BufferedReader<'a> {
         Ok(bits)
     }
 
-    fn peak_bits(&mut self, num: u32) -> Result<u64, Error> {
+    #[allow(dead_code)]
+    pub fn peak_bits(&mut self, num: u32) -> Result<u64, Error> {
         // save the current index and pos so we can reset them after calling `read_bits`
         let index = self.index;
         let pos = self.pos;
@@ -110,7 +125,7 @@ impl<'a> Read for BufferedReader<'a> {
 
 #[cfg(test)]
 mod tests {
-    use crate::bits::{Bit, BufferedReader, Error, Read};
+    use crate::bits::{Bit, BufferedReader, Error};
 
     #[test]
     fn read_bit() {
--- a/components/codec/src/bits/buffered_read.rs
+++ b/components/codec/src/bits/buffered_read.rs
@@ -1,0 +1,209 @@
+use crate::bits::{Bit, Error, Read};
+
+/// BufferedReader
+///
+/// BufferedReader encapsulates a buffer of bytes which can be read from.
+#[derive(Debug)]
+pub struct BufferedReader<'a> {
+    bytes: &'a [u8], // internal buffer of bytes
+    index: usize,    // index into bytes
+    pos: u32,        // position in the byte we are currently reading
+}
+
+impl<'a> BufferedReader<'a> {
+    /// new creates a new `BufferedReader` from `bytes`
+    pub fn new(bytes: &'a [u8]) -> Self {
+        BufferedReader {
+            bytes,
+            index: 0,
+            pos: 0,
+        }
+    }
+
+    fn get_byte(&mut self) -> Result<u8, Error> {
+        self.bytes.get(self.index).cloned().ok_or(Error::Eof)
+    }
+}
+
+impl<'a> Read for BufferedReader<'a> {
+    fn read_bit(&mut self) -> Result<Bit, Error> {
+        if self.pos == 8 {
+            self.index += 1;
+            self.pos = 0;
+        }
+
+        let byte = self.get_byte()?;
+
+        let bit = if byte & 1u8.wrapping_shl(7 - self.pos) == 0 {
+            Bit::Zero
+        } else {
+            Bit::One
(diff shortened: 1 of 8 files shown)


components/codec/src/bits/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            Error::Eof => "Encountered the end of the stream",
+        }
+    }
+}
+
+/// Read
+///
+/// Read is a trait that encapsulates the functionality required to read from a
+/// stream of bytes.
+pub trait Read {
```
ShiKaiWi, line 50: The `Read` and `Write` seems unnecessary, and it also leads to confusion with the `std::io::Read` and `std::io::Write`.

components/codec/src/columnar/mod.rs
```diff
@@ -134,6 +137,10 @@ struct ValuesEncoderImpl {
 /// The implementation for [`ValuesDecoder`].
 struct ValuesDecoderImpl;
 
+struct TimestampEncoderImpl;
```
ShiKaiWi, line 140: Move the `TimestampEncoderImpl` and `TimestampDecoderImpl` to the `timestamp.rs`.

components/codec/src/columnar/mod.rs
```diff
@@ -255,7 +262,7 @@ impl ColumnarEncoder {
         let enc = ValuesEncoderImpl::default();
         let data_size = match hint.datum_kind {
             DatumKind::Null => 0,
-            DatumKind::Timestamp => enc.estimated_encoded_size(
+            DatumKind::Timestamp => TimestampEncoderImpl.estimated_encoded_size(
```
ShiKaiWi, line 265: Shall use the `ValuesEncoderImpl` here? we can make `ValuesEncoderImpl` to delegate the encoding work with the `TimestampEncoderImpl` as the underlying encoder.

components/codec/src/bits/mod.rs
ZuLiangWang: Remove useless trait `Read` and `Write`.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::bits::{Bit, Error};
+
+/// BufferedReader
+///
```
ShiKaiWi, line 18: Remove the useless comments. And not just here, remove these comments in the PR.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+/// BufferedReader encapsulates a buffer of bytes which can be read from.
+#[derive(Debug)]
+pub struct BufferedReader<'a> {
+    bytes: &'a [u8], // internal buffer of bytes
+    index: usize,    // index into bytes
+    pos: u32,        // position in the byte we are currently reading
```
ShiKaiWi, line 24: Put the comments above every field. And I suggest renaming the two fields `index` and `pos` to `byte_idx` and `bit_idx`.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+impl<'a> BufferedReader<'a> {
+    pub fn read_bit(&mut self) -> Result<Bit, Error> {
+        if self.pos == 8 {
+            self.index += 1;
+            self.pos = 0;
+        }
+
+        let byte = self.get_byte()?;
+
+        let bit = if byte & 1u8.wrapping_shl(7 - self.pos) == 0 {
```
ShiKaiWi, line 51: I guess define a global static mask array will provide a better performance:
```
let BIT_MASKS: [u8;8] = [1, 2, 4, 8, 16, ..., 128];
```

With the BIT_MASKS, we can put it here:
```suggestion
        let bit = if byte & BIT_MASK[self.pos] == 0 {
```

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    fn get_byte(&mut self) -> Result<u8, Error> {
+        self.bytes.get(self.index).cloned().ok_or(Error::Eof)
+    }
+}
+
+impl<'a> BufferedReader<'a> {
+    pub fn read_bit(&mut self) -> Result<Bit, Error> {
```
ShiKaiWi, line 43: I guess `next_bit` is a better name.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+        self.index += 1;
+        b = self.get_byte()?;
+
+        byte |= b.wrapping_shr(8 - self.pos);
+
+        Ok(byte)
+    }
+
+    // example: read_bits(4): read data:u64 0000 0000 0000 000f
+    // datastore in low position
```
ShiKaiWi, line 87: Use the rust doc comments, that is to say, use `///` instead of `//`.

And the comment here are too casual and informal, shall we polish it up and make it more formal?

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+            self.read_bit()
+                .map(|bit| bits = bits.wrapping_shl(1) | bit.to_u64())?;
+
+            num -= 1;
+        }
+
+        Ok(bits)
+    }
+
+    #[allow(dead_code)]
```
ShiKaiWi, line 111: If not used, how about removing it?

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,224 @@
...
+        Ok(byte)
+    }
+
+    // example: read_bits(4): read data:u64 0000 0000 0000 000f
+    // datastore in low position
+    pub fn read_bits(&mut self, mut num: u32) -> Result<u64, Error> {
+        // can't read more than 64 bits into a u64
+        if num > 64 {
+            num = 64;
+        }
```
ShiKaiWi, line 92: Let's make an assertion: `assert!(num <= 64)`. And add comments to explain the `num` parameter should not exceed 64.

components/codec/src/bits/buffered_write.rs
```diff
@@ -0,0 +1,279 @@
...
+            pos: 8,
+        }
+    }
+
+    fn grow(&mut self) {
+        self.buf.push(0);
+    }
+
+    fn last_index(&self) -> usize {
+        self.buf.len() - 1
```
ShiKaiWi, line 61: If the buf is empty, it may overflow.

components/codec/src/bits/buffered_write.rs
```diff
@@ -0,0 +1,279 @@
...
+
+use crate::bits::Bit;
+
+/// BufferedWriter
+///
+/// BufferedWriter writes bytes to a buffer.
+#[derive(Debug, Default, Clone)]
+pub struct BufferedWriter {
+    buf: Vec<u8>,
+    pos: u32, // position in the last byte in the buffer
```
ShiKaiWi, line 25: I guess `pos` should be strictly in the range [0, 7], but I find the 8 is used to sentinel, and if so, you have to check the `pos` to ensure the validness every time it is used. And I guess this is a bad pattern.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// new creates a new `BufferedReader` from `bytes`
+    pub fn new(bytes: &'a [u8]) -> Self {
+        BufferedReader {
+            bytes,
+            byte_idx: 0,
+            bit_idx: 0,
+        }
+    }
+
+    fn get_byte(&mut self) -> Result<u8, Error> {
```
ShiKaiWi, line 39: I guess the `self.byte_idx` is always valid, so there is no need to use `self.bytes.get` here?

And here a read-only self reference is enough here.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        Some(byte)
+    }
+
+    /// Fetch the next `num` bits, and advance the inner cursor. And the
+    /// returned bits are stored in a u64 integer in little-endian order.
+    ///
+    /// Example: the returned value will be `0x0000 0000 0000 000F` if 4 bits is
+    /// fetched and they are all set.
+    pub fn next_bits(&mut self, mut num: u32) -> Option<u64> {
```
ShiKaiWi, line 103: It may perform better if follow these steps:
1. Read some bits to make the `bit_idx` == 0;
2. Read the remaining bits byte by byte;
3. Read the remaining bits which can't form a byte;

If this implementation is adopted, `next_byte` can be implemented more simply.

components/codec/src/columnar/mod.rs
```diff
@@ -88,6 +88,9 @@ pub enum Error {
 
     #[snafu(display("Bytes is not enough, length:{len}.\nBacktrace:\n{backtrace}"))]
     NotEnoughBytes { len: usize, backtrace: Backtrace },
+
+    #[snafu(display("Failed to read encoded data"))]
+    ReadEncode {},
```
ShiKaiWi, line 93: If there is no cause error, please attach the backtrace to it.

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+                    first = false;
+                    last_ts = v;
+                    if num == 1 {
+                        writer.write_bit(Bit(0))
+                    } else {
+                        writer.write_bit(Bit(1))
+                    }
+                    continue;
+                }
+                let cur_delta = v - last_ts;
```
ShiKaiWi, line 126: Shall we consider the overflow?

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+                let delta_of_delta = cur_delta - last_delta;
+                match TimestampEncoder::encode_delta_of_delta_masks(delta_of_delta) {
+                    DeltaOfDeltaEncodeMasks::Zero() => {
+                        writer.write_bit(Bit(0));
+                        dod_encode_ts_len += 1;
+                    }
+                    DeltaOfDeltaEncodeMasks::Normal(
+                        control_bis,
```
ShiKaiWi, line 134:
```suggestion
                        control_bits,
```

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+const END_MARKER: u64 = 0b1111_0000_0000_0000_0000_0000_0000_0000_0000;
+/// END_MARKER_LEN is the length, in bits, of END_MARKER
+const NUM_BITS_END_MARKER_LEN: usize = 36;
+
+// Encode the timestamp using the delta-of-delta algorithm
+struct TimestampEncoder {}
+
+struct TimestampDecoder {}
+
+enum DeltaOfDeltaEncodeMasks {
```
ShiKaiWi, line 52: Could you explain these variants?

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+        }
+    }
+
+    pub fn encode<B, I>(buf: &mut B, values: I) -> Result<()>
+    where
+        B: BufMut,
+        I: Iterator<Item = i64>,
+    {
+        let (lower, higher) = values.size_hint();
+        let num = lower.max(higher.unwrap_or_default());
```
ShiKaiWi, line 86: The size hint is not ensured to be exactly correct, it is just a hint.

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+                    if num == 1 {
+                        writer.write_bit(Bit(0))
+                    } else {
+                        writer.write_bit(Bit(1))
+                    }
+                    continue;
+                }
+                let cur_delta = v - last_ts;
+                let delta_of_delta = cur_delta - last_delta;
+                match TimestampEncoder::encode_delta_of_delta_masks(delta_of_delta) {
```
ShiKaiWi, line 128:
```suggestion
                match Self::encode_delta_of_delta_masks(delta_of_delta) {
```

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+use crate::{
+    bits::{Bit, BufferedReader, BufferedWriter},
+    columnar::{
+        DecodeContext, InvalidVersion, NotEnoughDatums, ReadEncode, Result, ValuesDecoder,
+        ValuesDecoderImpl, ValuesEncoder, ValuesEncoderImpl,
+    },
+};
+
+/// We use delta-of-delta algorithm to compress timestamp, for details, please refer to this paper http://www.vldb.org/pvldb/vol8/p1816-teller.pdf
+/// The layout for the timestamp:
```
ShiKaiWi, line 29:
```suggestion
/// The layout for the encoded payload:
```

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+        I: Iterator<Item = Timestamp>,
+    {
+        let (lower, higher) = values.size_hint();
+        let num = lower.max(higher.unwrap_or_default());
+        NUM_BYTES_ENCODE_VERSION_LEN
+            + 1
+            + NUM_BITS_END_MARKER_LEN
+            + NUM_BYTES_TIMESTAMP_LEN
+            + NUM_BYTES_TIMESTAMP * num
+            + 1
```
ShiKaiWi, line 299: The two 1 byte stands for which part?

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+    pub fn decode<B, F>(buf: &B, mut f: F) -> Result<()>
+    where
+        B: Buf,
+        F: FnMut(Timestamp) -> Result<()>,
+    {
+        let mut reader = BufferedReader::new(buf.chunk());
+
+        let version = reader
+            .next_bits((NUM_BYTES_ENCODE_VERSION_LEN * 8) as u32)
+            .context(ReadEncode)? as u8;
```
ShiKaiWi, line 192: The `version` can be retrieved by `buf.get_u8()`.

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+        let first_timestamp = reader.next_bits(64).context(ReadEncode)?;
+        f(Timestamp::new(first_timestamp as i64))?;
+
+        let control_bit = reader.next_bit().context(ReadEncode)?;
+        if control_bit == Bit(0) {
+            return Ok(());
+        }
+
+        let mut last_delta: u64 = 0;
+        let mut last_timestamp = first_timestamp;
```
ShiKaiWi, line 204:
```suggestion
        let mut prev_delta: u64 = 0;
        let mut prev_timestamp = first_timestamp;
```

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+        loop {
+            let mut dod_control_bits_size = 0;
+            // 0,10,110,1110,1111
```
ShiKaiWi, line 207: I can't understand the comment.

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+            }
+        }
+
+        buf.put_slice(writer.close().as_ref());
+        Ok(())
+    }
+}
+
+impl TimestampDecoder {
+    fn encode_delta_of_delta_masks(dod_control_bits_size: i32) -> DeltaOfDeltaEncodeMasks {
```
ShiKaiWi, line 172: Would you `impl From<i32> for DeltaOfDeltaEncodeMasks`?

components/codec/src/columnar/timestamp.rs
ZuLiangWang: Here, the reader is used uniformly to read data, and in order to read subsequent data, we need to advance the inner cursor inside the reader.

components/codec/src/columnar/timestamp.rs
ZuLiangWang: No need, `cur_delta` only overflow when the two timestamp is negative, I think we do not need to consider this situation.

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+        // variables used to encode dod
+        let mut first = true;
+        let mut last_ts: i64 = 0;
+        let mut last_delta: i64 = 0;
+
+        for v in values {
+            if !need_fallback {
+                if first {
+                    // store the first timestamp exactly
+                    writer.write_bits(v as u64, 64);
```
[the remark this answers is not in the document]
ZuLiangWang: It has no effect, and timestamp cannot be negative under normal circumstances, I think this check should be done by the upper layer if it needs.

components/codec/src/columnar/timestamp.rs
```diff
@@ -11,3 +11,301 @@
...
+            }
+        );
+
+        let mut writer = BufferedWriter::with_capacity(0);
+
+        // write version
+        writer.write_bits(
+            ENCODE_VERSION as u64,
+            (8 * NUM_BYTES_ENCODE_VERSION_LEN) as u32,
+        );
```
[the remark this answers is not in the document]
ZuLiangWang: `writer.write_byte` only support one byte input, `write_bytes` is currently not implemented.

components/codec/src/columnar/timestamp.rs
ZuLiangWang: Add some comments to explain the use of these variants.

components/codec/src/bits/buffered_read.rs
```diff
@@ -0,0 +1,201 @@
...
+        // can't read more than 64 bits into a u64
+        assert!(num <= 64);
+
+        let mut bits: u64 = 0;
+        while num >= 8 {
+            let byte = self.next_byte().map(u64::from)?;
+            bits = bits.wrapping_shl(8) | byte;
```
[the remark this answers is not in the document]
ZuLiangWang: Is there any good way to replace it?

components/codec/src/columnar/timestamp.rs
ShiKaiWi: After `buf.get_u8()`, the remaining bytes can be used to build the bit reader.

components/codec/src/columnar/timestamp.rs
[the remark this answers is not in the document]
ShiKaiWi: Actually, timestamp can be nagative.

components/codec/src/columnar/timestamp.rs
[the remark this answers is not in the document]
ShiKaiWi: Isn't `ENCODE_VERSION` a byte?

components/codec/src/bits/buffered_read.rs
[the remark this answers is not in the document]
ShiKaiWi: The shift operator: `<<` and `>>`.

components/codec/src/columnar/timestamp.rs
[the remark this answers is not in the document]
ZuLiangWang: Sorry, I read it wrong, it is indeed a byte

components/codec/src/columnar/timestamp.rs
ZuLiangWang: `delta` will over flow if `v - last_ts` > ` 9223372036854775807`,  which means the time difference is `922337203685477` seconds, It is impossible to appear in normal time series data.

PR #1170, opened 2023-08-23
feat: implement boolean columnar encoding

ShiKaiWi:
## Rationale
The columnar compression encoding for boolean is not supported.

## Detailed Changes
Supported compression for the encoding for boolean.

## Test Plan
New unit tests.

### DIFF
--- a/common_types/src/row/bitset.rs
+++ b/common_types/src/row/bitset.rs
@@ -46,6 +46,19 @@ impl BitSet {
         }
     }
 
+    /// Create a u8 according to a given 8bits array.
+    ///
+    /// The values in the `bits` whose index is greater than 8 will be ignored.
+    pub fn one_byte(bits: &[bool]) -> u8 {
+        let mut v = 0u8;
+        for (idx, set) in bits.iter().take(8).map(|v| *v as u8).enumerate() {
+            let (_, bit_idx) = RoBitSet::compute_byte_bit_index(idx);
+            v |= set << bit_idx
+        }
+
+        v
+    }
+
     /// Initialize a [`BitSet`] with all bits set.
     pub fn all_set(num_bits: usize) -> Self {
         Self {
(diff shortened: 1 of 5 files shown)


components/codec/src/columnar/bool.rs
```diff
@@ -0,0 +1,245 @@
...
+        Ok(())
+    }
+
+    fn encode_without_compression<B, I>(buf: &mut B, values: I) -> Result<()>
+    where
+        B: BufMut,
+        I: Iterator<Item = bool>,
+    {
+        for v in values {
+            buf.put_u8(v as u8);
```
jiacai2050, line 133: use bit to represent bool?

components/codec/src/columnar/bool.rs
```diff
@@ -0,0 +1,245 @@
...
+                buf.put_u8(bit_set);
+
+                // Reset the offset and the bits buf.
+                offset = 0;
+                one_byte_bits = [false; 8];
+            }
+        }
+
+        // Put the remaining bits.
+        if offset > 0 {
```
jiacai2050, line 182: what if last bit is 0, will this lost one element?

common_types/src/row/bitset.rs
```diff
@@ -46,6 +46,19 @@ impl BitSet {
         }
     }
 
+    /// Create a u8 according to a given 8bits array.
+    ///
+    /// The values in the `bits` whose index is greater than 8 will be ignored.
+    pub fn one_byte(bits: &[bool]) -> u8 {
```
jiacai2050, line 52: `as_u8`

components/codec/src/columnar/bool.rs
ShiKaiWi: This encoding method is chosen when the number of boolean is small, which may lead to
low compression ratio.

components/codec/src/columnar/bool.rs
ShiKaiWi: This is offset, the last element has been encoded into the buf if it is equal 0.

jiacai2050 (approved):
LGTM

PR #1171, opened 2023-08-24
feat: make float as number encoding

ShiKaiWi:
## Rationale
Encoding float values in columnar way is not supported yet.

## Detailed Changes
Support encode float values.

## Test Plan
New unit tests.


jiacai2050 (approved):
LGTM

PR #1172, opened 2023-08-24
chore: remove hybrid related logic

jiacai2050:
## Rationale
Hybrid format have no obvious advantages, and maintain it in our codebase need lots of work, so we decide to remove from our codebase.

## Test Plan
Existing UT


analytic_engine/src/table_options.rs
```diff
@@ -345,7 +323,9 @@ impl From<manifest_pb::StorageFormat> for StorageFormat {
     fn from(format: manifest_pb::StorageFormat) -> Self {
         match format {
             manifest_pb::StorageFormat::Columnar => Self::Columnar,
-            manifest_pb::StorageFormat::Hybrid => Self::Hybrid,
+            manifest_pb::StorageFormat::Hybrid => {
+                panic!("Hybrid format is already deprecated, and cannot be used anymore!")
```
ShiKaiWi, line 327: I guess we should throw an error rather than panic.

ShiKaiWi (approved):
LGTM

PR #1177, opened 2023-08-25
feat: teach ceresdb to encode/decode datafusion physical plan

Rachelint:
## Rationale
Part of #1112 
If we want to send plan to remote, we must encode it first.

## Detailed Changes
+ Impl extension codec of dist_sql_query related physical plans.
+ Remove the useless `PhysicalPlanCodec` and its impl.
+ Refactor the pb converting about `ReadRequest`.

## Test Plan
Test by new ut.


df_engine_extensions/src/codec.rs
```diff
@@ -0,0 +1,110 @@
...
+            DataFusionError::Internal(format!("failed to decode extension physical plan, err{e}"))
+        })?;
+
+        let typed_extension = extension_node
+            .typed_extension
+            .ok_or(DataFusionError::Internal(
+                "typed extension not found".to_string(),
+            ))?;
+
+        for typed_codec in &self.typed_codecs {
```
jiacai2050, line 73: It seems strange to loop codec to decode Plan, a more elegant way is to match, something like:
```
match extension {
  A => DecoderA.try_decode()...
  B => DecoderB.try_decode()...
}
```

Rachelint: It is similar to chain of responsibility pattern?
Each sub codec check if it can encode/decode the plan. If not just return None, and let next sub codec to try it.
We can just modify `PhysicalExtensionCodecImpl::new` to register and support a new codec (e.g. promql)?

jiacai2050 (approved):
LGTM

PR #1179, opened 2023-08-28
feat: support write wal logs in columnar format

ShiKaiWi:
## Rationale
After the columnar encoding is supported for wal format, we should adapt such encoding in the wal write procedure.

## Detailed Changes
Adapt the columnar encoding in the wal write procedure. And now, both rowwise and columnar encoding in the wal write procedure are supported, which to use depends on the configuration.

## Test Plan
Existing tests. The compatibility tests have been designed as:
- Write rowwise-encoded wal logs and restart the server to replay it
- Write columnar-encoded wal logs and restart the server to replay it

### DIFF
--- a/analytic_engine/src/payload.rs
+++ b/analytic_engine/src/payload.rs
@@ -16,15 +16,20 @@
 
 use bytes_ext::{Buf, BufMut, SafeBuf, SafeBufMut};
 use ceresdbproto::{manifest as manifest_pb, table_requests};
-use codec::{row::WalRowDecoder, Decoder};
+use codec::{
+    columnar::{ColumnarDecoder, DecodeContext, DecodeResult},
+    row::WalRowDecoder,
+    Decoder,
+};
 use common_types::{
-    row::{RowGroup, RowGroupBuilder},
+    row::{RowGroup, RowGroupBuilder, RowGroupBuilderFromColumn},
     schema::Schema,
+    table::TableId,
 };
 use macros::define_result;
 use prost::Message;
 use snafu::{Backtrace, OptionExt, ResultExt, Snafu};
-use wal::log_batch::{Payload, PayloadDecoder};
+use wal::log_batch::{Payload, PayloadDecodeContext, PayloadDecoder};
 
 use crate::{table_options, TableOptions};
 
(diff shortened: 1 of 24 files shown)


ShiKaiWi:
My local test has found that compatibility problem: restart with columnar wal encoding format fails after write some data in rowwise wal encoding format.

common_types/src/row/mod.rs
```diff
@@ -421,6 +394,88 @@ impl<'a> Iterator for IterCol<'a> {
...
+                }
+            }
+        }
+
+        let rows = rows.into_iter().map(Row::from_datums).collect::<Vec<_>>();
+
+        RowGroup {
+            schema: self.schema,
+            rows,
+            min_timestamp: Timestamp::now(),
```
Rachelint, line 468: Seems unnecessary to call `now`, and just use a fixed default?

common_types/src/datum.rs
```diff
@@ -1274,6 +1274,13 @@ impl<'a> DatumView<'a> {
         }
     }
 
+    pub fn as_time(&self) -> Option<Timestamp> {
```
Rachelint, line 1277: Just name it `timestamp`?
`Time` may be a specific type as I remembered?

analytic_engine/src/payload.rs
```diff
@@ -163,10 +176,21 @@ pub enum ReadPayload {
 }
 
 impl ReadPayload {
-    fn decode_write_from_pb(buf: &[u8]) -> Result<Self> {
+    fn decode_write_from_pb(schema: Schema, buf: &[u8]) -> Result<Self> {
```
Rachelint, line 179: Seem schema is just used in the `Columnar` path, how about pass the reference and just do the late clone in related path?

common_types/src/row/mod.rs
```diff
@@ -421,6 +394,88 @@ impl<'a> Iterator for IterCol<'a> {
...
+        }
+
+        let old = self.cols.insert(col_id, col);
+        ensure!(old.is_none(), DuplicateColumnId { column_id: col_id });
+
+        Ok(())
+    }
+
+    pub fn build(mut self) -> RowGroup {
+        let num_rows = self.num_rows().unwrap_or(0);
```
Rachelint, line 431: Early return in the conercases?

components/codec/src/columnar/mod.rs
```diff
@@ -680,6 +683,72 @@ mod tests {
         check_encode_end_decode(10, datums, DatumKind::Int64);
     }
 
+    #[test]
```
Rachelint, line 686: Maybe we should add some cases for the bad case?
Such as zero columns or zero rows?

common_types/src/datum.rs
ShiKaiWi: `as_timestamp` has been used for the variant `Datum::Timestamp`.

analytic_engine/src/payload.rs
ShiKaiWi: 👍

common_types/src/row/mod.rs
ShiKaiWi: Oh. Maybe the correct timestamp should be computed here.

common_types/src/datum.rs
Rachelint: How about just adding a branch to original `as_timestamp`?

common_types/src/row/mod.rs
Rachelint: > Oh. Maybe the correct timestamp should be computed here.

It seems time range is unnecessary actually? Maybe default value is enough?

ShiKaiWi: Maybe we can't make such an assumption.

jiacai2050 (approved):
LGTM

ShiKaiWi:
50% WAL payload is reduced after this patch applied to the production environment.

PR #1189, opened 2023-09-01
feat: use array for small size map

jiacai2050:
## Rationale
When benchmark, cost of  schema's index_of is obvious large.

## Detailed Changes
- Use array based map to avoid unnecessary hash.

## Test Plan
UT

### DIFF
--- a/common_types/src/schema.rs
+++ b/common_types/src/schema.rs
@@ -34,6 +34,7 @@ use prost::Message;
 use snafu::{ensure, Backtrace, OptionExt, ResultExt, Snafu};
 
 use crate::{
+    array_map::ArrayMap,
     column_schema::{self, ColumnId, ColumnSchema},
     datum::DatumKind,
     row::{contiguous, RowView},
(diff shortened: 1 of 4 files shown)


ShiKaiWi:
@jiacai2050 Any benchmark data to support this PR?

common_types/src/schema.rs
```diff
@@ -380,7 +381,7 @@ pub(crate) struct ColumnSchemas {
     columns: Vec<ColumnSchema>,
     /// Column name to index of that column schema in `columns`, the index is
     /// guaranteed to be valid
-    name_to_index: HashMap<String, usize>,
+    name_to_index: ArrayMap<String, usize>,
```
ShiKaiWi, line 384: If the cost comes from the hash computing, `BTreeMap` may work better?

jiacai2050: I don't know how node inside BTreeMap is stored, ArrayMap is optimized for `get` when map size is small, array is the best structure I can think of.

jiacai2050:
No benchmark is done now, I think we can deploy this online, and check flamegraph again.

jiacai2050:
Flamegraph show this map is worse than before.

PR #1193, opened 2023-09-04
feat: support random partition rule

ShiKaiWi:
## Rationale
Currently, hash/key rule are the only supported rules for partition table. However, it will lead to small write batches if the user tries to use primary key or all tags as the partition key for a uniform distribution. And it is known that the small write batches hurt write performance badly.

Actually, for a partition table in `append` mode with primary key or all tags as partition key, most of the queries to this table will fan out to all the partitions because the partition key won't be provided in most cases.  So what is the gain from the cost (the massive small write batches) to do such complex partition?

It will work much better if a simple partition rule is supported, by which a random partition is chosen for a given write request.

## Detailed Changes
- Introduce `RandomPartitionRule`
- Add a new method called `validate_create_table` to `TableEngine` trait

## Test Plan
Add unit test to apply `RandomPartitionRule` for a partition table in different mode.


table_engine/src/engine.rs
```diff
@@ -315,6 +313,9 @@ pub trait TableEngine: Send + Sync {
     /// Close the engine gracefully.
     async fn close(&self) -> Result<()>;
 
+    /// Validate the request of create table.
+    async fn validate_create_table(&self, request: &CreateTableParams) -> Result<()>;
```
Rachelint, line 317: Add a comment to explain why this method is needed? I think it seems a strange acutally...

query_frontend/src/parser.rs
```diff
@@ -332,8 +332,14 @@ impl<'a> Parser<'a> {
         let table_name = self.parser.parse_object_name()?.into();
         let (columns, constraints) = self.parse_columns()?;
 
-        // PARTITION BY...
-        let partition = self.maybe_parse_and_check_partition(Keyword::PARTITION, &columns)?;
+        // Parse the simple partition rule, starting with `Partitions $Num`.
```
Rachelint, line 335: Why not make it `partition random partitions xxx`?
It seems a bit strange, and I think it will mislead people that the `random partitioning` is default partitioning way in ceresdb?

ShiKaiWi: Sounds reasonable.

Rachelint (approved):
LGTM

PR #1199, opened 2023-09-07
fix: deadlock when dedup stream read

ShiKaiWi:
## Rationale
The deadlock will happen when dedup stream read requests because the serial execution of sending and receiving data on a bounded channel.

## Detailed Changes
Send the data in background.

## Test Plan
Test it manually.

### DIFF
--- a/server/src/grpc/remote_engine_service/mod.rs
+++ b/server/src/grpc/remote_engine_service/mod.rs
@@ -14,7 +14,11 @@
 
 // Remote engine rpc service implementation.
 
-use std::{hash::Hash, sync::Arc, time::Instant};
+use std::{
+    hash::Hash,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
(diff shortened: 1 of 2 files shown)


server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -237,27 +236,45 @@ impl RemoteEngineServiceImpl {
...
         // We should set cancel to guard, otherwise the key will be removed twice.
         guard.cancel();
         let notifiers = request_notifiers.take_notifiers(&request_key).unwrap();
+        self.runtimes.read_runtime.spawn(async move {
+            Self::send_dedupped_resps(resps, notifiers).await;
+
+            REMOTE_ENGINE_GRPC_HANDLER_DURATION_HISTOGRAM_VEC
+                .stream_read
+                .observe(instant.saturating_elapsed().as_secs_f64());
+        });
```
Rachelint, line 259: How about place these logics of First to Frist match branch?

jiacai2050 (approved):
LGTM

PR #1200, opened 2023-09-07
fix: use ExecutionGuard to ensure notifiers released when futures got cancelled

tanruixiang:
## Rationale
Add ExecutionGuard to ensure notifiers go released when future got cancelled during `get_bytes`, otherwise all ranges belonging to this notifiers will never receive response.

## Detailed Changes
1. Add a fast path when no need_fetch_block is found
2. Use ExecutionGuard to ensure notifier released.

## Test Plan
Old UT.


components/object_store/src/disk_cache.rs
```diff
@@ -434,6 +434,35 @@ impl Paging {
...
+            filename,
+        }
+    }
+
+    pub fn get_filename(&self) -> &String {
+        &self.filename
+    }
+
+    /// Generate the filename of a page file.
+    pub fn page_cache_name(location: &Path, range: &Range<usize>) -> String {
```
jiacai2050, line 456: This functions should read `self.range`, `self.location`, rather than pass in new args

jiacai2050 (approved):

PR #1201, opened 2023-09-07
feat: add table statistics api

ZuLiangWang:
## Rationale
At present, it often happens that some tables fail to open when opening a shard, but we cannot count the specific opening status. We add an interface to count the opening status of the shard.

## Detailed Changes
* Add an interface to count the opening status of the shard

### DIFF
--- a/catalog/src/table_operator.rs
+++ b/catalog/src/table_operator.rs
@@ -12,12 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Instant;
+use std::{collections::HashMap, time::Instant};
 
 use generic_error::BoxError;
 use log::{error, info, warn};
 use snafu::{OptionExt, ResultExt};
-use table_engine::{engine, table::TableRef};
+use table_engine::{
+    engine,
+    table::{TableId, TableRef},
+};
 use time_ext::InstantExt;
 
 use crate::{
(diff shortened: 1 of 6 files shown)


catalog/src/table_operator.rs
```diff
@@ -71,9 +78,9 @@ impl TableOperator {
             .context(TableOperatorWithCause { msg: None })?;
 
         // Check and register successful opened table into schema.
-        let mut success_count = 0_u32;
-        let mut missing_table_count = 0_u32;
-        let mut open_table_errs = Vec::new();
+        let mut open_success_tables = HashMap::new();
+        let mut open_missing_tables = HashMap::new();
+        let mut open_failed_tables = HashMap::new();
```
tanruixiang, line 83: Why not use `OpenShardResult ` here  🤣

cluster/src/shard_operator.rs
```diff
@@ -151,20 +151,31 @@ impl ShardOperator {
...
+            .fail();
+        }
 
         info!("ShardOperator open sequentially finish, shard_id:{shard_info:?}");
 
         Ok(())
     }
 
-    pub async fn close(&self, ctx: CloseContext) -> Result<()> {
+    pub async fn close(&self, mut ctx: CloseContext) -> Result<()> {
```
chunshao90, line 178: Why use mut?

catalog/src/table_operator.rs
```diff
@@ -285,3 +302,23 @@ impl TableOperator {
             })
     }
 }
+
+pub struct OpenShardResult {
+    open_success_tables: HashMap<TableId, SchemaRef>,
```
chunshao90, line 307: Schema is useless?

cluster/src/shard_operator.rs
```diff
@@ -410,4 +421,19 @@ impl ShardOperator {
 
         Ok(())
     }
+
+    pub async fn get_open_success_tables(&self) -> Result<Vec<String>> {
+        let data = self.data.read().unwrap();
+        Ok(data.open_success_tables.clone())
```
chunshao90, line 427: Add `open_shard_result` in ShardDataRef?

PR #1202, opened 2023-09-07
chore: add meta stable check into integration test

Rachelint:
## Rationale
Currently, we just left 20s for ceresmeta to do the initialization work in integration test.
However, it will be not enough something... And the test failure will make us so panic...
In this pr, I add a meta stable check to reduce the integration test failure.

## Detailed Changes
Add a meta stable check by keep creating a radom named table utill succeeding(surely, retry max exists).

## Test Plan
Test manually.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1299,6 +1299,7 @@ dependencies = [
  "serde",
  "sqlness",
  "tokio",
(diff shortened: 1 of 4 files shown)


ShiKaiWi (approved):
LGTM

PR #1203, opened 2023-09-08
refactor: stream read metric

baojinri:
## Rationale
Now, we can't collect the time cost  of stream read

## Detailed Changes
- Use drop trait to collect time cost  of stream read

## Test Plan
Existing tests

### DIFF
--- a/server/src/grpc/remote_engine_service/mod.rs
+++ b/server/src/grpc/remote_engine_service/mod.rs
@@ -16,7 +16,9 @@
 
 use std::{
     hash::Hash,
+    pin::Pin,
     sync::Arc,
+    task::{Context, Poll},
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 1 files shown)


server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -84,35 +86,40 @@ impl StreamReadReqKey {
...
 
-struct ExecutionGuard<F: FnMut()> {
-    f: F,
-    cancelled: bool,
+pub type StreamReadRequestNotifiers =
+    Arc<RequestNotifiers<StreamReadReqKey, mpsc::Sender<Result<RecordBatch>>>>;
+pub type BoxedStream<T> = Box<dyn Stream<Item = T> + Send + Unpin>;
+
+struct ReceiverStreamWrapper {
+    inner: BoxedStream<Result<RecordBatch>>,
```
ShiKaiWi, line 94: Shall we define it as a generic params? I guess there is no need to make it as trait object.

server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -84,35 +86,40 @@ impl StreamReadReqKey {
...
 }
 
-struct ExecutionGuard<F: FnMut()> {
-    f: F,
-    cancelled: bool,
+pub type StreamReadRequestNotifiers =
+    Arc<RequestNotifiers<StreamReadReqKey, mpsc::Sender<Result<RecordBatch>>>>;
+pub type BoxedStream<T> = Box<dyn Stream<Item = T> + Send + Unpin>;
+
+struct ReceiverStreamWrapper {
```
ShiKaiWi, line 93: Maybe we can give it a better name: `StreamWithMetric`.

ShiKaiWi (approved):
LGTM

PR #1207, opened 2023-09-11
fix: lock contention on acquiring the arena stats

ShiKaiWi:
## Rationale
There is no need to acquire the write lock when fetch arena's stats.

## Detailed Changes
Use `RwLock` instead of `Mutex` to protect the core of the `MonoIncArena` to avoid lock contention.

## Test Plan
Existing tests.


jiacai2050 (approved):
LGTM

PR #1208, opened 2023-09-11
refactor: avoid memory allocation during partition write requests

ShiKaiWi:
## Rationale
Some memory allocation during partitioning write request can be avoided.

## Detailed Changes
Introduce an enum called `PartitionedRows` as the return type of `PartitionRule.locate_partitions_for_write` to avoid unnecessary memory allocation.

## Test Plan
Existing test.

### DIFF
--- a/partition_table_engine/src/partition.rs
+++ b/partition_table_engine/src/partition.rs
@@ -19,7 +19,7 @@ use std::{collections::HashMap, fmt};
 use analytic_engine::{table::support_pushdown, TableOptions};
 use async_trait::async_trait;
 use common_types::{
-    row::{Row, RowGroupBuilder},
+    row::{Row, RowGroup, RowGroupBuilder},
     schema::Schema,
 };
 use futures::{stream::FuturesUnordered, StreamExt};
@@ -27,7 +27,12 @@ use generic_error::BoxError;
 use snafu::ResultExt;
 use table_engine::{
     partition::{
-        format_sub_partition_table_name, rule::df_adapter::DfPartitionRuleAdapter, PartitionInfo,
+        format_sub_partition_table_name,
+        rule::{
+            df_adapter::DfPartitionRuleAdapter, PartitionedRow, PartitionedRowGroup,
+            PartitionedRowsIter,
+        },
+        PartitionInfo,
     },
     remote::{
         model::{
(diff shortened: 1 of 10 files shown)


partition_table_engine/src/partition.rs
```diff
@@ -84,6 +99,86 @@ impl PartitionTableImpl {
...
+        schema: Schema,
+        partitioned_rows: PartitionedRowsIter,
+    ) -> Result<usize> {
+        let mut split_rows = HashMap::new();
+        for PartitionedRow { partition_id, row } in partitioned_rows {
+            split_rows
+                .entry(partition_id)
+                .or_insert_with(Vec::new)
+                .push(row);
+        }
```
Rachelint, line 135: How about move here into `location_partitions_for_write`, and just return final `HashMap<partition, RowGroup>`?

partition_table_engine/src/partition.rs
```diff
@@ -84,6 +99,86 @@ impl PartitionTableImpl {
...
+
+        // Insert split write request through remote engine.
+        let mut request_batch = Vec::with_capacity(split_rows.len());
+        for (partition, rows) in split_rows {
+            let sub_table_ident = self.get_sub_table_ident(partition);
+            let row_group = RowGroupBuilder::with_rows(schema.clone(), rows)
```
jiacai2050, line 141: With_rows invoke ` Self::new(schema);` inside, `with_capacity` is more suitable.

table_engine/src/partition/rule/mod.rs
```diff
@@ -20,19 +20,38 @@ mod filter;
...
 
-use common_types::{datum::DatumKind, row::RowGroup};
+use common_types::row::{Row, RowGroup};
 
 use self::filter::PartitionFilter;
 use crate::partition::Result;
 
+/// The partitioned rows of the written requests.
+pub enum PartitionedRows {
+    One {
```
jiacai2050, line 30: `Single` is more suitable than `One`?

table_engine/src/partition/rule/mod.rs
```diff
@@ -48,17 +67,4 @@ pub trait PartitionRule: Send + Sync + 'static {
...
-}
-
-impl ColumnWithType {
-    pub fn new(column: String, datum_type: DatumKind) -> Self {
-        Self { column, datum_type }
-    }
-}
-
-pub type PartitionRuleRef = Box<dyn PartitionRule>;
+pub type PartitionRulePtr = Box<dyn PartitionRule>;
```
jiacai2050, line 70: IMO `PartitionRuleRef` is better, in safe Rust we don't operate on pointers.

ShiKaiWi: I guess `Ref` is usually used for `Arc`?

partition_table_engine/src/partition.rs
ShiKaiWi: I guess `RowGroupBuilder` does nothing with this PR.

table_engine/src/partition/rule/mod.rs
ShiKaiWi: Done.

jiacai2050 (approved):
LGTM

PR #1209, opened 2023-09-12
chore: modify `enable_others` default true for happy debugging

Rachelint:
## Rationale
It is too struggle to debug partitioned table because `enable_others` in `sub_table_access_perm` default to false...
When we try to query the specific sub table, we will always be prohibited... It will make us extremely unhappy...

## Detailed Changes
Modify `enable_others` default true.

## Test Plan
Test manually.


jiacai2050 (approved):
LGTM

PR #1210, opened 2023-09-12
fix: skip update shard status when create/remove table

jiacai2050:
## Rationale
In one of our deployments, I have one shard change its status from Ready to Init, there is no open/close shards requests between this change, so I add some info log to help investigate this issue.

## Detailed Changes
- Skip update shard status when create/remove table

## Test Plan
TBD


ShiKaiWi (approved):
LGTM

PR #1219, opened 2023-09-19
refactor: stop reschedule when pending task larger than max ongoing

jiacai2050:
## Rationale
Current schedule stragety is too simple, when there are many pending task, we should stop reschedule.

## Test Plan
No need.

### DIFF
--- a/analytic_engine/src/compaction/scheduler.rs
+++ b/analytic_engine/src/compaction/scheduler.rs
@@ -226,6 +226,12 @@ impl OngoingTaskLimit {
     }
 
     #[inline]
+    fn pending_task_size(&self) -> usize {
+        let buf = self.request_buf.read().unwrap();
+        buf.len()
+    }
+
+    #[inline]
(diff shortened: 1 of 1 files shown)


analytic_engine/src/compaction/scheduler.rs
```diff
@@ -480,7 +486,8 @@ impl ScheduleWorker {
         waiter_notifier: WaiterNotifier,
         token: MemoryUsageToken,
     ) {
-        let keep_scheduling_compaction = compaction_task.contains_min_level();
+        let keep_scheduling_compaction = self.limit.pending_task_size() < self.max_ongoing_tasks
```
Rachelint, line 489: I think it is like a condition for judging the degree of pending queue crowding? It seems not very relevant with `max_onging_tasks`? 
How about make it a new config item?

jiacai2050: fixed

Rachelint (approved):
LGTM

PR #1220, opened 2023-09-19
fix: ignore error when open partition table failed

jiacai2050:
## Rationale
When sub_table[0] is unhealthy, we can not drop the partition table.

## Detailed Changes
Only log open partition table error, not return to outside.


chunshao90 (approved):
LGTM

PR #1224, opened 2023-09-20
fix: loop all sub tables to get table info

jiacai2050:
## Rationale
Followup PR of #1220

## Test Plan
Manually.

### DIFF
--- a/proxy/src/lib.rs
+++ b/proxy/src/lib.rs
@@ -323,6 +323,7 @@ impl Proxy {
         part_info: &PartitionInfo,
     ) -> Result<TableInfo> {
         let get_inner = |i| async move {
+            // TODO: the remote engine should provide a method to get all sub table names.
             let sub_partition_table_name = util::get_sub_partition_name(base_name, part_info, i);
             let table = self
                 .instance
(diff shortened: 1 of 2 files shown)


proxy/src/lib.rs
```diff
@@ -314,27 +315,55 @@ impl Proxy {
...
+                .with_context(|| ErrWithCause {
+                    code: StatusCode::INTERNAL_SERVER_ERROR,
+                    msg: "Failed to get table",
+                })?;
+
+            Ok(table)
+        };
+
+        let part_num = part_info.get_definition_num();
+        if part_num == 1 {
```
chunshao90, line 348: Do not check the `partition number == 1`, because it is very likely that it will not be achieved.

proxy/src/lib.rs
```diff
@@ -314,27 +315,55 @@ impl Proxy {
...
-        // When open remote table failed, we currently don't return error outside.
-        // This is because when sub_table[0] is unhealthy, we can not drop the partition
-        // table.
-        // TODO: maybe we can find a more elegant way to deal with this issue.
-        Ok(())
+        // Loop get sub tables to get table info in case of some of them has problems.
+        for i in 0..part_info.get_definition_num() - 1 {
+            let ret = get_inner(i).await;
+            if let Err(err) = ret {
+                warn!("Failed to get table info, err:{err:?}");
```
chunshao90, line 356:
```suggestion
            if let Err(e) = ret {
                warn!("Failed to get table info, err:{e:?}");
```

proxy/src/lib.rs
```diff
@@ -314,27 +315,55 @@ impl Proxy {
...
-    ) -> Result<()> {
-        if let Err(e) = self
-            .open_partition_table_inner(catalog_name, schema_name, table_name)
-            .await
-        {
-            warn!("Open partition table failed, err:{e:?}");
+        base_name: &str,
+        part_info: &PartitionInfo,
+    ) -> Result<TableInfo> {
+        let get_inner = |i| async move {
```
chunshao90, line 325: Add `TODO`, put this logic into `partition rules` mod.

chunshao90 (approved):
LGTM

PR #1225, opened 2023-09-22
feat: store real time range in sst

jiacai2050:
## Rationale
Currently CeresDB will store aligned time range when write SST, this will cause IO amplification for certain query.

For example, the SST only contains one row, and its timestamp is 10:01, its time range in metadata will be aligned to `[10:00, 12:00)`, if query contains filter like `timestamp > 10:01`, currently query plan will include this SST, but in fact we could skip this SST to avoid unnecessary IO.

## Detailed Changes
- When do compaction, calculate time range based on every row, and store this in metadata.

## Test Plan
- Modified UT

### DIFF
--- a/analytic_engine/src/sst/parquet/writer.rs
+++ b/analytic_engine/src/sst/parquet/writer.rs
@@ -22,7 +22,7 @@ use generic_error::BoxError;
 use log::{debug, error};
 use object_store::{ObjectStoreRef, Path};
 use parquet::data_type::AsBytes;
-use snafu::ResultExt;
+use snafu::{OptionExt, ResultExt};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use super::meta_data::RowGroupFilter;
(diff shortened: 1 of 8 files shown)


common_types/src/column_block.rs
```diff
@@ -146,6 +146,27 @@ define_numeric_column!(
 #[derive(Debug, Clone)]
 pub struct TimestampColumn(TimestampMillisecondArray);
 
+impl TimestampColumn {
+    pub fn min_max(&self) -> Option<TimeRange> {
```
chunshao90, line 150: min_max -> timerange?

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -213,6 +260,14 @@ impl RecordBatchGroupWriter {
 
             let num_batches = row_group.len();
             for record_batch in row_group {
+                let column_block = record_batch.column(timestamp_index);
+                let ts_col = column_block
+                    .as_timestamp()
+                    .context(RequireTimestampColumn {
```
chunshao90, line 266: expect xxx

chunshao90 (approved):
LGTM

PR #1226, opened 2023-09-23
refactor: first step to move all packages under src folder

tisonkun:
## Rationale

The final result will be moved all packaged under the `src` folder so that we have a consistent and concise layout.

Take materialize as an example: https://github.com/MaterializeInc/materialize/tree/main/src

## Detailed Changes

First step: move `src` to `src/server` while keep all the effective code and config the same. So that we get rid of the special case first.

## Test Plan

Everything should keep AS IS and existing tests should pass.


tisonkun:
Perhaps we still need changes some other relative paths, I'd like to let CI helps me :D

src/ceresdb/bin/ceresdb-server.rs
Rachelint, line 1: How about make a `server` module and move `config.rs`, `setup.rs`, `signal_handler.rs` into it?
Because It seems they are just used by `server`.
If we have other bins needed to be builded in later, we make realted modules like this?

tisonkun: Do you mean that we rename `src/ceresdb` to `src/server` and merge this crate with the top-level `server` crate?

Rachelint: > Do you mean that we rename `src/ceresdb` to `src/server` and merge this crate with the top-level `server` crate?

Emmm, I just means maybe we should make the organization like following, beause it seems all of them are commponents for the `ceresdb-server`.
```
bin
-- ceresdb-server.rs

ceresdb
-- src
---- server
------ mod.rs
------ config.rs
------ setup.rs
------ signal_handler.rs 

-- lib.rs
```
I thinks we can move things in `tools` crate to `ceresdb` crate in later, and the dir organization can be like:
```
bin
-- ceresdb-server.rs
-- tools...

ceresdb
-- src
---- server
------ mod.rs
------ config.rs
------ setup.rs
------ signal_handler.rs 

---- tools
------ mod.rs
------ submodules...

---- lib.rs
```

tisonkun: @Rachelint I get what you mean, but for further moving, I'd prefer:

```
src
-- ceresdb
---- bin
---- src
-- tools
---- bin
---- src
```

instead of what you proposed:

```
ceresdb
-- bin
---- ceresdb-server.rs
---- tools
-- src
--- server
------ mod.rs
---- tools
------ mod.rs
---- lib.rs
```

... which seems like one crate for all.

Please read materialize as an example: https://github.com/MaterializeInc/materialize/tree/main/src

tisonkun: If I get you wrong, please directly send a PR against this branch to demostrate.

Rachelint: OK, it seems more reasonable.

jiacai2050 (approved):
LGTM

PR #1227, opened 2023-09-25
fix: set and fetch environment variables error

tanruixiang:
## Rationale
After https://github.com/CeresDB/ceresdb/pull/1226 changing directories , `build.rs`  will not be used to set environment variables, `ceresdb-server --verion` will return None.

## Detailed Changes
move `build.rs` into`src/ceresdb/`

## Test Plan
add a test case to ensure all environment variables are got.


tanruixiang:
Please review this if you have time.❤️ @tisonkun

tisonkun (approved):
Thanks for your follow up!

Rachelint (approved):
LGTM

PR #1228, opened 2023-09-26
test: add integration test for query plan

jiacai2050:
## Detailed Changes
- Bump sqlness to 0.5.0, which contains interceptor to replace random values before compare
  - https://github.com/CeresDB/sqlness/releases/tag/v0.5.0
- add query-plan test case
- add pre_cmd interceptor, which is used to flush memtable.

## Test Plan
CI


integration_tests/cases/common/dml/insert_mode.result
```diff
@@ -1,3 +1,4 @@
+-- overwrite
```
jiacai2050, line 1: In old sqlness version, it will ignore sql comments, 0.5.0 add it back.

jiacai2050: Also reviewer only need check `query-plan.sql`, others are all auto-formatted.

table_engine/src/provider.rs
```diff
@@ -414,18 +414,10 @@ impl ExecutionPlan for ScanTable {
...
-                count: Count::new(),
-            },
-            None,
-        )));
-
         let pushdown_filters = &self.request.predicate;
         metric_set.push(Arc::new(Metric::new(
             MetricValue::Count {
-                name: format!("\n{pushdown_filters:?}").into(),
+                name: format!("\n{metrics_desc}\n\n{pushdown_filters:?}").into(),
```
jiacai2050, line 420: Combine those two metrics into one, otherwise their order is random in final output.

chunshao90 (approved):
LGTM

PR #1229, opened 2023-09-26
fix: panic if dedupped query fails

ShiKaiWi:
## Rationale
Panics when failed to dedupping query.

## Detailed Changes
Send the error to the waiters rather than exiting directly.

## Test Plan
Manual tests.


Rachelint (approved):
LGTM

Issue #1222, opened 2023-09-19
Cannot build librocksdb-sys on macOS with Apple M1 chip


tisonkun:
### Describe this problem

```
  [100%] Building CXX object CMakeFiles/rocksdb.dir/build_version.cc.o
[...]

PR #1231, opened 2023-09-26
refactor!(wal): factor out wal-rocksdb

tisonkun:
This is a possible impl for https://github.com/CeresDB/ceresdb/issues/1222#issuecomment-1726786002.

But still the transitive solution introduce too much `cfg` code IMO, while otherwise the deeply coupled code cannot be easily opt out.

Send for preview and demonstrate the solution direction. Perhaps we can have a better solution.


analytic_engine/src/lib.rs
```diff
@@ -288,30 +288,41 @@ pub struct KafkaWalConfig {
...
-impl Default for RocksDBConfig {
-    fn default() -> Self {
-        Self {
-            data_dir: "/tmp/ceresdb".to_string(),
-            data_namespace: Default::default(),
-            meta_namespace: Default::default(),
+    /// Config for wal based on RocksDB
+    #[derive(Debug, Clone, Deserialize, Serialize)]
+    #[serde(default)]
+    pub struct RocksDBConfig {
```
tisonkun, line 302: This config can be moved to `wal-rocksdb` while tests `use`s should be updated.

analytic_engine/src/lib.rs
```diff
@@ -288,30 +288,41 @@ pub struct KafkaWalConfig {
...
+                meta_namespace: Default::default(),
+            }
         }
     }
 }
+
 /// Options for wal storage backend
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(tag = "type")]
 pub enum WalStorageConfig {
```
tisonkun, line 324: But this enum cannot be moved to any specific `wal-`impl, not suitable for `wal` also.

analytic_engine/src/setup.rs
```diff
@@ -158,97 +157,117 @@ pub trait WalsOpener: Send + Sync + Default {
...
-                return InvalidWalConfig {
-                    msg: format!(
-                        "invalid wal storage config while opening rocksDB wal, config:{config:?}"
-                    ),
+    /// [RocksEngine] builder.
+    #[derive(Default)]
+    pub struct RocksDBWalsOpener;
+
+    #[async_trait]
+    impl WalsOpener for RocksDBWalsOpener {
```
tisonkun, line 184: This struct should have been moved to `wal-rocksdb` but now it depends on `EngineRuntimes` and `OpenedWals`:

* `OpenedWals`: This can be possibly moved to `wal`;
* `EngineRuntimes`: This comes from table_engine which is tricky to move. But the actually used field is `component/runtime`'s `Runtime` while the certain field is not fixed so we use `engine_runtimes: Arc<EngineRuntimes>` arg here.

analytic_engine/src/lib.rs
```diff
@@ -288,30 +288,41 @@ pub struct KafkaWalConfig {
...
+    #[cfg(feature = "wal-rocksdb")]
```
jiacai2050, line 325: I think this should be not in feature gate, we should throw error at runtime when user configure to use rocksdb as WAL for binary built without this feature.

analytic_engine/src/setup.rs
jiacai2050: > This can be possibly moved to wal;

Agree

> But the actually used field is component/runtime's Runtime while the certain field is not fixed so we use engine_runtimes: Arc<EngineRuntimes> arg here.

I think you can only pass Runtime to this function, is there any other issues?

Cargo.toml
```diff
@@ -174,7 +177,10 @@ trace_metric_derive = { path = "components/trace_metric_derive" }
 trace_metric_derive_tests = { path = "components/trace_metric_derive_tests" }
 tonic = "0.8.1"
 tokio = { version = "1.29", features = ["full"] }
-wal = { path = "wal" }
+wal = { path = "src/wal" }
```
jiacai2050, line 180: I wonder if this is necessary to split different wal implementation to different crates, any advantages?

analytic_engine/src/lib.rs
tisonkun: But the `RocksDBConfig` transitively depends on rocksdb so when we build a binary without rocksdb, the RocksDB config cannot be constructed also.

Cargo.toml
tisonkun: It's possible we retain the code in wal crate and guard them with different feature gate. Factor out impl into dedicated crates is for:

* per https://github.com/CeresDB/ceresdb/issues/1222#issuecomment-1726786002 suggested
* exclude a crate can be straightforward than conditionally opt-out some code in the same crate, but perhaps we can narrow the opt code into the same mod.

Generally, I don't have a strong preference, either should be viable.

analytic_engine/src/setup.rs
tisonkun: > I think you can only pass Runtime to this function, is there any other issues?

You can check the impls:

```rust
// ObkvWalsOpener
let obkv = engine_runtimes
            .write_runtime ...
open_wal_and_manifest_with_table_kv(*obkv_wal_config, engine_runtimes, obkv).await

// MemWalsOpener
open_wal_and_manifest_with_table_kv(
            *obkv_wal_config,
            engine_runtimes,
            self.table_kv.clone(),
        )
        .await

// KafkaWalsOpener
let default_runtime = &engine_runtimes.default_runtime;

// RocksDBWalsOpener
let write_runtime = engine_runtimes.write_runtime.clone();

// ... where open_wal_and_manifest_with_table_kv is
let runtimes = WalRuntimes {
        read_runtime: engine_runtimes.read_runtime.clone(),
        write_runtime: engine_runtimes.write_runtime.clone(),
        default_runtime: engine_runtimes.default_runtime.clone(),
    };
```

So passing a bare `Runtime` may not be suitable for all these variants.

jiacai2050: Don't expect it get so complicated here...

In theory, all WAL implementation should run where the caller is, but when there are blocking API it's best to spawn on another runtime.

And WAL is a write mainly components, so I guess we only need to pass write runtime to it, other runtime make less sense since WAL are only read on server starts up, no need to pass dedicated read runtime into it.

So maybe we can try combine runtime into one first, if this is possible, then I think this PR could move further.

analytic_engine/src/lib.rs
jiacai2050: > the RocksDB config cannot be constructed also.

In this case, a dummy struct could be constructed to make compile success. Something like this:

```
#[cfg(feature = "wal-rocksdb")]
RocksDBConfig = wal-rocksdb::config
#[cfg(not(feature = "wal-rocksdb"))]
RocksDBConfig = struct {};
```

Cargo.toml
jiacai2050: OK, I also don't have strong preference, keep it as it's now.

analytic_engine/src/setup.rs
tisonkun: > So maybe we can try combine runtime into one first, if this is possible, then I think this PR could move further

This sounds like the proper way to go. Let me spend some time for investigation.

tisonkun:
Superseded by https://github.com/CeresDB/ceresdb/pull/1272.

PR #1234, opened 2023-10-01
fix: when drop table first check its existing

jiacai2050:
## Rationale
In obkv WAL implementation, when a table is deleted by one server, other servers will still delete same table,

In theory we should avoid this, and only allow a server to delete tables belonging to this server, but this will
major changes to OBKV WAL implementation, so here I add a check before do delete.


Rachelint (approved):
LGTM

PR #1236, opened 2023-10-01
feat: rewrite `not in` expr to `in`

jiacai2050:
## Rationale

Current filter is based on bloom-filter like structure, it can give accurate answer if an item doesn't exist in one collection, so by convert `col != value` to `col == value2`, we can fully utilize this feature.

## Detailed Changes
- Save column value to SST metadata when distinct value is less than 20.
- Rewrite not-like expr to its opposite.

## Test Plan
New UT.

### DIFF
--- a/analytic_engine/src/sst/parquet/row_group_pruner.rs
+++ b/analytic_engine/src/sst/parquet/row_group_pruner.rs
@@ -67,7 +67,7 @@ pub struct RowGroupPruner<'a> {
     schema: &'a SchemaRef,
     row_groups: &'a [RowGroupMetaData],
     parquet_filter: Option<&'a ParquetFilter>,
-    predicates: Cow<'a, Vec<Expr>>,
+    predicates: Cow<'a, [Expr]>,
     metrics: Metrics,
 }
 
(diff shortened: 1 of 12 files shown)


analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -203,6 +225,38 @@ impl RecordBatchGroupWriter {
...
+            }
+
+            let col_values = match col_values {
+                None => continue,
+                Some(v) => v,
+            };
+            let rows_num = record_batch.num_rows();
+            let column_block = record_batch.column(col_idx);
+            for row_idx in 0..rows_num {
+                match col_values {
```
chunshao90, line 248: Check KEEP_COLUMN_VALUE_THRESHOLD in this loop.

analytic_engine/src/sst/parquet/row_group_pruner.rs
```diff
@@ -212,4 +354,54 @@ mod tests {
...
+            (
+                // ip != 127.0.0.1
+                col("ip").not_eq(lit("127.0.0.1")),
+                col("ip").not_eq(lit("127.0.0.1")),
+            ),
+            (
+                // ip = 127.0.0.1
+                col("ip").eq(lit("127.0.0.1")),
+                col("ip").eq(lit("127.0.0.1")),
+            ),
```
chunshao90, line 401: Add a test, col is `ip2`.

jiacai2050: fixed

chunshao90 (approved):
LGTM

PR #1238, opened 2023-10-02
fix: throw error when create a table with a different table id

ShiKaiWi:
## Rationale
Currently, the existing table will be returned if create table with an expected table name but a different table id.

## Detailed Changes
Throws an error if the provided table id is different.

## Test Plan
This is fairly simple which doesn't need a test for it.

### DIFF
--- a/catalog/src/schema.rs
+++ b/catalog/src/schema.rs
@@ -157,14 +157,12 @@ pub enum Error {
     },
 
     #[snafu(display(
-        "Invalid table id, msg:{}, table_id:{}.\nBacktrace:\n{}",
-        msg,
-        table_id,
-        backtrace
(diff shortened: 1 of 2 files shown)


catalog/src/schema.rs
```diff
@@ -157,14 +157,12 @@ pub enum Error {
...
-        table_id,
-        backtrace
+        "Different table ids are used for the same table, table_name:{table_name}, expect_table_id:{expect_table_id}, given_table_id:{given_table_id}.\nBacktrace:\n{backtrace}",
     ))]
-    InvalidTableId {
-        msg: &'static str,
-        table_id: TableId,
+    DifferentTableId {
+        table_name: String,
+        expect_table_id: TableId,
```
jiacai2050, line 164: expected_table_id

jiacai2050 (approved):
LGTM

PR #1249, opened 2023-10-09
refactor: use `notifier::RequestNotifiers` instead of `dedup_requests::RequestNotifiers`

baojinri:
## Rationale
Now, there are same code of RequestNotifiers in notifier and dedup_requests, so we can use `notifier::RequestNotifiers` instead of `dedup_requests::RequestNotifiers`

## Detailed Changes
- use `notifier::RequestNotifiers` instead of `dedup_requests::RequestNotifiers`  
- remove `dedup_requests` file

## Test Plan
Existing tests


server/src/server.rs
```diff
@@ -458,7 +452,10 @@ impl Builder {
             engine_runtimes.clone(),
             self.cluster.is_some(),
             self.server_config.sub_table_access_perm,
-            request_notifiers,
+            self.server_config
+                .query_dedup
+                .enable
+                .then(|| Arc::new(RequestNotifiers::default())),
```
ShiKaiWi, line 458: Declare the `request_notifiers` above to keep the `Proxy::new()` simple.

ShiKaiWi (approved):

PR #1251, opened 2023-10-09
feat: support re-acquire shard lock in a fast way

ShiKaiWi:
## Rationale
Currently, a restarted ceresdb-server must wait for the previous lock lease to expire before acquiring the new shard locks, leading to long failover time.

## Detailed Changes
Support fast reacquire the shard lock by reusing the previous lock lease.

## Test Plan
Manually test this feature.

### DIFF
--- a/cluster/src/shard_lock_manager.rs
+++ b/cluster/src/shard_lock_manager.rs
@@ -428,6 +428,9 @@ impl ShardLock {
             .context(GetLockKey {
                 shard_id: self.shard_id,
             })?;
+
+        // Only one or zero key-value will be fetched, and it can continue only if one
+        // key-value is returned.
         if resp.kvs().len() != 1 {
             warn!(
                 "Expect exactly one key value pair, but found {} kv pairs, shard_id:{}",
@@ -436,14 +439,15 @@ impl ShardLock {
             );
             return Ok(None);
         }
-        let lease_id = resp.kvs()[0].lease();
+        let kv = &resp.kvs()[0];
+        let lease_id = kv.lease();
         if lease_id == 0 {
             // There is no lease attached to the lock key.
             return Ok(None);
         }
 
         // FIXME: A better way is to compare the specific field of the decoded values.
-        if resp.kvs()[0].value() != self.value {
+        if kv.value() != self.value {
             warn!(
                 "Try to acquire a lock held by others, shard_id:{}",
                 self.shard_id
--- a/cluster/src/shard_lock_manager.rs
+++ b/cluster/src/shard_lock_manager.rs
@@ -68,6 +68,25 @@ pub enum Error {
     },
 
     #[snafu(display(
+        "Failed to get lease, shard_id:{shard_id}, lease_id:{lease_id}, err:{source}.\nBacktrace:\n{backtrace:?}"
+    ))]
+    GetLease {
+        shard_id: ShardId,
+        lease_id: i64,
+        source: etcd_client::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display(
+        "Failed to get lock key, shard_id:{shard_id}, err:{source}.\nBacktrace:\n{backtrace:?}"
+    ))]
+    GetLockKey {
+        shard_id: ShardId,
+        source: etcd_client::Error,
+        backtrace: Backtrace,
+    },
+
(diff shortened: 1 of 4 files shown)


cluster/src/shard_lock_manager.rs
```diff
@@ -380,6 +411,93 @@ impl ShardLock {
...
+            return Ok(None);
+        }
+
+        let resp = etcd_client
+            .get(self.key.clone(), None)
+            .await
+            .context(GetLockKey {
+                shard_id: self.shard_id,
+            })?;
+        if resp.kvs().len() != 1 {
```
Rachelint, line 431: Maybe it should print an error log?

cluster/src/shard_lock_manager.rs
```diff
@@ -380,6 +411,93 @@ impl ShardLock {
...
+            warn!(
+                "Expect exactly one key value pair, but found {} kv pairs, shard_id:{}",
+                resp.kvs().len(),
+                self.shard_id
+            );
+            return Ok(None);
+        }
+        let lease_id = resp.kvs()[0].lease();
```
Rachelint, line 439: Should add a judgement about if `kvs` is empty?

cluster/src/shard_lock_manager.rs
```diff
@@ -380,6 +411,93 @@ impl ShardLock {
...
+        if lease_id == 0 {
+            // There is no lease attached to the lock key.
+            return Ok(None);
+        }
+
+        // FIXME: A better way is to compare the specific field of the decoded values.
+        if resp.kvs()[0].value() != self.value {
```
Rachelint, line 446: `let kvs = resp.kvs[0]` and reuse?

cluster/src/shard_lock_manager.rs
```diff
@@ -400,28 +518,30 @@ impl ShardLock {
...
-                shard_id: self.shard_id,
-            })?;
-        ensure!(
-            resp.ttl() > 0,
-            GrantLeaseWithInvalidTTL {
-                shard_id: self.shard_id,
-                ttl_sec: resp.ttl()
+        let lease_info = match self.maybe_fast_acquire_lock(etcd_client).await {
+            Ok(Some(v)) => {
+                info!("Shard lock is acquired fast, shard_id:{}", self.shard_id);
```
Rachelint, line 523: I think maybe a concrete prefix for log in a specific module is useful?
When debugging, I found we always need to grep a series of logs in a module rather than just a single log, and that can be convenient by adding prefix?

ShiKaiWi: I guess the file name in the log can meet your requirements, and manual prefix is verbose.

cluster/src/shard_lock_manager.rs
ShiKaiWi: For most cases, the `len()` will be 0 or 1, and the two values are correct both.

cluster/src/shard_lock_manager.rs
ShiKaiWi: I guess the L431 has ensured it must contain exactly one element in the list.

cluster/src/shard_lock_manager.rs
```diff
@@ -380,6 +411,93 @@ impl ShardLock {
         Bytes::from(key)
     }
 
+    /// The shard lock is allowed to acquired in a fast way if the lock has been
+    /// created by this node just before.
+    ///
+    /// The slow way has to wait for the previous lock lease expired. And the
+    /// fast way is to reuse the lease if it is not expired and the lock's value
+    /// is the same.
+    async fn maybe_fast_acquire_lock(&self, etcd_client: &mut Client) -> Result<Option<LeaseInfo>> {
```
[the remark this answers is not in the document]
ShiKaiWi: `fast_acquire_lock` corresponds to `slow_acquire_lock`, and reusing lock is the way to be fast, which has been explained in the comments.

cluster/src/shard_lock_manager.rs
Rachelint: Hoa about adding a comment to analyze this? I think it an error when reading just code here...

jiacai2050 (approved):
LGTM

jiacai2050 (approved):

PR #1252, opened 2023-10-10
feat: support simple late materialization and page filter

### LINKED ISSUE #1242: Late materialization for paquet reading
### Describe This Problem

We found one of the cpu bottlenecks for query is parquet's decoding in production, and `late materialization` is the effective method for optimizing this.

### Proposal

Parquet's late materialization impl is too naive now, we must be very careful when using it, so as I see filters can be pushed down for late materialization should satisfy following conditions now:
- Should contain just a single column (just need to pull one column for eval).
- Should be selective enough(such as `=`, `in`).
- We should sort the filters according to the encoded columns size in them.
However, it is too tired for us users to use this feature, maybe we need to help to improve this in parquet.

- [x] Impl first version late materialization following design above.
- [x] Define metrics to measure the effect of late materialization.
- [x] Improve late materialization impl in parquet.

### Additional Context

_No response_

Rachelint:
## Rationale
Closes #1242 

## Detailed Changes
+ Support simple late materialization.
+ Support dynamically enable/disable `late materialization`.
+ Add metrics for observing effect of `late materialization`.
+ Re-enable page filter and ignore errors.

## Test Plan
Test by exist tests and manually.

### DIFF
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -125,11 +125,20 @@ pub struct Config {
     pub metrics: MetricsOptions,
 }
 
-#[derive(Debug, Default, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(default)]
 pub struct ParquetReaderOptions {
     enable_page_filter: bool,
     enable_lazy_row_filter: bool,
+}
+
+impl Default for ParquetReaderOptions {
+    fn default() -> Self {
+        Self {
+            enable_page_filter: true,
+            enable_lazy_row_filter: true,
+        }
+    }
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
--- a/analytic_engine/src/lib.rs
+++ b/analytic_engine/src/lib.rs
@@ -36,12 +36,15 @@ pub mod table_meta_set_impl;
 #[cfg(any(test, feature = "test"))]
 pub mod tests;
 
+use std::sync::{atomic::AtomicBool, Arc};
+
 use manifest::details::Options as ManifestOptions;
 use object_store::config::StorageOptions;
 use serde::{Deserialize, Serialize};
 use size_ext::ReadableSize;
 use wal::config::StorageConfig;
 
+use crate::sst::DynamicConfig as SstDynamicConfig;
 pub use crate::{compaction::scheduler::SchedulerConfig, table_options::TableOptions};
 
 /// Config of analytic engine
@@ -90,6 +93,8 @@ pub struct Config {
     pub scan_max_record_batches_in_flight: usize,
     /// Sst background reading parallelism
     pub sst_background_read_parallelism: usize,
+    /// Reader options
+    pub sst_reader_options: ReaderOptions,
     /// Number of streams to prefetch
     pub num_streams_to_prefetch: usize,
     /// Max buffer size for writing sst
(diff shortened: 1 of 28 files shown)


ShiKaiWi:
@Rachelint Some conflicts need resolving.

analytic_engine/src/lib.rs
```diff
@@ -90,6 +93,8 @@ pub struct Config {
     pub scan_max_record_batches_in_flight: usize,
     /// Sst background reading parallelism
     pub sst_background_read_parallelism: usize,
+    /// Reader options
+    pub sst_reader_options: ReaderOptions,
```
ShiKaiWi, line 97: Shall keep the same style with the field `table_opts`?

analytic_engine/src/lib.rs
```diff
@@ -120,6 +125,28 @@ pub struct Config {
...
+        Self {
+            enable_page_filter: true,
+            enable_lazy_row_filter: true,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(tag = "type")]
+pub enum ReaderOptions {
```
ShiKaiWi, line 146: I guess it should not use enumeration here, because in the future multiple readers may be used all. And I suggest it to be:
```rust
pub struct SstReaderOptions {
    pub enable_parquet_page_filter: bool,
    pub enable_parquet_row_filter: bool,
}
```

analytic_engine/src/lib.rs
```diff
@@ -166,3 +194,20 @@ impl Default for Config {
         }
     }
 }
+
+/// Analytic dynamic config
+#[derive(Debug, Default, Clone)]
+pub struct DynamicConfig {
```
ShiKaiWi, line 200: I guess this is not a good way to provide the feature for dynamic configurations because the details of the dynamic config has been exposed and any changes of it will lead to the caller's code update. So how about exposing a trait or struct called `DynamicConfigger` to provide specific method to do config dynamic change?

analytic_engine/src/sst/parquet/lazy_row_filter.rs
```diff
@@ -0,0 +1,445 @@
...
+                self.maybe_table_level_metrics
+                    .record_batch_before_filter_counter
+                    .inc_by(bool_arr.len() as u64);
+                self.maybe_table_level_metrics
+                    .record_batch_after_filter_counter
+                    .inc_by(bool_arr.true_count() as u64);
+                Ok(bool_arr)
+            }
+            Err(e) => Err(arrow::error::ArrowError::ComputeError(format!(
+                "Error evaluating filter predicate: {:?}",
```
ShiKaiWi, line 327:
```suggestion
                "Error evaluating filter predicate, err:{:?}",
```

analytic_engine/src/sst/parquet/lazy_row_filter.rs
```diff
@@ -0,0 +1,445 @@
...
+            expr,
+            schema,
+            stats: FilterStats::new(),
+        }
+    }
+
+    fn build(mut self) -> Result<FilterStats> {
+        self.expr.visit(&mut self).context(DataFusionError)?;
+
+        self.stats.highly_selective = match self.expr {
```
ShiKaiWi, line 239: Maybe whether a filter is `highly_selective` depends on not only the filter form but also the number of the unique values of the filtered column. If the column has only a few values, the filters applied to it may be badly selective.

And is there any information describing the UV of the columns in the metadata?

PR #1254, opened 2023-10-10
feat: support access etcd with tls

ShiKaiWi:
## Rationale
For the most etcd clusters on the production environment, the tls is enabled, but currently ceresdb-server can't support access etcd cluster via tls.

## Detailed Changes
Support access etcd cluster by tls configgurations.

## Test Plan
Muanul tests.


jiacai2050 (approved):
LGTM

PR #1256, opened 2023-10-10
feat: support schema validate in remote write

ShiKaiWi:
## Rationale
The schema must be ensured to be correct during the remote write rpc call. Howerver, such check is missing now.

## Detailed Changes
Introduce schema version check during remote write rpc call.

## Test Plan
Add a new integration test for it.

### DIFF
--- a/table_engine/src/remote/model.rs
+++ b/table_engine/src/remote/model.rs
@@ -280,6 +280,7 @@ fn validate_contiguous_payload_schema(schema: &Schema, column_descs: &[ColumnDes
             }
         );
     }
+
     Ok(())
 }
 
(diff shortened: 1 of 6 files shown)


table_engine/src/remote/model.rs
```diff
@@ -242,6 +259,30 @@ impl WriteRequest {
...
+
+    for (expect_column_schema, desc) in schema.columns().iter().zip(column_descs.iter()) {
+        ensure!(
+            expect_column_schema.is_correct_desc(desc),
+            SchemaMismatch {
+                msg: format!("expect column schema:{expect_column_schema:?}, but got {desc:?}")
+            }
+        );
+    }
+    Ok(())
```
jiacai2050, line 283: Add a newline before.

PR #1257, opened 2023-10-11
feat: avoid flush when drop table

jiacai2050:
## Rationale
When drop a table, flush is unnecessary, and this cause the drop take a long time.

## Detailed Changes
Mark WAL as deleted directly when drop table.

## Test Plan
Manually

### DIFF
--- a/analytic_engine/src/instance/drop.rs
+++ b/analytic_engine/src/instance/drop.rs
@@ -21,7 +21,7 @@ use table_engine::engine::DropTableRequest;
 
 use crate::{
     instance::{
-        engine::{Result, WriteManifest},
+        engine::{PurgeWal, Result, WriteManifest},
         SpaceStoreRef,
     },
     manifest::meta_edit::{DropTableMeta, MetaEdit, MetaEditRequest, MetaUpdate},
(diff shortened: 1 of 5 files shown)


analytic_engine/src/instance/drop.rs
```diff
@@ -58,19 +54,21 @@ impl Dropper {
...
+        // when table_data is dropped.
+        let table_location = table_data.table_location();
+        let wal_location =
+            crate::instance::create_wal_location(table_location.id, table_location.shard_info);
+        // Use max to represent delete all WAL.
+        // TODO: add a method in wal_manager to delete all WAL with same prefix.
+        let sequence = MAX_SEQUENCE_NUMBER;
+        self.space_store
+            .wal_manager
+            .mark_delete_entries_up_to(wal_location, sequence)
```
ShiKaiWi, line 67: Maybe we should mark wal deleted after the table is dropped from the manifest.

wal/src/table_kv_impl/table_unit.rs
```diff
@@ -996,13 +996,6 @@ impl TableUnitWriter {
             table_unit_state.table_id, sequence_num, table_unit_meta_table
         );
 
-        ensure!(
```
ShiKaiWi, line 999: Why remove this check?

jiacai2050: - This check is useless here, sequence could only advance in put. 
- We use MAX_SEQUENCE_NUMBER to represent delete all.

Rachelint (approved):
LGTM

analytic_engine/src/instance/drop.rs
jiacai2050: Added in comments.

PR #1259, opened 2023-10-12
chore: delete the configuration related to github cache

tanruixiang:
## Rationale
Due to the huge cache of rust before, unit tests often failed due to insufficient capacity. [CI without cache](https://github.com/CeresDB/ceresdb/actions/runs/6492219953/job/17630769103?pr=1259) takes an acceptable time and does not cause failure problems.

https://github.com/CeresDB/ceresdb/pull/1261 show CI always fail in Rust Cache.
The github action's time seems have huge volatile, https://github.com/CeresDB/ceresdb/actions/runs/6517181839/job/17701414760?pr=1259 takes 24 min without cache, some pr have cache even is slower than this action
## Detailed Changes
Delete out the configuration related to github cache

## Test Plan
CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,10 @@ jobs:
       - run: |
           rustup set auto-self-update disable
           rustup toolchain install ${RUST_VERSION} --profile minimal
+      - name: Release Disk Quota
+        run: |
+          sudo rm -rf /usr/local/lib/android # release about 10 GB
+          sudo rm -rf /usr/share/dotnet # release about 20GB
       - name: Check License Header
         uses: korandoru/hawkeye@v3
       - name: Setup Build Environment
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -56,26 +56,8 @@ jobs:
       - run: |
           rustup set auto-self-update disable
           rustup toolchain install ${RUST_VERSION} --profile minimal
-      - name: Release Disk Quota
```
ShiKaiWi, line 59: The disk may be not enough to run the ci, so how about keeping this step?

tanruixiang: Solved

jiacai2050 (approved):
LGTM

jiacai2050:
I plan to merge to main and see if it works well in daily dev.

PR #1264, opened 2023-10-17
chore: add metrics to inspect write path

Rachelint:
## Rationale
Write path metrics about pending queue is not enough, improve it in this pr.

## Detailed Changes
+ Add time cost for pending queue wirter/waiter.
+ Add counter for related future cacellations.

## Test Plan
Test manually.

### DIFF
--- a/analytic_engine/src/table/mod.rs
+++ b/analytic_engine/src/table/mod.rs
@@ -288,8 +288,11 @@ impl TableImpl {
 
         match queue_res {
             QueueResult::First => {
-                let _timer = self.table_data.metrics.start_table_write_queue_writer_timer();
-                
+                let _timer = self
+                    .table_data
+                    .metrics
+                    .start_table_write_queue_writer_timer();
+
                 // This is the first request in the queue, and we should
                 // take responsibilities for merging and writing the
                 // requests in the queue.
@@ -302,6 +305,7 @@ impl TableImpl {
 
                 match CancellationSafeFuture::new(
                     Self::write_requests(write_requests),
+                    "pending_queue_writer".to_string(),
                     self.instance.write_runtime().clone(),
                 )
                 .await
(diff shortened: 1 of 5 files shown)


components/future_ext/src/cancel.rs
```diff
@@ -36,6 +47,9 @@ where
     F: Future + Send + 'static,
     F::Output: Send,
 {
+    /// Token for metrics
+    token: String,
```
ShiKaiWi, line 51: Maybe we can declare the token as a generic parameter T, where `T: AsRef<str> + 'static`. And it can avoid memory allocation for some cases where only a static constant string is used.

analytic_engine/src/table/mod.rs
```diff
@@ -311,7 +317,22 @@ impl TableImpl {
...
                 // write result.
-                match rx.await {
+                let _timer = self
+                    .table_data
+                    .metrics
+                    .start_table_write_queue_waiter_timer();
+
+                // We have ever observed that `rx` is closed in production but it is impossible
+                // in theory(especially after warping actual write by
+                // `CancellationSafeFuture`)... So we also warp `rx` by
```
ShiKaiWi, line 327:
```suggestion
                // `CancellationSafeFuture`). So we also warp `rx` by
```

analytic_engine/src/table/mod.rs
```diff
@@ -300,6 +305,7 @@ impl TableImpl {
 
                 match CancellationSafeFuture::new(
                     Self::write_requests(write_requests),
+                    "pending_queue_writer".to_string(),
```
ShiKaiWi, line 308:
```suggestion
                    "pending_queue_writer",
```

analytic_engine/src/table/mod.rs
```diff
@@ -311,7 +317,22 @@ impl TableImpl {
...
+                    .metrics
+                    .start_table_write_queue_waiter_timer();
+
+                // We have ever observed that `rx` is closed in production but it is impossible
+                // in theory(especially after warping actual write by
+                // `CancellationSafeFuture`). So we also warp `rx` by
+                // `CancellationSafeFuture` for not just retrying but better observing.
+                match CancellationSafeFuture::new(
+                    rx,
+                    "pending_queue_waiter".to_string(),
```
ShiKaiWi, line 331:
```suggestion
                    "pending_queue_waiter",
```

ShiKaiWi (approved):

PR #1265, opened 2023-10-17
refactor: improve error readability

jiacai2050:
## Rationale
Currently when there are errors, sql is in the middle of err message, the most important message is in the last, when sql is too long, it's hard to find real root cause.

## Detailed Changes
- Put sql in the last part of error message
- Remove failed log, use slow_log instead.
## Test Plan
Integration tests.


ShiKaiWi (approved):
LGTM

Issue #1144, opened 2023-08-09
Infer timestamp column when possible


jiacai2050:
### Describe This Problem

When create table with sql below

```sql
CREATE TABLE `abc` (
    `tag` string TAG,
    `id` int TAG,
    `value` double NOT NULL,
    `t` timestamp NOT NULL,
    )
```

It will throw
```
MysqlWorker on_query failed. err:Failed to handle sql:CREATE TABL
E `abc` (                                                                                                                          `tag` string TAG,                                          
    `id` int TAG,                                                                                                                  `value` double NOT NULL,                                   
    `t` timestamp NOT NULL,                                                                                                        ), err:Rpc error, code:500, message:Failed to create plan, query:CREATE TABLE `abc` (                                          `tag` string TAG,                                                                                                          
    `id` int TAG,                                                                                                                  `value` double NOT NULL,                                                                                                       `t` timestamp NOT NULL,                                                                                                    
    ), err:Failed to create plan, err:Table must contain timestamp constraint    
```

### Proposal

In theory, if there is only one timestamp column in our SQL, we can refer this without user to set explicitly.

Dennis40816:
Hi, @jiacai2050. Is anyone currently working on this? If not, I'd like to give it a try.

chunshao90:
> Hi, @jiacai2050. Is anyone currently working on this? If not, I'd like to give it a try.

Thanks. Feel free to try it.

PR #1266, opened 2023-10-17
feat: infer timestamp constraint for single-timestamp column

Dennis40816:
## Rationale

This PR aims to resolve issue #1144 by simplifying the process of table creation with timestamp constraints. Specifically, the system will automatically infer and set the timestamp constraint under the following conditions:

1. No timestamp constraint has been previously set.
2. There exists exactly one column with a timestamp data type(`DataType::Timestamp`).

By introducing this automatic inference, we reduce the complexity for users who otherwise would have to manually set this constraint.

## Detailed Changes

- **Automatic Inference of Timestamp Constraint**: The system now detects if there is only a single timestamp column during table creation and automatically sets it as the timestamp constraint.
  - Added a function `build_and_check_timestamp_key_constraint` for checking and inference.
  - Modified the `create_table` function to call the new function.

- **Updated Test Expectations for `create_table`**:
  - Refactored test cases for `create_table` in `parser.rs` to align with the latest timestamp inference changes.
  - Updated the expected statements to include unique constraints for timestamp columns.

## Test Plan

Utilized the default CI Action for testing. The result was a [pass](https://github.com/Dennis40816/ceresdb/actions/runs/6545705109).

## Remarks

- Made modifications to a particular section of the code. Would appreciate it if experienced team members could review and confirm whether any changes or corrections are needed.

### DIFF
--- a/query_frontend/src/parser.rs
+++ b/query_frontend/src/parser.rs
@@ -995,34 +995,52 @@ mod tests {
         expect_parse_ok(sql, expected).unwrap();
 
         // positive case, multiple columns
+        let columns = vec![
+            make_column_def("c1", DataType::Timestamp(None, TimezoneInfo::None)),
+            make_column_def("c2", DataType::Double),
+            make_column_def("c3", DataType::String),
+        ];
+
         let sql = "CREATE TABLE mytbl(c1 timestamp, c2 double, c3 string,) ENGINE = XX";
         let expected = Statement::Create(Box::new(CreateTable {
             if_not_exists: false,
             table_name: make_table_name("mytbl"),
-            columns: vec![
-                make_column_def("c1", DataType::Timestamp(None, TimezoneInfo::None)),
-                make_column_def("c2", DataType::Double),
-                make_column_def("c3", DataType::String),
-            ],
+            columns: columns.clone(),
             engine: "XX".to_string(),
-            constraints: vec![],
+            constraints: vec![TableConstraint::Unique {
+                name: Some(Ident {
+                    value: TS_KEY.to_owned(),
+                    quote_style: None,
+                }),
+                columns: vec![columns[0].name.clone()],
+                is_primary: false,
+            }],
             options: vec![],
             partition: None,
         }));
         expect_parse_ok(sql, expected).unwrap();
 
         // positive case, multiple columns with comment
+        let columns = vec![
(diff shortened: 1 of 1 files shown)


query_frontend/src/parser.rs
```diff
@@ -791,6 +791,47 @@ fn check_column_expr_validity_in_hash(column: &Ident, columns: &[ColumnDef]) ->
     valid_column.is_some()
 }
 
+/// Builds and checks for the existence of a timestamp key constraint named
+/// __ts_key within the given list of table constraints. If such a constraint
+/// does not exist, the function searches for a unique timestamp column and
+/// creates a new constraint for it.
+fn build_and_check_timestamp_key_constraint(
```
jiacai2050, line 798: I suggest we name it `build_or_infer_timestamp_key_constraint`.

query_frontend/src/parser.rs
```diff
@@ -954,34 +995,52 @@ mod tests {
         expect_parse_ok(sql, expected).unwrap();
 
         // positive case, multiple columns
+        let columns = vec![
```
jiacai2050, line 998: Would mind add a new case with explicit `timestamp key` ?

query_frontend/src/parser.rs
Dennis40816: Thank you for the suggestion. The name `build_or_infer_timestamp_key_constraint` does seem clearer than the previous one. I'll make the change shortly.

query_frontend/src/parser.rs
Dennis40816: See 1c1706bf0807d807495b16a7531d9ddc0eafcfbb.

query_frontend/src/parser.rs
Dennis40816: See c59aed1a612c608df8e7bcec5b95c06980f69079.

Dennis40816:
Thank you for the review and for your positive feedback! 👍

I've gone ahead and made some additional updates based on our discussion:

1. Renamed the function for better clarity.
2. Added a new test case to cover more scenarios.
Your minor suggestions make complete sense to me, and I've taken them into account.

Here's the [link to the updated CI results](https://github.com/Dennis40816/ceresdb/actions/runs/6550816538).

Please let me know if there's anything else to improve.

jiacai2050 (approved):
LGTM

PR #1269, opened 2023-10-18
fix: dist query dedup

Rachelint:
## Rationale
We place some dynamic fields in plans needed to send for remote execution, that lead to failure of request dedup.
I make a quick fix in this pr by overwriting such fields before encoding, and refill them in remote.

## Detailed Changes
+ Overwrite dynamic fields of `ReadRequest` in `Resovler`.
+ Refill them in `ExecutableScanBuilderImpl`.

## Test Plan
Test by ut.


df_engine_extensions/src/dist_sql_query/mod.rs
```diff
@@ -53,8 +52,127 @@ pub trait ExecutableScanBuilder: fmt::Debug + Send + Sync + 'static {
...
+impl fmt::Debug for TableScanContext {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let exprs = self
+            .predicate
+            .exprs()
+            .iter()
+            .map(|expr| expr.to_string())
+            .collect::<Vec<_>>()
+            .join(",");
+        let predicate = format!("[{}]", exprs);
```
ShiKaiWi, line 159: `Predicate` seems derived from `Debug`, so shall we just use it directly?

Rachelint: It is not so readable use default `Debug`...
I impl a custom debug dispaly for it?

ShiKaiWi: Ok. How about use `f.debug_list`, which receives an iterator, to avoid the heap memory allocation?

jiacai2050 (approved):
LGTM

df_engine_extensions/src/dist_sql_query/mod.rs
ShiKaiWi: Looks great!

ShiKaiWi (approved):

ShiKaiWi:
@Rachelint Integration test fails.

PR #1272, opened 2023-10-19
refactor: conditionally compile wal impls 

tisonkun:
## Rationale

Supersede #1231

After compile time with clean - 

```
    Finished dev [unoptimized + debuginfo] target(s) in 2m 18s
```

## Detailed Changes

Conditionally compile wal impls with feature flags.

## Test Plan

Code refactor. All existing tests should pass.


analytic_engine/Cargo.toml
```diff
@@ -25,7 +25,9 @@ workspace = true
 workspace = true
 
 [features]
+default = ["wal-table-kv"]
```
ShiKaiWi, line 28: I think we should use rocksdb wal as default wal to ensure the first time trial to be friendly for users.

tisonkun: The origin we go here is to opt-out rocksdb wal impl. But it's OK to work as explicit instead of by default.

I'm open to other advice for what is desired here @jiacai2050.

BTW, table-kv can use memory impl as default for smoothy first time trial?

ShiKaiWi: > BTW, table-kv can use memory impl as default for smoothy first time trial?
If so, it may be confusing about that the data is lost after restarting the server.

ShiKaiWi: @Rachelint @jiacai2050 Any suggestions about the default wal implemenation?

jiacai2050: I prefer default to rocksdb implementation, as it was before. 

Memory impl only for testing and may cause confusion if users restart server and found data is lost.

ShiKaiWi: @tisonkun After some discussions, we guess this is a proper way to set the feature gates:
- The default feature should include all the WAL implementations, to avoid any possible confusion about data loss, config not working, etc.
- A specific WAL implementation can be chosen for a fast deployment. (Currently, in our prod env, there is no need to build ceresdb with rocksdb).

And in order to build ceresdb without rocksdb by default, I guess another simple WAL implementation based on file system should be provided.

tisonkun:
@jiacai2050 with `make release-slim` it gives:

```
l target/release-slim/ceresdb-server 
-rwxr-xr-x  1 tison  staff    83M Oct 20 15:01 target/release-slim/ceresdb-server
```

analytic_engine/Cargo.toml
tisonkun: OK. Let me try to modifiy this patch with the resolution above :D

tisonkun:
Updated. PTAL again @jiacai2050 @

tisonkun:
Updated. PTAL again.

BTW, golang related CI failed with a new issue:

```
server/coordinator/scheduler/static/scheduler.go:6:2: package cmp is not in GOROOT (/opt/hostedtoolcache/go/1.20.10/x64/src/cmp)
note: imported by a module that requires go 1.21
server/coordinator/scheduler/rebalanced/scheduler.go:8:2: package maps is not in GOROOT (/opt/hostedtoolcache/go/1.20.10/x64/src/maps)
note: imported by a module that requires go 1.21
server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go:38:2: package slices is not in GOROOT (/opt/hostedtoolcache/go/1.20.10/x64/src/slices)
note: imported by a module that requires go 1.21
make[1]: *** [Makefile:37: build-meta] Error 1
make[1]: Leaving directory '/home/runner/work/ceresdb/ceresdb/integration_tests'

```

tisonkun:
Perhaps we should setup go toolchain in the CI env.

tisonkun:
CI passed. PTAL @jiacai2050 @

ShiKaiWi (approved):
👍

PR #1275, opened 2023-10-20
ci: setup golang in CI

tisonkun:
We may use `go-version-file` instead to avoid manually maintain the `go-version` field.

But the approach we currently clone ceresmeta and build is a bit complex to change; so I'd prefer to postpone it to a follow-up nice-to-have task.


tisonkun:
cc @ShiKaiWi

ShiKaiWi (approved):
LGTM

PR #1277, opened 2023-10-24
chore: remove default features in analytic_engine

jiacai2050:
## Rationale
Currently although wal is gated by different features,  we can't disable a wal implementation.

## Detailed Changes
- Remove default default in analytic_engine, and let upper layer to choose which wal to use.

## Test Plan
Manually.


ShiKaiWi (approved):
LGTM

PR #1282, opened 2023-10-26
feat: skip record column values for level0 sst

jiacai2050:
## Rationale
There are not many rows in min level, so we shouldn't record values for them, as they cost too much CPU.

## Detailed Changes
- Skip record column values for level0

## Test Plan
Existing UT.

### DIFF
--- a/analytic_engine/src/sst/parquet/writer.rs
+++ b/analytic_engine/src/sst/parquet/writer.rs
@@ -111,7 +111,7 @@ impl RecordBatchGroupWriter {
         level: Level,
     ) -> Self {
         let column_values = if level.is_min() {
-            // There are not many rows in min level, so we don't record values.
+            // There are not many rows in min level, so we don't record values for them.
             None
         } else {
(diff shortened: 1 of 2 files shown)


analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -97,7 +97,7 @@ struct RecordBatchGroupWriter {
     input_exhausted: bool,
     // Time range of rows, not aligned to segment.
     real_time_range: Option<TimeRange>,
-    column_values: Vec<Option<ColumnValueSet>>,
+    column_values: Option<Vec<Option<ColumnValueSet>>>,
```
ShiKaiWi, line 100: Add some comments to explain this field.

ShiKaiWi (approved):

analytic_engine/src/sst/parquet/writer.rs
jiacai2050: Fixed

PR #1287, opened 2023-10-31
test: simplify ceresmeta-server installation

tisonkun:
This refers to https://github.com/CeresDB/ceresmeta/pull/265.

### DIFF
--- a/integration_tests/build_meta.sh
+++ b/integration_tests/build_meta.sh
@@ -7,7 +7,8 @@ set -o pipefail
 BIN_PATH=${BIN_PATH:-""}
 if [[ -z "${BIN_PATH}" ]]; then
   echo "Fetch and install ceresmeta-server..."
-  go install github.com/CeresDB/ceresmeta/cmd/ceresmeta-server@5861780
+  COMMIT=$(gh api -H "Accept: application/vnd.github.VERSION.sha" repos/CeresDB/ceresmeta/commits/main)
+  go install github.com/CeresDB/ceresmeta/cmd/ceresmeta-server@${COMMIT}
   BIN_PATH="$(go env GOPATH)/bin/ceresmeta-server"
 fi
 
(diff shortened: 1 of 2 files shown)


integration_tests/build_meta.sh
```diff
@@ -1,19 +1,14 @@
...
-  echo "Pull meta repo..."
-  git clone --depth 1 https://github.com/ceresdb/ceresmeta.git ${SRC}
+set -o errexit
+set -o nounset
+set -o pipefail
+
+BIN_PATH=${BIN_PATH:-""}
+if [[ -z "${BIN_PATH}" ]]; then
+  echo "Fetch and install ceresmeta-server..."
+  go install github.com/CeresDB/ceresmeta/cmd/ceresmeta-server@5861780
```
tisonkun, line 10: We need a formal new tag for ceresmeta to switch to `latest`.

tisonkun:
cc @ZuLiangWang @jiacai2050

integration_tests/build_meta.sh
tisonkun: What's your proposal for this ref?

I'd retain `@latest` while it will be resolved to `v1.2.7` where https://github.com/CeresDB/ceresmeta/pull/265 is not included. So I wrote "We need a formal new tag".

For local debugging, the env var `BIN_PATH` can be used to specify the local version of ceresmeta-server.

ZuLiangWang: We hope that the `ceresmeta-server` used by the integration test is always the latest code, rather than based on a certain tag. Can go install do this?

tisonkun:
> tests::read_write_test::test_space_write_buffer_size_rocks

Seems an unstable test.

tanruixiang (approved):
LGTM. This way of installing looks much better than the old way that would leave code behind. 👍

ShiKaiWi:
> > tests::read_write_test::test_space_write_buffer_size_rocks
> 
> Seems an unstable test.

Yes... This test case often fails. I'll fix it when I have some time.

PR #1289, opened 2023-11-02
fix: log third party crates

jiacai2050:
## Rationale
Currently we set tag to empty string to tell it from slow log, this works for crates within ceresdb,
but for crates outside ceresdb, it will discard log.


Rachelint (approved):
LGTM

PR #1292, opened 2023-11-03
fix: ensure primary key order

jiacai2050:
## Rationale
Current schema builder set `primary_key_index` inside `add_key_column`, which means key_indexes could only be in ascending order.

## Detailed Changes
- Remove nullable limit for pk
- Refactor `primary_key_index`, user need to set it explicitly

## Test Plan
Existing tests.

### DIFF
--- a/common_types/src/schema.rs
+++ b/common_types/src/schema.rs
@@ -88,6 +88,16 @@ pub enum Error {
     },
 
     #[snafu(display(
+        "Column id is missing in schema, id:{}.\nBacktrace:\n{}",
+        id,
+        backtrace
+    ))]
+    ColumnIdMissing { id: ColumnId, backtrace: Backtrace },
+
+    #[snafu(display("Primary key indexes cannot be empty.\nBacktrace:\n{}", backtrace))]
+    EmptyPirmaryKeyIndexes { backtrace: Backtrace },
+
+    #[snafu(display(
         "Unsupported key column type, name:{}, type:{:?}.\nBacktrace:\n{}",
         name,
         kind,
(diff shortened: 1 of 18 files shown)


common_types/src/schema.rs
```diff
@@ -958,62 +961,38 @@ impl Schema {
...
+                    .iter()
+                    .enumerate()
+                    .find(|(_, col)| &col.id == id)
+                {
+                    Ok(v.0)
+                } else {
+                    ColumnIdMissing { id: *id }.fail()
+                }
             })
             .collect::<Result<Vec<_>>>()?;
```
ShiKaiWi, line 982:
```suggestion
        let primary_key_indexes = primary_key_ids
            .iter()
            .cloned()
            .map(|id| {
                let column_idx = schema
                    .columns
                    .iter()
                    .enumerate()
                    .find_map(|(idx, col)| if col.id == id {
                        Some(idx)
                    } else {
                        None
                    })
                    .context(ColumnIdMissing { id: *id })?;
                
                Ok(column_idx)
            })
            .collect::<Result<Vec<_>>>()?;
```

common_types/src/schema.rs
```diff
@@ -958,62 +961,37 @@ impl Schema {
...
-                                given_column: col.name.clone(),
-                            }
-                            .fail();
-                        }
-
-                        timestamp_index = Some(idx);
-                    }
-                    break;
-                }
+        for column_schema_pb in schema.columns.into_iter() {
```
ShiKaiWi, line 984: I guess `into_iter()` is not needed?
```suggestion
        for column_schema_pb in schema.columns {
```

ShiKaiWi (approved):

Issue #1157, opened 2023-08-16
Enable to get schema and catalog information from table


Rachelint:
### Describe This Problem

We ofen need to know table's schema and catalog information, and we can do it just by passing from caller now...
It make the codes verbose and messy...

### Proposal

Store `schema id`,  `catalog manager` in `TableData`.

dust1:
If I want to finish this issue, is it directly assigned to myself?

Rachelint:
> If I want to finish this issue, is it directly assigned to myself?

Yes, welcome!

PR #1294, opened 2023-11-05
edit: add schema id, schema name, catalog name in TableData

dust1:
## Rationale
Closes https://github.com/CeresDB/ceresdb/issues/1157

To be able to get the catalog from the table

## Detailed Changes

I added the `schema id`, `schema name`, `catalog name` to `MetaEditRequest` for passing them to `TableData`, but because `MetaEdit` involves passing the protoc protocol, I don't save them here.

## Test Plan

pass

### DIFF
--- a/analytic_engine/src/instance/create.rs
+++ b/analytic_engine/src/instance/create.rs
@@ -90,6 +90,7 @@ impl Instance {
                 space_id: space.id,
                 table_id: request.table_id,
                 table_name: request.params.table_name.clone(),
+                catalog_name: request.params.catalog_name.clone(),
                 schema: request.params.table_schema,
                 opts: table_opts,
             });
--- a/analytic_engine/src/instance/create.rs
+++ b/analytic_engine/src/instance/create.rs
@@ -90,7 +90,6 @@ impl Instance {
                 space_id: space.id,
                 table_id: request.table_id,
                 table_name: request.params.table_name.clone(),
-                catalog_name: request.params.catalog_name.clone(),
                 schema: request.params.table_schema,
                 opts: table_opts,
             });
--- a/analytic_engine/src/instance/create.rs
+++ b/analytic_engine/src/instance/create.rs
@@ -96,7 +96,7 @@ impl Instance {
             MetaEditRequest {
                 shard_info: TableShardInfo::new(request.shard_id),
                 meta_edit: MetaEdit::Update(meta_update),
-                extr_info: TableCatalogInfo {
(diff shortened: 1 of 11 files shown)


Rachelint:
Actually, `schema` and `table schema` are not the same things although with a same name... This confused me at the beginning, too.`Schema` can be seemed as something like `tenant`, `database`. We don't have detailed design about `How to create schema` and `Where to store the schema` now...

This is my current idea for the issue:
+ Now `schema id` is sent to `datanode` through `CreateTableRequest` and `OpenTableRequest` , we just use it and don't need to care persistence of it currently.
+ Just simply place it into `LoadRequest` and `MetaEditRequest`, and then place it into `TableData`.

dust1:
Ok, let me revise it

Rachelint:
I see this pr has conflicts with `main` branch now, let me solve it?

dust1:
Just a moment, please. I'll sort it out😂

Rachelint:
See codes, and I found `schema_name` and `catalog_name` may be needed, too currently for the final target about expose `schema_name` and `catalog_name` through `Table` interface...
Can please help to add thems with `schema_id` together in this pr...?

dust1:
sure!

analytic_engine/src/instance/create.rs
```diff
@@ -32,7 +32,7 @@ use crate::{
     },
     manifest::meta_edit::{AddTableMeta, MetaEdit, MetaEditRequest, MetaUpdate},
     space::SpaceRef,
-    table::data::{TableDataRef, TableShardInfo},
+    table::data::{TableDataExtraneousInfo, TableDataRef, TableShardInfo},
```
Rachelint, line 35: how about naming it TableCatalogInfo?

dust1: no problem!

analytic_engine/src/instance/create.rs
```diff
@@ -96,6 +96,11 @@ impl Instance {
             MetaEditRequest {
                 shard_info: TableShardInfo::new(request.shard_id),
                 meta_edit: MetaEdit::Update(meta_update),
+                extr_info: TableCatalogInfo {
```
Rachelint, line 99: How about refactor field names together, and we merge it after that!

dust1: ok!

Rachelint (approved):
LGTM

Rachelint:
Thanks @dust1 .

PR #1297, opened 2023-11-06
refactor: remove unused min/max timestamp in the RowGroup

ShiKaiWi:
## Rationale
The min/max timestamp in `RowGroup` is not used any more.

## Detailed Changes
- Remove the min/max timestamp from the `RowGroup`
- Remove the `RowGroupBuilder` because it's too simple without min/max timestamp updating

## Test Plan
All the tests in the CI should pass.

### DIFF
--- a/common_types/src/row/mod.rs
+++ b/common_types/src/row/mod.rs
@@ -268,7 +268,7 @@ impl RowGroup {
     /// [None] will be thrown if the rows have different schema from the
     /// provided one.
     #[inline]
-    pub fn new_checked(schema: Schema, rows: Vec<Row>) -> Result<Self> {
+    pub fn try_new(schema: Schema, rows: Vec<Row>) -> Result<Self> {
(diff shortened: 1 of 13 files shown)


common_types/src/row/mod.rs
```diff
@@ -252,14 +251,30 @@ pub struct RowGroup {
...
+    pub fn new_unchecked(schema: Schema, rows: Vec<Row>) -> Self {
+        Self { schema, rows }
+    }
+
+    /// Check and create row group.
+    ///
+    /// [None] will be thrown if the rows have different schema from the
+    /// provided one.
+    #[inline]
+    pub fn new_checked(schema: Schema, rows: Vec<Row>) -> Result<Self> {
```
jiacai2050, line 271: I think this method should be called `try_new`.

jiacai2050 (approved):
LGTM

PR #1298, opened 2023-11-06
fix: use flag in preflush to indicate whether reorder is required

jiacai2050:
## Rationale
Current we suggest new primary keys in preflush, and do a reorder in flush task.

There is no easy way for flush task to know if reorder if required.

## Detailed Changes
- Introduce a flag to represent whether reorder is required
- Change `col` to `ident` when build logical plan, `ident` will not normalize column name.

## Test Plan
Integration test.

### DIFF
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -197,8 +197,8 @@ pub struct TableFlushRequest {
     /// Max sequence number to flush (inclusive).
     pub max_sequence: SequenceNumber,
 
-    /// We may suggest new primary keys in preflush. if suggest happens, we need
-    /// to ensure data is in new order.
+    /// We may suggest new primary keys in preflush. if suggestion happened, we
+    /// need to ensure data is in new order.
     need_reorder: bool,
 }
 
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -196,6 +196,10 @@ pub struct TableFlushRequest {
     pub table_data: TableDataRef,
     /// Max sequence number to flush (inclusive).
     pub max_sequence: SequenceNumber,
+
+    /// We may suggest new primary keys in preflush. if suggest happens, we need
+    /// to ensure data is in new order.
+    need_reorder: bool,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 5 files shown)


analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -196,6 +196,10 @@ pub struct TableFlushRequest {
     pub table_data: TableDataRef,
     /// Max sequence number to flush (inclusive).
     pub max_sequence: SequenceNumber,
+
+    /// We may suggest new primary keys in preflush. if suggest happens, we need
```
Rachelint, line 200: If suggestion happened

analytic_engine/src/sampler.rs
```diff
@@ -292,7 +292,7 @@ impl PrimaryKeySampler {
             .iter()
             .enumerate()
             .map(|(idx, col)| {
-                if idx == timestamp_index {
+                if col.data_type.is_timestamp() {
```
Rachelint, line 295: It is possible that, `Timestamp` typed column but not a `time key` is suggested to be one of the order keys?

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -589,11 +603,13 @@ impl FlushTask {
...
         let timestamp_idx = self.table_data.schema().timestamp_index();
-        if let Some(pk_idx) = self.table_data.current_version().suggest_primary_key() {
+        if need_reorder {
+            let schema = self.table_data.schema();
+            let primary_key_indexes = schema.primary_key_indexes();
             let reorder = Reorder {
                 iter,
                 schema: self.table_data.schema(),
-                order_by_col_indexes: pk_idx,
+                order_by_col_indexes: primary_key_indexes.to_vec(),
```
Rachelint, line 612: indices

jiacai2050: both indexes and indices are valid plural for index, and most of our code base use indexes.
- https://www.merriam-webster.com/dictionary/index

analytic_engine/src/sampler.rs
jiacai2050: It make no sense for me, I think we should be strict at first.

Rachelint (approved):
LGTM

PR #1304, opened 2023-11-09
fix: alter partition table tag column

chunshao90:
## Rationale
Fix adding column in `partition table` bug. 

## Detailed Changes
1. Checking the schema inconsistency error in the write process, caused by adding columns. 
    - Note: This only works for grpc write
3. Add a Go SDK test to check adding a column in a "partition table".
4. Start ceresdb cluster to run go-sdk tests.

## Test Plan
GO SDK test.

### DIFF
--- a/integration_tests/sdk/go/alteraddcolumn.go
+++ b/integration_tests/sdk/go/alteraddcolumn.go
@@ -1,0 +1,172 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/CeresDB/ceresdb-client-go/ceresdb"
+)
+
+func checkAddColumn(ctx context.Context, client ceresdb.Client) error {
+	err := dropTable(ctx, client, partitionTable)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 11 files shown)


proxy/src/write.rs
```diff
@@ -527,7 +527,8 @@ impl Proxy {
                 Err(e) => {
                     // TODO: remove this logic.
                     // Refer to https://github.com/CeresDB/ceresdb/issues/1248.
-                    if e.error_message().contains("No field named") {
+                    if e.error_message().contains("decode row group payload") {
```
jiacai2050, line 530: Define a method `need_evict_table() bool` to include all this error message required to evict table.

integration_tests/sdk/go/alteraddcolumn.go
```diff
@@ -0,0 +1,193 @@
...
+
+	_, err = ddl(ctx, client, partitionTable, fmt.Sprintf("ALTER TABLE `%s` ADD COLUMN (%s string);", partitionTable, fieldName))
+	if err != nil {
+		return err
+	}
+
+	ts := currentMS()
+
+	// First write will fail, because the schema is not updated yet.
+	// Currently, write failed will update the schema.
```
jiacai2050, line 40:
```suggestion
	// Currently, we will update the schema when write failed .
```

jiacai2050 (approved):
LGTM

PR #1305, opened 2023-11-10
feat: support stack size of read threads configurable

ShiKaiWi:
## Rationale
Complex sqls, e.g. sql with massive condition in the where clause, may lead to stack overflow during query procedure. And currently, there is no a good to handle it in the query engine powered by datafusion so a simple way is just to enlarge the stack for the read threads.

## Detailed Changes
- Support the stack size of read threads confiugrable
- Make the columnar format as the default wal payload

## Test Plan
All the tests in the ci should pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1286,6 +1286,7 @@ dependencies = [
  "serde",
  "server",
  "signal-hook",
+ "size_ext",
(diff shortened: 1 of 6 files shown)


Rachelint (approved):
LGTM

PR #1311, opened 2023-11-17
feat: impl DoNothing wal

jiacai2050:
## Rationale
In some deployments when latest data is allowed to lost, wal is not a required module.

Also, wal may cost too much resources if write throughput is high, by disable it, we can save those resources.

## Detailed Changes
- Add a `DoNothing` wal implementation
- Introduce `WalConfig`, and add a new `disable_data` flag in it.
  - This `Config` wrap old `StorageConfig`,  and use `#[serde(flatten)]` to keep backwards compatible.

## Test Plan
Manually

### DIFF
--- a/src/wal/src/config.rs
+++ b/src/wal/src/config.rs
@@ -44,9 +44,39 @@ pub enum StorageConfig {
 impl StorageConfig {
     pub fn data_disabled(&self) -> bool {
         match self {
-            Self::RocksDB(c) => c.disable_data,
-            Self::Obkv(c) => c.disable_data,
-            Self::Kafka(c) => c.disable_data,
+            Self::RocksDB(c) => {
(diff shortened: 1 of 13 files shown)


src/wal/src/config.rs
```diff
@@ -40,3 +40,43 @@ pub enum StorageConfig {
     Obkv(Box<ObkvStorageConfig>),
     Kafka(Box<KafkaStorageConfig>),
 }
+
+impl StorageConfig {
```
Rachelint, line 44: Can we extract it rather than placing it in each detailed impl config?

Rachelint (approved):
LGTM

src/wal/src/config.rs
jiacai2050: fixed

PR #1313, opened 2023-11-17
feat: support split wal logs in table kv

ShiKaiWi:
## Rationale
The wal based on table kv requires the log's size must be in valid range, but current log split is done by split the write request in the analytic engine, which is above the wal, and this is unnatural because the real size of the logs may still exceeds the limit even if the splitted write request don't.

## Detailed Changes
- Support log split in the table kv implementation
- Remove the request split in the analytic write procedure

## Test Plan
Should pass all the ci, and the compatiblity test should be taken manually.


jiacai2050:
@ShiKaiWi I guess we don't need this feature any more?

PR #1314, opened 2023-11-21
chore: ignore seq check for DoNothing wal

jiacai2050:
## Rationale
Currently DoNothing wal will incur too many logs 

## Test Plan
Manually


Rachelint (approved):
LGTM

PR #1134, opened 2023-08-05
chore: add apache license checker

tanruixiang:
## Rationale
see title

## Detailed Changes
Remove old license script and use `hawkeye` to check license in CI.

PR #1322, opened 2023-11-29
fix: the license year was changed incorrectly due to pr #1134

tanruixiang:
## Rationale
Due to a previous [pr](https://github.com/CeresDB/horaedb/pull/1134), the year of license was changed incorrectly from 2022 to 2023, which will be corrected back to 2022 here.

## Detailed Changes
change license `2023` back to `2022`

## Test Plan
ut.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,4 +1,4 @@
-# Copyright 2023 The HoraeDB Authors
+# Copyright 2022 The HoraeDB Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-# Copyright 2023 The HoraeDB Authors
+# Copyright 2022 The HoraeDB Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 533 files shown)


jiacai2050:
I check other popular apache projects, I find there is no year in their file header, maybe we can follow this?

- https://github.com/apache/echarts/blob/master/src/echarts.all.ts
- https://github.com/apache/arrow-datafusion/blob/main/rustfmt.toml

tisonkun:
@jiacai2050 that's what happens when we transfer to the ASF domain.

See https://github.com/apache/skywalking/issues/10223#issuecomment-1369393730 for more details.

jiacai2050:
Aha, it's tedious to include year in each file.

@tanruixiang Please remove year section in file header, skywalking also follow this style.
- https://github.com/apache/skywalking/blob/master/.asf.yaml

tanruixiang:
@tisonkun Does HawkEye's format support configurations without properties? when I remove the properties it comes up `// Copyright ${inceptionYear} ${copyrightOwner}`

tanruixiang:
> Aha, it's tedious to include year in each file.
> 
> @tanruixiang Please remove year section in file header, skywalking also follow this style.
> 
> * https://github.com/apache/skywalking/blob/master/.asf.yaml

This should only be done once we're in the ASF, as its protocols include `Licensed to the Apache Software Foundation (ASF)`.Do we need to remove the year's information now? @jiacai2050 @tisonkun

tisonkun:
> Do we need to remove the year's information now

No.

tanruixiang:
No fixes will be made at this stage

tisonkun:
> Does HawkEye's format support configurations without properties

Yes. You can even use `inlineHeader ` and configure any header template with TOML's multiline string.

PR #1330, opened 2023-12-02
chore: rename binary to horaedb-server and more

tisonkun:
This refers to https://github.com/CeresDB/horaedb/issues/1319.


tanruixiang (approved):
LGTM

jiacai2050 (approved):
LGTM

jiacai2050:
Since this PR update default schema and etcd root path,  we need votes from others before merge. @ShiKaiWi

jiacai2050:
We have only one catalog in our codebase now, so change its value is not a breaking changes.

The only breaking change is default etcd root path, root path itself support config, but we change its default value here, so for migration, we need add following lines to config to make it backward-compatible, which is a small changes for me.
```
[cluster_deployment.etcd_client]
server_addrs = ['127.0.0.1:2379']
root_path = "/ceresdb"
```

tisonkun:
> we change its default value here, so for migration, we need add following lines to config to make it backward-compatible

+1 "redundant" confg entries is the way to migration.

PR #1336, opened 2023-12-04
chore(license): rename `license-header.txt`'s CeresDB to HoraeDB

caicancai:
## Rationale
Related  https://github.com/CeresDB/horaedb/issues/1319

## Detailed Changes
update `license-header.txt`'s org

## Test Plan
UT.

### DIFF
--- a/scripts/license-header.txt
+++ b/scripts/license-header.txt
@@ -1,4 +1,4 @@
-Copyright 2023 The CeresDB Authors
+Copyright 2023 The HoraeDB Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


scripts/license-header.txt
```diff
@@ -10,4 +10,4 @@ Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
-limitations under the License.
\ No newline at end of file
+limitations under the License.
```
tanruixiang, line 13: That's weired, it could be a github display error.

tanruixiang (approved):
LGTM. thanks for your contribution.

tanruixiang:
> Do we still use this file? I wonder if we can just remove it.

This is supposed to be used with a git hook, which some people may think use docker  HawkEye is too heavy.
https://github.com/CeresDB/horaedb/blob/034ac75444c90e9c4317e0a032b9ee9636bd3a77/.pre-commit-config.yaml#L14

tisonkun:
If we can use precommit to format license header, maybe we don't need HawkEye but just keep one tool. However, it may not desire time to investigate XD

tanruixiang:
> If we can use precommit to format license header, maybe we don't need HawkEye but just keep one tool. However, it may not desire time to investigate XD

Yes, we can check in CI using the same way as precommit.🤣

tisonkun:
Or use shell command to replace precommit hooks since I don't find how the hooks can configure ..

tanruixiang:
> Or use shell command in precommit hooks to replace the `insert-license` hook since I don't find how the hooks can configure ..

create a [issue](https://github.com/CeresDB/horaedb/issues/1339). This hook's repo is https://github.com/Lucas-C/pre-commit-hooks, maybe you mean https://github.com/Lucas-C/pre-commit-hooks#handy-shell-functions?

PR #1344, opened 2023-12-05
chore: update create table integration test result

ZuLiangWang:
## Rationale
Because the logic related to create table has been reconstructed in `horaemeta` https://github.com/CeresDB/horaemeta/pull/286, the results returned when reporting errors during create table have changed. Therefore, we need to correct the results of table creation in the integration test.

## Detailed Changes
* Update create table result in integration test.

## Test Plan
No need.


jiacai2050 (approved):
LGTM

Issue #1339, opened 2023-12-04
Local and CI license checking and formatting use the same tools.


tanruixiang:
### Describe This Problem

Now Ci  uses HawkEye to check license, but precommit use https://github.com/Lucas-C/pre-commit-hooks to format license.

can refer to the discussion in https://github.com/CeresDB/horaedb/pull/1336

### Proposal

Local and CI license checking and formatting use the same tools.

tanruixiang:
> @tanruixiang I would love to work on this!

@ananta  Thank you for your attention. Feel free to do that. Have fun! Note that dicussion in https://github.com/CeresDB/horaedb/pull/1336 is important information.

PR #1348, opened 2023-12-06
chore: unify license header checks

tisonkun:
This closes https://github.com/CeresDB/horaedb/issues/1339.

@ananta I found your comment later this patch was made, you can join in the review instead >_<


ananta:
> This closes #1339.
> 
> @ananta I found your comment later this patch was made, you can join in the review instead >_<

Hehe you're good 😄 . Thanks for the contribution! 🚀

tanruixiang:
Make sense to me. The original way also will format. @ShiKaiWi @jiacai2050 Should be using git hooks a lot. Do you accept this dependency on docker?

tisonkun:
> The original way also will format

Then we can change `check` to `format`.

jiacai2050:
I don't like this change very much...
1. This hook will be invoked on every commit, run docker is more heavier than a simple python script.
2. docker has poor performance on mac, and user may don't even have docker installed at all.

So I suggest we adopt a simple way to do this task.

PR #1349, opened 2023-12-06
feat: use string for request id

### LINKED ISSUE #1178: Request id will be duplicated in partitioned table situation
### Describe This Problem

Now, request id in ceresdb wil just be unique in a standalone node.
However, partitioned table's query/write will involve multiple nodes...

### Proposal

Maybe we can add `node ip` or `hostname` as a part of `request id`?

### Additional Context

_No response_

jiacai2050:
## Rationale
Close #1178

## Detailed Changes
- Use string to represent request_id

## Test Plan
Manually.

### DIFF
--- a/common_types/src/request_id.rs
+++ b/common_types/src/request_id.rs
@@ -14,26 +14,24 @@
 
 //! Request id.
 
-use std::{
-    fmt,
-    sync::atomic::{AtomicU64, Ordering},
-};
+use std::fmt;
 
-#[derive(Debug, Clone, Copy)]
-pub struct RequestId(u64);
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct RequestId(String);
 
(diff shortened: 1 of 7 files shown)


common_types/src/request_id.rs
```diff
@@ -14,27 +14,21 @@
...
+pub struct RequestId(String);
 
 impl RequestId {
     /// Acquire next request id.
     pub fn next_id() -> Self {
-        static NEXT_ID: AtomicU64 = AtomicU64::new(1);
-
-        let id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
-
+        let id = uuid::Uuid::new_v4().to_string();
```
jiacai2050, line 25: v4 is random string like `575c02e1-cd92-4c35-a5f3-353781163e93`

Rachelint (approved):
LGTM

PR #1352, opened 2023-12-07
chore: disable frequently failed tests

jiacai2050:
## Rationale
<del>This issue happens quite a lot in CI, and developer can do nothing beside retry, which is a very annoying thing. Since this test is cased by race condition, and there is no easy way to fix it, so I suggest we disable it for now.
</del>

## Detailed Changes
Increase wait time to 10s.

## Test Plan

CI


Rachelint (approved):
LGTM

PR #1362, opened 2023-12-11
chore: ignore flush failure when flush

ShiKaiWi:
## Rationale
The flush failure before table close may lead to shard close failure. However, such failure is tolerable because the unflushed can still be recovered during the following table open.

## Detailed Changes
Ignore the flush failure before closing table.

## Test Plan
Pass all the tests in the ci.

### DIFF
--- a/analytic_engine/src/instance/close.rs
+++ b/analytic_engine/src/instance/close.rs
@@ -25,6 +25,7 @@ use crate::{
     },
     manifest::{ManifestRef, SnapshotRequest},
     space::SpaceRef,
+    table::data::TableDataRef,
 };
 
 pub(crate) struct Closer {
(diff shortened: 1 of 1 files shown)


analytic_engine/src/instance/close.rs
```diff
@@ -47,13 +62,34 @@ impl Closer {
             }
         };
 
+        // Do flush before close for the fast recovery during the following opening.
+        // And it should not stop closing if flush fails.
+        if let Err(e) = self.flush(&table_data).await {
+            warn!("Ignore the failure to flush data before close, table:{}, err:{e}", table_data.name);
```
jiacai2050, line 68: Better to also include `table_id` in log.

PR #1363, opened 2023-12-11
feat: support metrics for number of bytes fetched from object storage

ShiKaiWi:
## Rationale
The metrics about the fetched data from the object storage helps estimate the load of query on one specific table, and the table's load distribution can help build a cluster topology with better load balance.

## Detailed Changes
Collect metrics for bytes fetched from object storage.

## Test Plan
Check the added metrics manually.


jiacai2050 (approved):
LGTM

PR #1364, opened 2023-12-12
fix: collect metrics for `get_ranges`

ShiKaiWi:
## Rationale
Metrics about fetched sst bytes with `get_ranges` is not collected.

## Detailed Changes
Collect the missing metrics.

## Test Plan
Query and check the metrics after executing a query.


jiacai2050 (approved):
LGTM

PR #1365, opened 2023-12-12
feat: avoid building dictionary for massive unique column values

### LINKED ISSUE #1105: Avoid abuse dictionary encoding in parquet writer
### Describe This Problem

Currently, the default configuration is used for parquet writer to generate the sst, and dictionary encoding is used for bytes column (including string column) by default. Only the dictionary page exceeds 1MB, the parquet writer will fall back to plain encoding without the dictionary.

However, it maybe not very efficient for the column of high cardinality to be encoded by dictionary.

### Proposal

Maybe we should choose to control whether to choose dictionary encoding according to the column's cardinality. However, the threshold of cardinality to enable dictionary encoding should be benchmarked and tested.

### Additional Context

_No response_

ShiKaiWi:
## Rationale
Close #1105

## Detailed Changes
- Reduce the parameters in the sst write path
- Avoid building dictionary for massive unique column values

## Test Plan
- New unit test for the changeset
- Observe the metrics for disable/enable dictionary encoding

### DIFF
--- a/analytic_engine/Cargo.toml
+++ b/analytic_engine/Cargo.toml
@@ -28,61 +28,61 @@ workspace = true
 workspace = true
 
 [features]
-test              = ["tempfile"]
-wal-table-kv      = ["wal/wal-table-kv"]
+test = ["tempfile"]
+wal-table-kv = ["wal/wal-table-kv"]
 wal-message-queue = ["wal/wal-message-queue"]
-wal-rocksdb       = ["wal/wal-rocksdb"]
+wal-rocksdb = ["wal/wal-rocksdb"]
 
 [dependencies]
 # In alphabetical order
-arc-swap             = "1.4.0"
-arena                = { workspace = true }
-arrow                = { workspace = true }
-async-stream         = { workspace = true }
-async-trait          = { workspace = true }
-atomic_enum          = { workspace = true }
-base64               = { workspace = true }
-bytes_ext            = { workspace = true }
-ceresdbproto         = { workspace = true }
-codec                = { workspace = true }
-common_types         = { workspace = true }
-datafusion           = { workspace = true }
-future_ext           = { workspace = true }
-futures              = { workspace = true }
-generic_error        = { workspace = true }
-hash_ext             = { workspace = true }
-hex                  = { workspace = true }
-hyperloglog          = { workspace = true }
-id_allocator         = { workspace = true }
-itertools            = { workspace = true }
-lazy_static          = { workspace = true }
-logger               = { workspace = true }
-lru                  = { workspace = true }
-macros               = { workspace = true }
-message_queue        = { workspace = true }
-metric_ext           = { workspace = true }
-object_store         = { workspace = true }
-parquet              = { workspace = true }
-parquet_ext          = { workspace = true }
-prometheus           = { workspace = true }
-prost                = { workspace = true }
+arc-swap = "1.4.0"
+arena = { workspace = true }
+arrow = { workspace = true }
+async-stream = { workspace = true }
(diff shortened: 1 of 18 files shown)


analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -511,6 +518,49 @@ impl<'a> SstWriter for ParquetSstWriter<'a> {
...
 }
 
+struct ColumnEncodingSampler<'a> {
+    sample_row_groups: &'a [RecordBatchWithKey],
+    meta_data: &'a MetaData,
+    min_num_sample_rows: usize,
+    max_unique_value_ratio: f64,
+}
+
+impl<'a> ColumnEncodingSampler<'a> {
```
jiacai2050, line 528: This sampler is not necessary here, we already keep distinct values for low cardinality column in sst metadata, we can use them directly.

When pick sst for compaction,  we can get metadata of input sst, then we can decide which column need dict encoding there directly.

ShiKaiWi: For the sst whose level > 0, the stats about the column values in the metadata can be used (I'll fix it in a following PR). However, level-0 ssts don't contain such information so let's keep it here.

analytic_engine/Cargo.toml
```diff
@@ -28,60 +28,61 @@ workspace = true
...
-lru = { workspace = true }
-macros = { workspace = true }
-message_queue = { workspace = true }
-metric_ext = { workspace = true }
-object_store = { workspace = true }
-parquet = { workspace = true }
-parquet_ext = { workspace = true }
-prometheus = { workspace = true }
-prost = { workspace = true }
+arc-swap             = "1.4.0"
```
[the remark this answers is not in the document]
ShiKaiWi: I guess my local toml linter do this for me. I'll revert it.

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -102,21 +101,32 @@ struct RecordBatchGroupWriter {
...
+        options: &'a WriteOptions,
     ) -> Self {
-        let column_values = if level.is_min() {
-            // There are not many rows in min level, so we don't record values for them.
-            None
-        } else {
-            let column_values = meta_data
+        // No need to build complex index for the min-level sst so there is no need to
+        // collect the column values.
+        let column_values = (!options.need_custom_filter()).then(|| {
```
jiacai2050, line 128: Opposite predicate?

analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -307,28 +321,24 @@ impl RecordBatchGroupWriter {
...
+            ParquetEncoder::try_new(sink, &self.meta_data.schema, &encode_options)
+                .box_err()
+                .context(EncodeRecordBatch)?;
+        let mut parquet_filter = self
+            .options
+            .need_custom_filter()
+            .then(ParquetFilter::default);
         let timestamp_index = self.meta_data.schema.timestamp_index();
-
         loop {
```
jiacai2050, line 341: Could we change loop to `while(!row_group.is_empty())`?

analytic_engine/src/sst/factory.rs
```diff
@@ -25,6 +25,7 @@ use snafu::{ResultExt, Snafu};
 use table_engine::predicate::PredicateRef;
 use trace_metric::MetricsCollector;
 
+use super::parquet::writer::WriteOptions;
```
jiacai2050, line 28: No super import

jiacai2050 (approved):
LGTM

PR #1369, opened 2023-12-15
fix: ignore collecting fetched bytes stats when sst file is read only once

ShiKaiWi:
## Rationale
The stats about the number of bytes fetched from object store should not include the low-frequency reading, e.g. compaction because such stats are used to show the query load distribution across the tables.

## Detailed Changes
Ignore collecting the fetched bytes stats in the low-frequency reading.

## Test Plan
The ci's tests should pass.

### DIFF
--- a/analytic_engine/src/sst/parquet/async_reader.rs
+++ b/analytic_engine/src/sst/parquet/async_reader.rs
@@ -309,9 +309,12 @@ impl<'a> Reader<'a> {
         );
 
         let mut streams = Vec::with_capacity(target_row_group_chunks.len());
-        let metrics_collector = ObjectStoreMetricsObserver {
-            collect_fetched_bytes_stats: matches!(self.frequency, ReadFrequency::Frequent),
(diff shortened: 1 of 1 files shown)


analytic_engine/src/sst/parquet/async_reader.rs
```diff
@@ -763,18 +764,21 @@ impl<'a> SstReader for ThreadedReader<'a> {
...
+        debug!("ObjectStoreReader dropped, path:{path}, elapsed:{elapsed:?}");
     }
 
     fn num_bytes_fetched(&self, _: &Path, num_bytes: usize) {
-        self.table_level_sst_metrics
-            .num_fetched_sst_bytes
-            .fetch_add(num_bytes as u64, Ordering::Relaxed);
+        if self.collect_fetched_bytes_stats {
+            self.table_level_sst_metrics
+                .num_fetched_sst_bytes
```
jiacai2050, line 779: Could we make this field Optional? then we don't need this flag here.

jiacai2050 (approved):
LGTM

PR #1372, opened 2023-12-18
feat: utilize the column cardinality for deciding whether to do dict

ShiKaiWi:
## Rationale
The column value set has been summarized into the meta data if the column is low distinct. With such information, the sampling for the columns can be skipped.

## Detailed Changes
Skip sampling over the low-cardinality columns.

## Test Plan
Updated the unit tests.

### DIFF
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -964,7 +964,6 @@ impl SpaceStore {
                 .context(ReadSstMeta)?;
 
             let column_stats = collect_column_stats_from_meta_datas(&sst_metas);
-
             let merged_meta = MetaData::merge(sst_metas.into_iter().map(MetaData::from), schema);
             (merged_meta, column_stats)
         };
@@ -1100,6 +1099,8 @@ fn collect_column_stats_from_meta_datas(metas: &[SstMetaData]) -> HashMap<String
 
     // Only the column whose cardinality is low in all the metas is a
     // low-cardinality column.
+    // TODO: shall we merge all the distinct values of the column to check whether
+    // the cardinality is still thought to be low?
     let low_cardinality_cols = low_cardinality_counts
         .into_iter()
         .filter_map(|(col_name, cnt)| {
--- a/analytic_engine/src/instance/flush_compaction.rs
+++ b/analytic_engine/src/instance/flush_compaction.rs
@@ -950,8 +950,7 @@ impl SpaceStore {
             row_iter::record_batch_with_key_iter_to_stream(merge_iter)
         };
 
-        let mut column_stats = HashMap::new();
-        let sst_meta = {
+        let (sst_meta, column_stats) = {
             let meta_reader = SstMetaReader {
                 space_id: table_data.space_id,
                 table_id: table_data.id,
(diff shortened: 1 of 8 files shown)


analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -956,6 +964,10 @@ impl SpaceStore {
                 .await
                 .context(ReadSstMeta)?;
 
+            if let Some(meta) = sst_metas.get(0) {
```
jiacai2050, line 967: Now you now check first sst's metadata, a better way to to merge all metas together, and see if column cardinality is still low.

ShiKaiWi: The merged `sst_meta` may be useless, so let me extract the information about the cardinality one by one.

tanruixiang:
During the time when you don't have write permission for a while, you can ping  me if any prs are ready to merge. @apache/horaedb-committers

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -1062,6 +1079,40 @@ impl SpaceStore {
...
+                        .and_modify(|v| *v += 1)
+                        .or_insert(1);
+                }
+            }
+        }
+    }
+
+    // Only the column whose cardinality is low in all the metas is a
+    // low-cardinality column.
+    let low_cardinality_cols = low_cardinality_counts
```
jiacai2050, line 1103: Beside this, we may also need to check if column values from all sst merged is still in low candinality.

In prod this may be not a serious problem, we can leave a comments here to do this in future.

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -1062,6 +1079,40 @@ impl SpaceStore {
     }
 }
 
+/// Collect the column stats from a batch of sst meta data.
+fn collect_column_stats_from_meta_datas(metas: &[SstMetaData]) -> HashMap<String, ColumnStats> {
+    let mut low_cardinality_counts: HashMap<String, usize> = HashMap::new();
```
jiacai2050, line 1084: Could we init its capacity? we have very wide table in prod.

analytic_engine/src/instance/flush_compaction.rs
```diff
@@ -956,7 +963,10 @@ impl SpaceStore {
                 .await
                 .context(ReadSstMeta)?;
 
-            MetaData::merge(sst_metas.into_iter().map(MetaData::from), schema)
+            let column_stats = collect_column_stats_from_meta_datas(&sst_metas);
+
```
jiacai2050, line 967:
```suggestion
```

analytic_engine/src/instance/flush_compaction.rs
ShiKaiWi: There is no entry for the columns whose cardinality is high, so I guess in the most of cases there is no need to do memory allocation.

jiacai2050: That's a good point.

jiacai2050 (approved):
LGTM

tanruixiang (approved):
Help @jiacai2050  agree to merge codes

PR #1388, opened 2023-12-21
fix: no write stall

ShiKaiWi:
## Rationale
#1003 tries to avoid frequent flush requests which may generate massive small ssts, but the write stall is also removed in the normal write path.

## Detailed Changes
Introduce the `min_flush_interval` to avoid frequent flush requests and recover the write stall mechanism.

## Test Plan
Add unit tests for the frequent flush check.


ShiKaiWi:
@tisonkun @tanruixiang Would you please approve the workflow to run when it's convenient for you?

jiacai2050 (approved):
LGTM

tanruixiang:
merging..

PR #1391, opened 2023-12-21
feat: wait for write notification in io runtime

ShiKaiWi:
## Rationale
The runtime's schedule latency may lead to the queue waiter's slow execution, and the total write latency may be influenced.

## Detailed Changes
Do the waiting for write notification in the `io_runtime` which won't be busy to ensure the queue waiter's quick response.

## Test Plan
Should pass all the CI.


ShiKaiWi:
Maybe this is caused by the write stall.

PR #1398, opened 2023-12-22
fix: missing and verbose logs

ShiKaiWi:
## Rationale
Some logs about query are verbose and some key logs about opening shard are missing.

## Detailed Changes
Remove verbose logs and add missing key logs.

## Test Plan
CI.


tanruixiang (approved):
LGTM. Merge after ci pass.

PR #1399, opened 2023-12-22
fix: fix broken link

caicancai:
## Rationale
use available link

## Detailed Changes
use available link

## Test Plan
ut

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ For features, we don't plan to support we will close the feature request ticket 
 
 ## Contributing Changes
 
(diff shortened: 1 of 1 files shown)


caicancai:
I didn't seem to find the address of the https://horaedb.apache.org/docs website, so I used the link to docs directly

caicancai:
I think it is more appropriate to use this for now, because the domain name does not seem to have been updated yet.
https://github.com/apache/incubator-horaedb-docs/blob/029e993f359907b7502b969a1cffeb8e43cb17fb/docs/src/en/dev/conventional_commit.md

tanruixiang (approved):
Thanks for your contribution. LGTM.

PR #1401, opened 2023-12-25
fix: skip wal encoding when data wal is disabled

jiacai2050:
## Rationale
When data wal is disable, data is still encoded, which waste cpu usage.

## Detailed Changes
Skip encode when data wal is disabled.

## Test Plan
CI.

### DIFF
--- a/analytic_engine/src/instance/write.rs
+++ b/analytic_engine/src/instance/write.rs
@@ -422,15 +422,23 @@ impl<'a> Writer<'a> {
 
         self.preprocess_write(&mut encode_ctx).await?;
 
-        let encoded_payload = {
-            let _timer = self.table_data.metrics.start_table_write_encode_timer();
-            let schema = self.table_data.schema();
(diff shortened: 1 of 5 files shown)


analytic_engine/src/instance/write.rs
```diff
@@ -422,15 +422,23 @@ impl<'a> Writer<'a> {
 
         self.preprocess_write(&mut encode_ctx).await?;
 
-        let encoded_payload = {
-            let _timer = self.table_data.metrics.start_table_write_encode_timer();
-            let schema = self.table_data.schema();
-            encode_ctx.encode(&self.instance.wal_encode, &schema)?
-        };
+        let seq = if self.instance.disable_wal {
+            MIN_SEQUENCE_NUMBER
```
ShiKaiWi, line 426: The real sequence number should be returned, otherwise the merge procedure may not work with the mock one.

ShiKaiWi (approved):
LGTM. And I guess we can remove codes for disabling wal writes in the wal manager.

PR #1260, opened 2023-10-13
feat: dist sql analyze

baojinri:
## Rationale
Currently, the analyze sql can not obtain detail metrics of partitioned table, 

## Detailed Changes
- return metrics of partitioned table to remote client
- collect metrics in client

## Test Plan
- Existing tests
- add new integration tests for explain analyze

PR #1410, opened 2023-12-28
refactor: avoid returning metrics in non-analyze sql

baojinri:
## Rationale
In #1260, we implemented distributed analyze, but for query that are not analyze, metrics will be returned, which will lead to a decrease in query performance. Therefore, we will fix it in this PR, and metrics will not be returned for normal queries.

## Detailed Changes
- Add is_analyze field to determine whether it is analyze

## Test Plan
Existing tests


jiacai2050 (approved):
LGTM

PR #1412, opened 2023-12-29
chore: fix error message

jiacai2050:
## Detailed Changes
- Attach endpoint to remote error

## Test Plan
CI

### DIFF
--- a/integration_tests/cases/env/cluster/ddl/partition_table.result
+++ b/integration_tests/cases/env/cluster/ddl/partition_table.result
@@ -99,9 +99,10 @@ ALTER TABLE partition_table_t ADD COLUMN (b string);
 
 affected_rows: 0
 
+-- SQLNESS REPLACE endpoint:(.*?), endpoint:xx,
 INSERT INTO partition_table_t (t, id, name, value) VALUES (1651737067000, 10, "ceresdb0", 100);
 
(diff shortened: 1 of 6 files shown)


remote_engine_client/src/lib.rs
```diff
@@ -93,12 +94,14 @@ pub mod error {
         },
 
         #[snafu(display(
-            "Failed to query from table in server, table_idents:{:?}, code:{}, msg:{}",
+            "Failed to query from table in server, table_idents:{:?}, endpoint:{}, code:{}, msg:{}",
             table_idents,
+            endpoint.to_string(),
```
ShiKaiWi, line 99: Isn't `Debug` trait derived by `Endpoint`?

integration_tests/cases/env/cluster/ddl/partition_table.result
```diff
@@ -99,9 +99,10 @@ ALTER TABLE partition_table_t ADD COLUMN (b string);
...
+Failed to execute query, err: Server(ServerError { code: 500, msg: "Failed to execute plan. Caused by: Internal error, msg:Failed to execute interpreter, err:Failed to execute insert, err:Failed to write table, err:Failed to write tables in batch, tables:[\"__partition_table_t_1\"], err:Failed to query from table in server, table_idents:[TableIdentifier { catalog: \"ceresdb\", schema: \"public\", table: \"__partition_table_t_1\" }], endpoint:xx, code:401, msg:failed to decode row group payload. Caused by: Schema mismatch with the write request, msg:expect 6 columns, but got 5. sql:INSERT INTO partition_table_t (t, id, name, value) VALUES (1651737067000, 10, \"ceresdb0\", 100);" })
```
ShiKaiWi, line 105: Why is the endpoint `xx`?

ShiKaiWi (approved):

remote_engine_client/src/lib.rs
jiacai2050:
```
impl ToString for Endpoint {
    fn to_string(&self) -> String {
        format!("{}:{}", self.addr, self.port)
    }
}

```
This is more readable then debug format.

integration_tests/cases/env/cluster/ddl/partition_table.result
jiacai2050: IP is not a fixed value, so I replace to `xx` with this
```
-- SQLNESS REPLACE endpoint:(.*?), endpoint:xx,
```

PR #1420, opened 2024-01-03
feat: block rules support query

jiacai2050:
## Rationale
Query with long time range usually cost too much resources, which affect stable of the whole cluster

## Detailed Changes
- Support block query by query range

## Test Plan
Manually
```bash

curl 0:5000/admin/block -H 'content-type: application/json' -d '
{
  "operation": "Set",
  "write_block_list": [],
  "read_block_list": [],
  "block_rules": [
    {"type": "QueryRange", "content": "24h"}
  ]
}'
```


ShiKaiWi (approved):
LGTM

PR #1426, opened 2024-01-04
fix: get_ranges is not spawned in io-runtime

ShiKaiWi:
## Rationale
The `get_ranges` should be executed in the io-runtime, but it is not now.

## Detailed Changes
Spawn the `get_ranges` task in the io-runtime.

## Test Plan
Should pass the CI.


components/object_store/src/metrics.rs
```diff
@@ -225,11 +225,22 @@ impl ObjectStore for StoreWithMetrics {
 
     async fn get_ranges(&self, location: &Path, ranges: &[Range<usize>]) -> Result<Vec<Bytes>> {
         let _timer = OBJECT_STORE_DURATION_HISTOGRAM.get_ranges.start_timer();
-        let result = self.store.get_ranges(location, ranges).await?;
+        let store = self.store.clone();
+        let loc = location.clone();
+        let ranges = ranges.to_vec();
+        let result = self
+            .runtime
+            .spawn(async move { store.get_ranges(&loc, &ranges).await })
```
jiacai2050, line 233: How about apply this logic to all methods of ObjectStore, instead of add one by one.

It's really a footgun when some methods are run in isolated run, while others are not.

jiacai2050 (approved):
LGTM

PR #1430, opened 2024-01-08
chore: skip wal seq check when wal is disabled

jiacai2050:
## Detailed Changes
- Disable seq check when wal is disabled
- Fix request id in remote query.


Rachelint (approved):
LGTM

PR #1431, opened 2024-01-08
feat: update disk cache in another thread to avoid blocking normal query process

jiacai2050:
## Rationale
When there is a cache miss in disk cache, it will 
1. Fetch data from remote
2. Insert data to cache, which will incur disk IO
3. Return the data for query.

We can move the second step to another thread to avoid it blocking the normal query process.

## Test Plan
CI


ShiKaiWi (approved):
LGTM

PR #1432, opened 2024-01-09
fix: compatible for old table options

Rachelint:
## Rationale
Make it compatible for old table options.

## Detailed Changes
When `layered_memtable_opts` not found in `TableOptions`, we disable `layered_memtable`.

## Test Plan
Test manually.


jiacai2050 (approved):
LGTM

PR #1440, opened 2024-01-15
feat: persist sst meta size

jiacai2050:
## Rationale
When read sst's metadata, it will use `get` method to read whole file, but there is no cache in `get`, so perf may degrade.

## Detailed Changes
- Add `meta_size` in SST's custom meta.
- When read SST's metadata, use `get_range` when we have `size`.

> The parquet writer is kind of massy now, will do some refactor after this PR.
## Test Plan

CI


ShiKaiWi (commented):
After the review, I find it's weird to fix the object store cache problem by avoiding using `get` because the interface of the object store doesn't tell the caller `get` is worse than `get_range`.

Maybe the right way is to ensure `get` method to work as expected so that any caller won't be required to remember the file size to use `get_range`.

jiacai2050:
The best way to fix is indeed to add cache for `get` method, but it's require more work, such as:
- Request duplicate
- Deal with IO failure

So here I choose to bypass `get`. In future We may use `get_range` with `0..` to replace `get`, but
this require object_store's support, so we may need to contribute there first.

src/analytic_engine/src/sst/meta_data/metadata_reader.rs
```diff
@@ -77,18 +82,29 @@ impl CustomMetadataReader for MetaV2Reader {
...
+                        .get(meta_path)
+                        .await
+                        .with_context(|| FetchFromStore {
+                            file_path: meta_path.to_string(),
+                        })?
+                        .bytes()
+                        .await
+                        .with_context(|| FetchAndDecodeSstMeta {
+                            file_path: meta_path.to_string(),
+                        })?,
```
ShiKaiWi, line 106: Add a comment to explain why we are in favor of `get_range`?

ShiKaiWi (approved):

Issue #1438, opened 2024-01-15
Statistics about the shards' loads and instance's load


ShiKaiWi:
### Describe This Problem

The statistics about the shards' loads and instance's load is necessary to address #1424.

### Proposal

- Instance load statistics
	- cpu usage
	- memory usage
- Shards' load statistics:
	- written bytes
	- read bytes

PR #1442, opened 2024-01-16
feat: add system_stats lib to collect system stats

ShiKaiWi:
## Rationale
Part of #1438. The preparations for reporting system statistics.

## Detailed Changes
- Add a new component called `system_stats` to helps collect system statistics.
- Introduce a dependency [`sysinfo`](https://crates.io/crates/sysinfo).

## Test Plan
Add a new unit test.

### DIFF
--- a/components/system_stats/src/lib.rs
+++ b/components/system_stats/src/lib.rs
@@ -72,7 +72,7 @@ impl SystemStatsCollector {
         tokio::time::sleep(wait_dur).await;
 
         let mut system = self.system.lock().unwrap();
-        system.refresh_specifics(Self::make_system_refresh_kind());
+        system.refresh_specifics(Self::make_cpu_and_mem_refresh_kind());
 
         SystemStats {
             cpu_usage: self.compute_cpu_usage(system.cpus()),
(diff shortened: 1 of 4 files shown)


components/system_stats/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+/// The stats about the system.
+#[derive(Debug)]
+pub struct SystemStats {
+    /// The usage's range is [0, 1.0]
+    pub cpu_usage: f32,
+    /// The memory is counted in byte.
+    pub used_memory: u64,
+    /// The memory is counted in byte.
+    pub total_memory: u64,
+    pub load_avg: LoadAvg,
```
ZuLiangWang, line 34: Should we expose primitive types in sysinfo lib? Maybe defining a new type is more appropriate?

ShiKaiWi: I've re-exported the `LoadAvg` in the library. And I guess no such need for `LoadAvg` is simple enough.

ZuLiangWang (approved):
LGTM

PR #1446, opened 2024-01-19
fix: table name is normalized when find timestamp column

jiacai2050:
## Detailed Changes
- Use `get_table_ref` to get `TableReference`, which will not normalize table name
- Return error when timestamp column is not found.

## Test Plan
CI


src/query_frontend/src/plan.rs
```diff
@@ -125,15 +138,15 @@ impl QueryPlan {
...
+    fn extract_time_range(&self) -> Result<Option<TimeRange>> {
+        let ts_column = if let Some(v) = self.find_timestamp_column()? {
             v
         } else {
             warn!(
                 "Couldn't find time column, plan:{:?}, table_name:{:?}",
                 self.df_plan, self.table_name
             );
-            return Some(TimeRange::min_to_max());
+            return Ok(Some(TimeRange::min_to_max()));
```
ZuLiangWang, line 149: Consider return `Err(...)` rather than `OK(...)`, expose err to upper layers.
```suggestion
            return Ok(Some(TimeRange::min_to_max()));
```

jiacai2050: The error is already return in `self.find_timestamp_column()?`, the else branch is when query range is not found, such as `select 1`

ZuLiangWang (approved):
LGTM

ZuLiangWang (approved):
LGTM

PR #1451, opened 2024-01-22
feat: support collect statistics about the engine

### LINKED ISSUE #1438: Statistics about the shards' loads and instance's load

ShiKaiWi:
## Rationale
Close #1438.

## Detailed Changes
Support report statistics on shard level.

## Test Plan
New unit tests is added and the CI should pass.

### DIFF
--- a/src/analytic_engine/src/engine.rs
+++ b/src/analytic_engine/src/engine.rs
@@ -17,23 +17,31 @@
 
 //! Implements the TableEngine trait
 
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 use async_trait::async_trait;
+use common_types::table::ShardId;
 use generic_error::BoxError;
 use logger::{error, info};
+use prometheus::{core::Collector, HistogramVec, IntCounterVec};
 use snafu::{OptionExt, ResultExt};
 use table_engine::{
     engine::{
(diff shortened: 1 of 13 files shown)


src/analytic_engine/src/engine.rs
```diff
@@ -234,6 +242,63 @@ impl TableEngine for TableEngineImpl {
 
         self.close_tables_of_shard(close_requests).await
     }
+
+    async fn report_statistics(&self) -> Result<Option<TableEngineStats>> {
```
jiacai2050, line 246: I think of one issue when using prometheus metrics to collect stats here, when a shard is transferred to another node, this metrics will remain, so you may collect those table twice.

ShiKaiWi: This case has been taken into consideration, and the solution is that a filter based on shards will be applied when reporting to meta.

jiacai2050: OK, a better way I can think of is to register those metrics when open shard, and unregister them when close shard.

jiacai2050 (approved):

PR #1454, opened 2024-01-22
chore: update create tables result

ZuLiangWang:
## Rationale
Refer to: https://github.com/apache/incubator-horaedb-meta/pull/286, since we implement idempotent table creation, the error message when repeatedly creating a table has changed, and the results of the integration test need to be modified.

## Detailed Changes
* Update `create_tables.result`

## Test Plan
No need.


ShiKaiWi (approved):
LGTM

PR #1456, opened 2024-01-22
fix: missing filter index over the primary keys

ShiKaiWi:
## Rationale
Current filter index won't be created over the primary key columns, and it leads to unexpected low prune rate in some cases.

## Detailed Changes
- Separate the filter module as a sub-module of the meta_data
- Create filter index over the primary key columns except the tsid and timestamp column.

## Test Plan
Should pass the CI.


jiacai2050 (approved):
LGTM

PR #1457, opened 2024-01-22
refactor: adjust cpu's stats

ShiKaiWi:
## Rationale
Only the total cpu usage is implicit if the numbers of cpus are different across the cluster nodes.

## Detailed Changes
Introduce the `num_cpus` in the system stats, and adjust the average cpu_usage to the sum of all the cpus' usage, whose range is `[0.0, num_cpus]`.

## Test Plan
Add a new check to ensure the `cpu_usage`'s range to to be `[0.0, num_cpus]`.


ZuLiangWang (approved):
LGTM

PR #1459, opened 2024-01-23
fix: random failure of test_collect_system_stats

ShiKaiWi:
## Rationale
The unit test `test_collect_system_stats` often fails when running on the github ci server.

## Detailed Changes
Remove the check for cpu usages.

## Test Plan
Should pass the CI.


ZuLiangWang (approved):
LGTM

PR #1476, opened 2024-01-31
refactor: refactor compaction process for remote compaction

Rachelint:
## Rationale
The first step to support compaction offload #1480 

## Detailed Changes
- Extract the common compaction logic to `CompactionExecutor`.
- Define the `CompactionRunner` trait to support remote compaction after.
- Define the `Compactor` to integration the new compaction logics, and insert them into main process.

## Test Plan
Test manually.


.gitignore
```diff
@@ -6,6 +6,7 @@ target
 integration_tests/dist_query/dist-query-testing
 integration_tests/dist_query/tsbs
 integration_tests/dist_query/output
+integration_tests/ceresmeta/
```
jiacai2050, line 9: This file is not used any more, you can just delete this file.

In newer version, meta will be generated in `target` dir.

jiacai2050 (approved):
LGTM

PR #1483, opened 2024-02-20
feat(horaemeta): add node inspector

ZuLiangWang:
## Rationale
In some extreme scenarios, etcd events may be lost or update failures may occur, which may cause the mapping between nodes and shards to fail to be updated correctly. We need to add a cover-up mechanism to ensure that the mapping relationship is always correct.

## Detailed Changes
* Add `NodeInspector`, it will start when the cluster start, running in he background and detect the status of nodes.

## Test Plan
Pass CI.

### DIFF
--- a/horaemeta/server/coordinator/inspector/node_inspector.go
+++ b/horaemeta/server/coordinator/inspector/node_inspector.go
@@ -22,77 +22,102 @@ package inspector
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 	"time"
 
+	"github.com/apache/incubator-horaedb-meta/pkg/coderr"
 	"github.com/apache/incubator-horaedb-meta/pkg/log"
 	"github.com/apache/incubator-horaedb-meta/server/cluster/metadata"
 	"github.com/apache/incubator-horaedb-meta/server/storage"
 	"go.uber.org/zap"
 )
 
+var ErrStartAgain = coderr.NewCodeError(coderr.Internal, "try to start again")
+var ErrStopNotStart = coderr.NewCodeError(coderr.Internal, "try to stop a not-started inspector")
+
 const inspectInterval = time.Second * 5
 
 // NodeInspector will inspect node status and remove expired data.
 type NodeInspector struct {
 	logger          *zap.Logger
-	clusterMetadata *metadata.ClusterMetadata
+	clusterMetadata ClusterMetaDataManipulator
 
-	// This lock is used to protect the following field.
-	lock      sync.RWMutex
-	isRunning atomic.Bool
+	starter sync.Once
+	// After `Start` is called, the following fields will be initialized
+	stopCtx     context.Context
+	bgJobCancel context.CancelFunc
 }
 
-func NewNodeInspector(logger *zap.Logger, clusterMetadata *metadata.ClusterMetadata) *NodeInspector {
+// ClusterMetaDataManipulator provides the snapshot for NodeInspector to check and utilities of drop expired shard nodes.
+type ClusterMetaDataManipulator interface {
+	GetClusterSnapshot() metadata.Snapshot
+	DropShardNode(context.Context, []storage.ShardNode) error
+}
+
+func NewNodeInspector(logger *zap.Logger, clusterMetadata ClusterMetaDataManipulator) *NodeInspector {
 	return &NodeInspector{
 		logger:          logger,
 		clusterMetadata: clusterMetadata,
-		lock:            sync.RWMutex{},
-		isRunning:       atomic.Bool{},
+		starter:         sync.Once{},
+		stopCtx:         nil,
+		bgJobCancel:     nil,
 	}
 }
 
 func (i *NodeInspector) Start(ctx context.Context) error {
-	i.lock.Lock()
-	defer i.lock.Unlock()
+	started := false
+	i.starter.Do(func() {
+		log.Info("node inspector start")
+		started = true
+		i.stopCtx, i.bgJobCancel = context.WithCancel(ctx)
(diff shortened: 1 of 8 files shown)


horaemeta/server/coordinator/inspector/node_inspector.go
```diff
@@ -0,0 +1,122 @@
...
+const inspectInterval = time.Second * 5
+
+// NodeInspector will inspect node status and remove expired data.
+type NodeInspector struct {
+	logger          *zap.Logger
+	clusterMetadata *metadata.ClusterMetadata
+
+	// This lock is used to protect the following field.
+	lock      sync.RWMutex
+	isRunning atomic.Bool
```
ShiKaiWi, line 43: I guess no need to protect a atomic boolean with a lock.

horaemeta/server/coordinator/inspector/node_inspector.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+func (i *NodeInspector) Start(ctx context.Context) error {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	if i.isRunning.Load() {
+		return nil
+	}
```
ShiKaiWi, line 61: If you want to ensure the forloop to be started once, `sync.Once` may be a better choice. BTW, current way can't ensure that the following forloop is started only once.

horaemeta/server/coordinator/inspector/node_inspector.go
```diff
@@ -0,0 +1,122 @@
...
+func (i *NodeInspector) inspect(ctx context.Context, snapshot metadata.Snapshot) {
+	shardNodeMapping := make(map[string][]storage.ShardNode, len(snapshot.RegisteredNodes))
+	for _, shardNode := range snapshot.Topology.ClusterView.ShardNodes {
+		if _, exists := shardNodeMapping[shardNode.NodeName]; !exists {
+			shardNodeMapping[shardNode.NodeName] = []storage.ShardNode{}
+		}
+		shardNodeMapping[shardNode.NodeName] = append(shardNodeMapping[shardNode.NodeName], shardNode)
+	}
+
+	expiredShardNodes := make([]storage.ShardNode, 0, len(snapshot.RegisteredNodes))
```
ShiKaiWi, line 104: Mostly, expired nodes should not exist so it's better to do memory allocation here.

horaemeta/server/coordinator/inspector/node_inspector.go
```diff
@@ -0,0 +1,122 @@
...
+			if !exists {
+				log.Debug("try to remove non-existent node")
+				continue
+			}
+			expiredShardNodes = append(expiredShardNodes, expired...)
+		}
+	}
+
+	// Try to remove useless data if it exists.
+	if err := i.clusterMetadata.DropShardNode(ctx, expiredShardNodes); err != nil {
```
ShiKaiWi, line 119: Shall we add a log for deleting expired nodes for future troubleshooting?

horaemeta/server/coordinator/inspector/node_inspector.go
```diff
@@ -0,0 +1,122 @@
...
+	// Check node status.
+	now := time.Now()
+	for _, node := range snapshot.RegisteredNodes {
+		if node.IsExpired(now) {
+			expired, exists := shardNodeMapping[node.Node.Name]
+			if !exists {
+				log.Debug("try to remove non-existent node")
+				continue
+			}
+			expiredShardNodes = append(expiredShardNodes, expired...)
```
ShiKaiWi, line 114: I guess the debug log is verbose here:
```suggestion
			if expired, exists := shardNodeMapping[node.Node.Name]; exists {
			    expiredShardNodes = append(expiredShardNodes, expired...)
			}
```

horaemeta/server/coordinator/inspector/node_inspector.go
ZuLiangWang: `isRunning` is used to cancel the loop rather than started only once, and the lock is needed for ensure only one caller can enter this loop at the same time, it is ensure that the following forloop is started only once.

ShiKaiWi: > the lock is needed for ensure only one caller can enter this loop at the same time, it is ensure that the following forloop is started only once.

I guess the lock can't make it because the goroutine may not be started when the lock is released.

horaemeta/server/coordinator/inspector/node_inspector.go
ShiKaiWi: Done

horaemeta/server/coordinator/inspector/node_inspector.go
ShiKaiWi: I've fixed it in the newest commints.

horaemeta/server/coordinator/inspector/node_inspector.go
```diff
@@ -0,0 +1,147 @@
...
+		stopCtx:         nil,
+		bgJobCancel:     nil,
+	}
+}
+
+func NewNodeInspector(logger *zap.Logger, clusterMetadata ClusterMetaDataManipulator) *NodeInspector {
+	return NewNodeInspectorWithInterval(logger, clusterMetadata, defaultInspectInterval)
+}
+
+func (i *NodeInspector) Start(ctx context.Context) error {
```
jiacai2050, line 72: It's weird to use `i` here, `i`, `j` are usually used for temp variables. how about `ni`?

jiacai2050 (approved):
LGTM

PR #1486, opened 2024-02-21
feat: add exponential backoff when retry

zealchen:
## Rationale

Add exponential backoff retry.

## Detailed Changes

follow [this](https://github.com/apache/arrow-rs/blob/dfb642809e93c2c1b8343692f4e4b3080000f988/object_store/src/client/backoff.rs#L26) backoff implementation.

## Test Plan

New UT.


jiacai2050:
@zealchen CI fails.

src/components/future_ext/src/retry.rs
```diff
@@ -20,35 +20,99 @@
...
+    fn default() -> Self {
+        Self {
+            init_backoff: Duration::from_millis(100),
+            max_backoff: Duration::from_millis(500),
+            base: 3.,
+        }
+    }
+}
+
+pub struct Backoff {
```
jiacai2050, line 58: Add comments explain where this is copied.

src/components/future_ext/src/retry.rs
```diff
@@ -20,35 +20,99 @@
...
+
 pub async fn retry_async<F, Fut, T, E>(f: F, config: &RetryConfig) -> Fut::Output
 where
     F: Fn() -> Fut,
     Fut: Future<Output = Result<T, E>>,
 {
+    let mut backoff = Backoff::new(&config.backoff);
     for _ in 0..config.max_retries {
-        let result = f().await;
+        let result: Result<T, E> = f().await;
```
jiacai2050, line 110: Is this type hint a must?

src/components/future_ext/src/retry.rs
```diff
@@ -109,4 +175,47 @@ mod tests {
             assert_eq!(3, runs.load(Ordering::Relaxed));
         }
     }
+
+    #[test]
+    fn test_backoff() {
+        let init_backoff_secs = 1.;
```
jiacai2050, line 181: I suggest we write `1.0` explicitly here, in case of misunderstanding.

src/components/future_ext/src/retry.rs
```diff
@@ -20,35 +20,99 @@
...
+    pub init_backoff: Duration,
+    /// The maximum backoff duration
+    pub max_backoff: Duration,
+    /// The base of the exponential to use
+    pub base: f64,
+}
+
+impl Default for BackoffConfig {
+    fn default() -> Self {
+        Self {
```
jiacai2050, line 50: I suggest we keep the same with array:
```rust
    fn default() -> Self {
        Self {
            init_backoff: Duration::from_millis(100),
            max_backoff: Duration::from_secs(15),
            base: 2.,
        }
    }
```

src/analytic_engine/src/sst/file.rs
```diff
@@ -540,7 +540,11 @@ pub struct FilePurger {
 impl FilePurger {
     const RETRY_CONFIG: RetryConfig = RetryConfig {
         max_retries: 3,
-        interval: Duration::from_millis(500),
+        backoff: BackoffConfig {
+            init_backoff: Duration::from_millis(100),
```
jiacai2050, line 544:
```
init: 500ms, max: 5s
```

src/components/future_ext/src/retry.rs
zealchen: Not necessary.

jiacai2050 (approved):
LGTM

PR #1489, opened 2024-02-28
deps: upgrade clap version

baojinri:
## Rationale
In order to implement ctl in the future, the clap version is relatively old and many functions are not supported, so we upgraded to the new version.

## Detailed Changes
- upgrade clap version
- Adapt to new APIs

## Test Plan
CI


Cargo.toml
```diff
@@ -101,7 +101,7 @@ catalog_impls = { path = "src/catalog_impls" }
 horaedbproto = { git = "https://github.com/apache/incubator-horaedb-proto.git", rev = "19ece8f771fc0b3e8e734072cc3d8040de6c74cb" }
 codec = { path = "src/components/codec" }
 chrono = "0.4"
-clap = "3.0"
+clap = "=4.4.18"
```
ShiKaiWi, line 104: Is it necessary to lock the version to the "4.4.18"?

baojinri: The latest version of clap is currently 4.5.1. However, during local testing, I found that the latest version may not be compatible with the rustc version, so the version is temporarily locked at 4.4.18.

jiacai2050: Clap is very aggressive for rust version In my impression, I wonder which features do you depend on but missing from current version?

I think our user case is simple enough that there is hardly any change we need up bump it.

rust-toolchain.toml
```diff
@@ -16,5 +16,5 @@
 # under the License.
 
 [toolchain]
-channel = "nightly-2023-08-28"
```
jiacai2050, line 19: Try search `nightly-2023-08-28` in our codebase, I remember some CI files also hard code it.

baojinri: OK

jiacai2050 (approved):
LGTM

PR #1495, opened 2024-03-05
fix: bump influxql, fix query priority

jiacai2050:
## Rationale
When time column is not in `select`, `extract_time_range` will not find time range of one query

## Detailed Changes
- https://github.com/CeresDB/influxql/pull/16

## Test Plan
CI


ShiKaiWi (approved):

Rachelint (approved):
LGTM

PR #1497, opened 2024-03-11
fix: check ASF header for all source files

jiacai2050:
## Rationale
All source file should have ASF headers.

## Test Plan
CI


tanruixiang (approved):
LGTM

PR #1499, opened 2024-03-12
chore: add some dependency project descriptions in notice

chunshao90:
## Rationale
Add some dependency project descriptions in notice.

## Test Plan
No need.


jiacai2050 (approved):
LGTM

PR #1503, opened 2024-03-18
chore: remove tini binary

jiacai2050:
## Rationale
Our repo should not contain any binaries.

## Detailed Changes
- Download tini instead of hard copy

## Test Plan
CI

Also I use command below to ensure no other binaries exists.
```
fd -t file --exclude target --exec file {} \; | grep -v ASCII | grep -v UTF-8
```


horaemeta/Dockerfile
```diff
@@ -44,7 +44,9 @@ RUN chmod +x /usr/bin/horaemeta-server
 COPY ./docker/entrypoint.sh /entrypoint.sh
 COPY ./config/example-standalone.toml /etc/horaemeta/horaemeta.toml
 
-COPY ./docker/tini /tini
+ENV TINI_VERSION v0.19.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
```
ShiKaiWi, line 49: Remove this duplicate line?

horaemeta/Dockerfile
```diff
@@ -44,7 +44,9 @@ RUN chmod +x /usr/bin/horaemeta-server
 COPY ./docker/entrypoint.sh /entrypoint.sh
 COPY ./config/example-standalone.toml /etc/horaemeta/horaemeta.toml
 
-COPY ./docker/tini /tini
+ENV TINI_VERSION v0.19.0
```
ShiKaiWi, line 47: Maybe we should move this line after L19.
BTW, I guess `ARG` command should work better than `ENV` because `TINI_VERSION` is not necessary to be known by the started container.

horaemeta/Dockerfile
jiacai2050: Fixed

ShiKaiWi (approved):

PR #1505, opened 2024-03-23
Feat concurrent replay wal

zealchen:
## Rationale
Close https://github.com/apache/incubator-horaedb/issues/1498

## Detailed Changes
Recover table logs in parallel.

## Test Plan
CI


src/analytic_engine/src/instance/wal_replayer.rs
```diff
@@ -181,19 +181,41 @@ impl Replay for TableBasedReplay {
...
+                if let Err(e) = ret {
+                    // If occur error, mark this table as failed and store the cause.
+                    failed_tables.lock().unwrap().insert(table_id, e);
+                }
+            },
+        )
+        .await;
 
-        Ok(failed_tables)
+        Ok(Arc::try_unwrap(failed_tables)
```
jiacai2050, line 215: It's not necessary to wrap `failed_tables` in a Mutex, you can refer how we do like this before here:

https://github.com/apache/incubator-horaedb/blob/54a5db725e674934ca490a64118d50df1b1f4cde/src/analytic_engine/src/instance/wal_replayer.rs#L401-L408

zealchen: Actually, in my previous commit [here](https://github.com/apache/incubator-horaedb/pull/1505/commits/8c3bd80747d63fc3766bafc2c961d72a53301898), I did like that. I feel like the current commit is more rusty. If you feel it is not the best practice I will roll back.

jiacai2050: IMO, `unwrap` doesn't sound idiomatic, I guess you mean iterator API is more rusty than plain for loop.

It depends, in most case it is, but not here, since you are using unnecessary Mutex/Arc here.
So I suggest rollback to previous commit.

zealchen: Make sense. Rollback now.

src/analytic_engine/src/instance/wal_replayer.rs
```diff
@@ -182,13 +182,28 @@ impl Replay for TableBasedReplay {
         debug!("Replay wal logs on table mode, context:{context}, tables:{table_datas:?}",);
 
         let mut failed_tables = HashMap::new();
-        let read_ctx = ReadContext {
+        let read_ctx = Arc::new(ReadContext {
```
jiacai2050, line 185: This Arc is not required.

```diff
-        for table_data in table_datas {
-            let table_id = table_data.id;
-            if let Err(e) = Self::recover_table_logs(context, table_data, &read_ctx).await {
+
+        let mut tasks = futures::stream::iter(
+            table_datas
+                .iter()
+                .map(|table_data| {
+                    let table_id = table_data.id;
+                    let read_ctx = &read_ctx;
+                    async move {
+                        let ret = Self::recover_table_logs(context, table_data, read_ctx).await;
+                        (table_id, ret)
+                    }
+                })
+                .collect::<Vec<_>>(),
+        )
+        .buffer_unordered(20);
+        while let Some((table_id, ret)) = tasks.next().await {
+            if let Err(e) = ret {
+                // If occur error, mark this table as failed and store the cause.
                 failed_tables.insert(table_id, e);
             }
         }

```

This compile success for me.

jiacai2050 (approved):
LGTM

PR #1507, opened 2024-03-28
fix: table blacklist not work for write

Rachelint:
## Rationale
Table blocking is found disable in grpc write path, that is annoying... 
This pr fix it for making us happy when handling the problematic tables in production.

## Detailed Changes
- support table blocking in grpc write path.
- add related its.

## Test Plan
CI

### DIFF
--- a/integration_tests/sdk/rust/src/main.rs
+++ b/integration_tests/sdk/rust/src/main.rs
@@ -31,6 +31,7 @@ use horaedb_client::{
 };
 
 const ENDPOINT: &str = "127.0.0.1:8831";
+const BLOCKED_TABLE: &str = "block_test_table";
 
 struct TestDatas {
     col_names: Vec<String>,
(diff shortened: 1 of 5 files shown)


src/proxy/src/write.rs
```diff
@@ -506,7 +506,35 @@ impl Proxy {
...
+            self.instance
+                .limiter
+                .try_limit(&plan)
+                .box_err()
+                .context(ErrWithCause {
+                    code: StatusCode::INTERNAL_SERVER_ERROR,
+                    msg: "table is blocked",
+                })?;
+
+            let insert_plan = match plan {
```
jiacai2050, line 524: This check seems unnecessary, we could move this `try_limit` inside `write_request_to_insert_plan`, so you don't need to do this.

integration_tests/sdk/rust/src/main.rs
```diff
@@ -117,11 +119,71 @@ async fn test_add_column(
     sql_query(client, rpc_ctx, timestamp, test_datas, true).await;
 }
 
-async fn drop_table_if_exists(client: &Arc<dyn DbClient>, rpc_ctx: &RpcContext, timestamp: i64) {
+async fn test_block_table(client: &Arc<dyn DbClient>, rpc_ctx: &RpcContext, timestamp: i64) {
```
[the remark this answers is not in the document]
Rachelint: It is used to check if table blocking is enable in sdk write path, I think place it in sdk test is resonable?

src/proxy/src/write.rs
Rachelint: I think the steps should be:
- build plan
- check plan
- execute the valid plan

It maybe  strange to mixed build step and check step?

jiacai2050 (approved):
LGTM

PR #1511, opened 2024-04-09
feat: add wal replay benchmark

zealchen:
## Rationale
Add benchmark cases for wal replay.

## Detailed Changes
Add benchmark cases for wal replay.

## Test Plan
Benchmark test. The result is as follows:
```
replay/replay/0         time:   [932.44 ms 1.0022 s 1.0705 s]
                                  change: [-12.172% -2.4758% +8.7572%] (p = 0.67 > 0.05)
```


src/benchmarks/src/replay_bench.rs
```diff
@@ -0,0 +1,157 @@
...
+            runtime: Arc::new(runtime),
+            test_ctx,
+            table: fixed_schema_table,
+        }
+    }
+
+    pub fn run_bench(&mut self) {
+        self.runtime.block_on(async {
+            let start_ms = self.test_ctx.start_ms();
+            let rows = [
```
jiacai2050, line 78: There are too few rows, I'm afraid you can't get much valuable metrics.

zealchen: Sure. I make this a configuration to auto-generate row tuples.

jiacai2050 (approved):
LGTM

Issue #1513, opened 2024-04-12
Explore new error define pattern


jiacai2050:
### Describe This Problem

Currently, we rely on snafu to define/handle errors, there are some issues with current usage.

First, let me explain how error is defined within horaedb.
- Every module has an an `Enum`-style error, which contains detailed error information
- When convert error from one module to another, we use `context` to do the heavy work.

This sounds like a good design pattern, but it become very verbose in practice. Some example:
- Every module force an `Enum`-style error, which contains too many unnecessary item.
- When consume error from module A in module B, we usually box them, since we don't want to define too many error item in module B. This make the error opaque again...

All in all, enum based error design should be limited in some way.

### Proposal

Error details should be opaque, if caller want to program based on it, the error provide a `pub fn kind() -> ErrorKind`.

In this way, we can better to encapsulate error details, and make error small.

### Additional Context

- https://matklad.github.io/2020/10/15/study-of-std-io-error.html

PR #1524, opened 2024-04-25
refactor: first step to refactor error

jiacai2050:
## Rationale
Part of #1513

## Detailed Changes
Replace snafu-based Error to thiserror-based for memtable module.

## Test Plan
CI

### DIFF
--- a/src/analytic_engine/src/memtable/error.rs
+++ b/src/analytic_engine/src/memtable/error.rs
@@ -42,15 +42,6 @@ impl Error {
     }
 }
 
-#[macro_export]
-macro_rules! ensure {
-    ($cond:expr, $msg:expr) => {
-        if !$cond {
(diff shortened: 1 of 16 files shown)


src/analytic_engine/src/memtable/error.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Error, Debug)]
+pub(crate) enum InnerError {
+    #[error("too large key, max:{max}, current:{current}")]
+    KeyTooLarge { current: usize, max: usize },
+
+    #[error(transparent)]
+    Other {
+        #[from]
+        source: anyhow::Error,
```
jiacai2050, line 62: When we convert an error to anyhow::Error, if it contains no backtrace, anyhow will automatically add one.

So we don't need to care backtrace.

chunshao90 (approved):
LGTM

PR #1534, opened 2024-05-24
feat: support OSS in memory multipart upload 

ZuLiangWang:
## Rationale
In some object storage and test scenarios, the support for the multi upload function is incomplete, and we need to directly use put to replace it.

## Detailed Changes
* Support OSS in memory multipart upload, it will batch the data in memory and upload it.
* Fix proxy config, setting proxy before will not take effect.

## Test Plan
paas CI.


jiacai2050:
We already switch to opendal, this is not used anymore.

PR #1535, opened 2024-05-24
docs: Update README.md

tisonkun:
Draft.

Rendered version: https://github.com/apache/incubator-horaedb/tree/tisonkun-patch-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@
 
 Apache HoraeDB (incubating) is a high-performance, distributed, cloud native time-series database.
 
+> [!IMPORTANT]
+> Apache HoraeDB (incubating) is an effort undergoing incubation at the Apache
+> Software Foundation (ASF), sponsored by the Apache Incubator PMC.
+>
+> Please read the [DISCLAIMER](DISCLAIMER) and a full explanation of ["incubating"](https://incubator.apache.org/policy/incubation.html).
+
 ## Documentation
 
 - [User Guide](https://horaedb.apache.org)


README.md
```diff
@@ -10,6 +10,12 @@
 
 Apache HoraeDB (incubating) is a high-performance, distributed, cloud native time-series database.
 
+> [!IMPORTANT]
+> Apache HoraeDB (incubating) is an effort undergoing incubation at the Apache
+> Software Foundation (ASF), sponsored by the Apache Incubator PMC.
+>
+> Please read the [DISCLAIMER](DISCLAIMER) and a full explanation of ["incubating"](https://incubator.apache.org/policy/incubation.html).
```
tisonkun, line 17: I can instead use the exact content of the DISCLAIMER. But since we're seeking for a shorter description, here is a proposal that I suppose is clear and concise.

cc @justinmclean @sebbasf

tisonkun: > the linked version

I don't know what this is. Would you send a suggestion or patch to show it?

> but either would do

So the version proposed here is OK to you?

tisonkun: @sebbASF Thanks for sharing your opinion. Could you provide a sentence that I can use here?

I'd prefer to try the "link to README.md" way but I may not make a precious sentence :P

justinmclean:
I'd rather we were not too prescriptive on the exact wording as was mentioned in the discussion on the Incubator general list.

tisonkun:
Thanks for sharing your ideas @sebbASF @justinmclean.

Then I propose we move forward with this wording and apply to other HoraeDB repos, and finally finish INFRA-25790.

You may take a look at other horaedb repos if they're suitable to follow this pattern. Perhaps some satellite repos can have different situation and have a better solution.

dave2wave:
Looks reasonable and points to the important detail. I would only add that if the podling is releasing with DISCLAIMER-WIP that the linked DISCLAIMER should be the DISCLAIMER-WIP with that added information.

Best,
Dave

jiacai2050 (approved):
Thanks everyone involved, let's move forward.

PR #1536, opened 2024-05-29
feat: support INSERT INTO SELECT

### LINKED ISSUE #557: Support INSERT INTO SELECT
### Describe This Problem

Enhance SQL support.

`INSERT INTO SELECT` is useful to migrate table, benchmark(to quickly generate lots of data).

### Proposal

Implement this SQL syntax.
```sql
INSERT INTO table2 (column1, column2, column3, ...)
SELECT column1, column2, column3, ...
FROM table1
WHERE condition;
```
### Additional Context
- https://dev.mysql.com/doc/refman/8.0/en/insert-select.html
- https://www.w3schools.com/mysql/mysql_insert_into_select.asp

dracoooooo:
## Rationale

Close  #557.

## Detailed Changes

When generating the insert logical plan, alse generate the select logical plan and store it in the insert plan. Then execute the select logical plan in the insert interpreter, convert the result records into RowGroup and then insert it.

## Test Plan

CI


jiacai2050:
@dracoooooo Hi, it seems there are some CI errors, would you mind take a look?

dracoooooo:
> @dracoooooo Hi, it seems there are some CI errors, would you mind take a look?

Fixed.

integration_tests/cases/common/dml/insert_into_select.result
jiacai2050, line 1: As said in https://github.com/apache/horaedb/issues/1538#issuecomment-2216948573, sql with multiple tables may fails, so currently we can only enable this features in standalone mode.

In cluster mode, we can do a check to ensure tables belonging to same instance.

For now, you can move this test under local dir, something like `integration_tests/cases/env/local/dml/insert_into_select.sql`

jiacai2050 (approved):
LGTM

PR #1537, opened 2024-06-07
chore: log more info for debug when tables with same name found

Rachelint:
## Rationale
We just panic but log nothing when found two tables with the same table name, it turns debugging into a disaster...

## Detailed Changes
Log the needed table infos when found two tables with the same table name before panic.

## Test Plan
Test maually.


ZuLiangWang (approved):
LGTM

Issue #1515, opened 2024-04-18
Explore new serialization library used for communication between horaedb-server


jiacai2050:
### Describe This Problem

Our [RemoteEngineService](https://github.com/apache/incubator-horaedb/blob/55bbe273ab4a09e2f3428582ca9b4cdb5ee88763/src/server/src/grpc/remote_engine_service/mod.rs#L924) use gRPC to communicate between different servers, and adopt protobuf as encode/decode methods, protobuf is not a very high performance library, and we observe high CPU usage in our production env.

### Proposal

Explore new serialization library, which support following features:
- Zero copy, which means its wire format is the same as memory layout.
- Native rust binding
- Support integration with [tonic](https://github.com/hyperium/tonic)

### Additional Context

Some choices:
- https://flatbuffers.dev/
- https://fury.apache.org/docs/start/usage#rust

jiacai2050:
@zealchen

zealchen:
> @zealchen

Okay. Those libs look super awesome, let me build a benchmark to demonstrate performance first.

jiacai2050:
Our benchmarks show flatbuffers has the best perf, and it can be integrated with tonic, so let's refactor our remote service.

https://github.com/apache/incubator-horaedb/blob/6baf6df0445f085dba4c330e12d019104483f2ee/src/server/src/grpc/remote_engine_service/mod.rs#L950

Currently we can only refactor the `write` API, since the CPU cost of it is the most obvious.

PR #1539, opened 2024-06-29
feat: Support remote engine server with flatbuffer write

zealchen:
## Rationale

[Our benchmarks](https://github.com/CeresDB/serialization-benchmark-rs) show flatbuffers has the best perf, thus we replace the  remote server write operation from protobuf to flatbuffers protocol. 

#1515 

## Detailed Changes

1. Add a new flatbuffer(remote_engine.fbs) protocol schema.
2. Implement  RemoteEngineFbService trait for RemoteEngineServiceImpl as the new flatbuffer-based service.
3. Make the write& batch_write interface backward compatible through the macros TonicWriteRequestExt and TonicWriteBatchRequestExt.
4. Add Client/Server side implementation for flatbuffer write.

### Manual test as follows:
1. run horaemeta-server and two hoaredb-server on my Mac(M1) local machine.
2. create a partition table: 
```sh
curl --location --request POST 'http://127.0.0.1:5440/sql' \
-d '
CREATE TABLE `demo`(
    `name`string TAG,
    `id` int TAG,
    `value` double NOT NULL,
    `t` timestamp NOT NULL,
    TIMESTAMP KEY(t)
    ) PARTITION BY KEY(name) PARTITIONS 2 ENGINE = Analytic
'
```
4. send write request:
```sh
curl --location --request POST 'http://127.0.0.1:15440/sql' \
-d '
INSERT INTO demo (t, name, value, id)
    VALUES (1702224000000, "horaedb", 100, 1)
'
```

### Write request serialization benchmark test:
```sh
ANALYTIC_BENCH_CONFIG_PATH=bench.toml cargo bench -p benchmarks
```
1)  ContiguousRowWriter inner type build with ByteVec:
pb encode: 131.916µs, pb decode: 63µs, fb encode: 174.459µs, fb decode: 83ns
2) ContiguousRowWriter inner type build with FlatbufferBuilder:
pb encode: 131.583µs, pb decode: 71.292µs, fb encode: 376.5µs, fb decode: 42ns


src/components/fb_util/src/common.rs
jiacai2050, line 1: `common.rs` is not suitable here, the file is only designed for remote_service, so `remote_service.rs` maybe better name.

src/server/src/grpc/mod.rs
```diff
@@ -165,6 +166,7 @@ pub struct RpcServices {
     rpc_server: InterceptedService<StorageServiceServer<StorageServiceImpl>, AuthWithFile>,
     meta_rpc_server: Option<MetaEventServiceServer<MetaServiceImpl>>,
     remote_engine_server: RemoteEngineServiceServer<RemoteEngineServiceImpl>,
+    remote_engine_server_ext: RemoteEngineFbServiceServer<RemoteEngineServiceImpl>,
```
jiacai2050, line 169: `remote_engine_fb_server` ?

src/remote_engine_client/src/client.rs
```diff
@@ -140,33 +144,35 @@ impl Client {
...
         let table_ident = request.table.clone();
         let endpoint = route_context.endpoint.clone();
-        let request_pb = request.convert_into_pb().box_err().context(Convert {
-            msg: "Failed to convert WriteRequest to pb",
+        let request_fb = request.convert_into_fb().box_err().context(Convert {
+            msg: "Failed to convert WriteRequest to fb",
         })?;
-        let mut rpc_client = RemoteEngineServiceClient::<Channel>::new(route_context.channel);
+
+        let mut rpc_client = RemoteEngineFbServiceClient::<Channel>::new(route_context.channel);
```
jiacai2050, line 151: Does this mean remote server will only have request with fb payload?

zealchen: Adding a REQUEST_TYPE environment variable to support both protobuf and flatbuffer protocol.

src/remote_engine_client/src/client.rs
```diff
@@ -643,3 +728,90 @@ fn convert_arrow_payload(mut v: ArrowPayload) -> Result<RecordBatch> {
                 })
         })
 }
+
+fn runtime_request_type() -> RequestType {
+    // get request type from environment variable
+    std::env::var("REQUEST_TYPE").map_or(RequestType::Protobuf, |v| {
```
jiacai2050, line 734:
```suggestion
    // get request type from environment variable
    // TODO: remove this config when fb write format is stable
    std::env::var("HORAEDB_REMOTE_WRITE_REQUEST_TYPE").map_or(RequestType::Protobuf, |v| {
```

src/server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -643,24 +702,53 @@ impl RemoteEngineServiceImpl {
...
+                        .runtimes
+                        .write_runtime
+                        .spawn(handle_write(ctx, WriteRequestExt::Proto(one_request)));
+                    write_table_handles.push(handle);
+                }
+                (write_table_handles, RequestType::Protobuf)
+            }
+            TonicWriteBatchRequestExt::Flatbuffer(v) => {
+                let request = Arc::new(v.into_inner());
+                let req = request.deserialize::<FBWriteBatchRequest>().unwrap();
```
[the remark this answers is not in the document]
zealchen: 1. FlatBufferBytes::deserialize is changed to unsafe function call as internal request routing is supposed to be safe among servers.
2. Other unwraps are all based on batch() which cannot be None in internal procedure.
3. Always follow the backwards compatibility solution for the schema design  (https://flatbuffers.dev/flatbuffers_guide_tutorial.html)

src/server/src/grpc/remote_engine_service/mod.rs
```diff
@@ -1419,3 +1606,32 @@ fn find_schema_by_identifier(
             ),
         })
 }
+
+fn build_fb_write_response(resp: WriteResponse) -> FlatBufferBytes {
+    let mut builder = flatbuffers::FlatBufferBuilder::with_capacity(1024);
+    let header = resp.header.unwrap();
```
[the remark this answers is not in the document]
zealchen: Change it to unwrap_or_default?

src/table_engine/src/remote/model.rs
```diff
@@ -109,6 +130,16 @@ pub struct TableIdentifier {
     pub table: String,
 }
 
+impl From<FBTableIdentifier<'_>> for TableIdentifier {
```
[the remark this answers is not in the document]
zealchen: Done.

jiacai2050:
As said in description, the improvement after introduce Flatbuffer is not what we originally expected, the reason is that this PR only refactor upper layer of the put request, the `ContiguousRow`'s encoding is unchanged, however it's where cost mainly come from, so we cannot get much perf improvement without refactor it.

But ContiguousRow is a heavily used structure in write path, refactor it would introduce much more changes,  so I suggest we stop working on this, sorry for not find this earlier. 

Really appreciated your work on this.

PR #1541, opened 2024-07-11
chore: upgrade object store version

baojinri:
## Rationale
The object store version is upgraded to 0.10.1 to prepare for access to opendal

## Detailed Changes
- Impl AsyncWrite for ObjectStoreMultiUpload
- Impl MultipartUpload for ObkvMultiPartUpload
- Adapt new api on query writing path

## Test Plan
- Existing tests


src/components/object_store/src/obkv/meta.rs
```diff
@@ -113,13 +116,9 @@ pub struct ObkvObjectMeta {
...
     pub unique_id: Option<String>,
-    /// The size in bytes of one part. Note: maybe the size of last part less
-    /// than part_size.
-    #[serde(rename = "part_size")]
-    pub part_size: usize,
-    /// The paths of multi upload parts.
+    /// The paths and size of multi upload parts.
     #[serde(rename = "parts")]
-    pub parts: Vec<String>,
+    pub parts: Vec<(String, usize)>,
```
jiacai2050, line 121: This struct is serialized, so it's breaking changes to update its fields?

If it's, we can define a proto to serialize this struct instead of using serde.

src/analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -405,67 +411,129 @@ impl<'a> RecordBatchGroupWriter<'a> {
...
+    meta_sink: ObjectStoreMultiUpload,
+    parquet_metadata: ParquetMetaData,
+) -> Result<usize> {
+    let buf = encode_sst_meta_data(parquet_metadata).context(EncodePbData)?;
+    let buf_size = buf.len();
+    meta_sink
+        .multi_upload
+        .lock()
+        .await
+        .put_part(buf.into())
```
jiacai2050, line 519: No need to use part_part, we can just write metadata in one `put`.

src/analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -405,67 +411,129 @@ impl<'a> RecordBatchGroupWriter<'a> {
...
-    W: AsyncWrite + Send + Unpin,
-{
-    let buf = encode_sst_meta_data(parquet_metadata).context(EncodePbData)?;
-    let bytes = buf.as_bytes();
-    let bytes_size = bytes.len();
-    meta_sink.write_all(bytes).await.with_context(|| Io {
-        file: meta_path.clone(),
-    })?;
+impl AsyncWrite for ObjectStoreMultiUpload {
+    // TODO: Currently,the data writing is serial, and data may need to be written
```
jiacai2050, line 455: In older version, the write process is parallel?

If it's, we need to implement it as well.

jiacai2050 (changes requested):

src/analytic_engine/src/sst/parquet/writer.rs
```diff
@@ -405,67 +414,116 @@ impl<'a> RecordBatchGroupWriter<'a> {
...
+    }
+}
+
+async fn write_metadata(
+    meta_sink: ObjectStoreMultiUpload,
+    parquet_metadata: ParquetMetaData,
+) -> Result<usize> {
+    let buf = encode_sst_meta_data(parquet_metadata).context(EncodePbData)?;
+    let buf_size = buf.len();
+    meta_sink.multi_upload.lock().await.put(buf);
```
jiacai2050, line 510: It seems we can only lock once here.
```
let upload = meta_sink.multi_upload.lock().await;
upload.put();
upload.finish();
```

jiacai2050 (approved):
LGTM

PR #1544, opened 2024-07-17
refactor: insert select to stream mode

### LINKED ISSUE #1542: INSERT INTO SELECT in a streaming way
### Describe This Problem

Followup for https://github.com/apache/horaedb/pull/1536, now we will collect all rows to vec before do insert, it's best if we can do insert select in a streaming way.

### Proposal

Support insert select like
```rs
while let Some(rows) = query.next_batch().await? {
  insert(rows);
}
```

### Additional Context

_No response_

zealchen:
## Rationale
Close #1542 

## Detailed Changes
Do select and insert procedure in stream way.

## Test Plan
CI test.

### DIFF
--- a/src/interpreters/src/insert.rs
+++ b/src/interpreters/src/insert.rs
@@ -132,6 +132,10 @@ pub enum Error {
 
 define_result!(Error);
 
+// TODO: make those configurable
+const INSERT_SELECT_ROW_BATCH_NUM: usize = 1000;
+const INSERT_SELECT_PENDING_BATCH_NUM: usize = 3;
+
 pub struct InsertInterpreter {
     ctx: Context,
     plan: InsertPlan,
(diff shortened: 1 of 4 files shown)


src/interpreters/src/insert.rs
```diff
@@ -168,38 +179,98 @@ impl Interpreter for InsertInterpreter {
                 .await
                 .context(Insert)?;
 
-                if record_batches.is_empty() {
-                    return Ok(Output::AffectedRows(0));
+                // accumulate 8 record batches before writing to table and this param
+                // can be adjusted according to the performance test result.
+                let max_batch_size = 8;
```
jiacai2050, line 184: I think `max_row_count` is more suitable here, 1000 is a reasonable default value.

src/interpreters/src/insert.rs
```diff
@@ -168,38 +179,98 @@ impl Interpreter for InsertInterpreter {
...
+                let mut result_rows = 0;
+                let mut record_batches = Vec::new();
+                while let Some(response) = record_batches_stream.next().await {
```
jiacai2050, line 187:
```rust
while let Some(response) = record_batches_stream.try_next().await? {
  ...
}
```

This looks more elegant.

zealchen:
Refactor to do the select and insert in parallel way.

jiacai2050 (approved):
LGTM

PR #1548, opened 2024-07-19
fix(comment): update error documentation comment for remote engine service

LeslieKid:
## Rationale
Updating an error comment in the code to reflect the correct service name is needed.

## Test Plan
No need


jiacai2050 (approved):
LGTM

Issue #1279, opened 2023-10-24
Introduce a new disk-based WAL implementation for standalone deployment


jiacai2050:
### Describe This Problem

After #1272, we have successfully put different WAL implementation behind feature gates, this is important to reduce compile time since wal based on RocksDB is very slow.

However, we have to enable rocksdb wal by default since we have no reliable WAL implementation, message queue and table-kv introduce more complex, so a better choice is to implement another wal based on disk directly.

### Proposal

The main trait to implement is WalManager
- https://github.com/CeresDB/ceresdb/blob/affa2c65a279252d10475c6a6d201cd7f60b5689/src/wal/src/manager.rs#L320

A simple structure I think of is like this:
```bash
$ tree wal
wal
└──region-1
   ├──f1
   ├──f2
   └──....
```
WAL of different tables is saved together in a fixed-size file(such as 64M), this have the advantage of fast write.

As for deletion, since different tables are saved intertwined, we need to loop all tables to get the minimal sequence location to delete.

The layout of each file could refer prometheus design:
https://github.com/prometheus/prometheus/blob/main/tsdb/docs/format/wal.md

### Additional Context

We will finish this task with help of @dracoooooo at [OSPP](https://summer-ospp.ac.cn/org/prodetail/2444e0563?list=org&navpage=org),  TODOs are listed below:

- [ ] Implement a disk-based WAL that can pass the existing unit tests.
  - [x] write
  - [x] read
  - [x] scan
  - [ ] delete
  - [ ] multiple segments
- [x] Remove `unwarp` and handle errors.
- [x] Add unit tests for the new code.
- [ ] Test on large-scale data.
- [ ] Compare with the existing RocksDB WAL implementation and optimize
performance.
_No response_

ShiKaiWi:
@jiacai2050 Should the replication of wal be taken into considerations?

jiacai2050:
Do you mean distributed WAL storage?

If it's, I don't think we need a component like this, kafka works well for this case.

PR #1552, opened 2024-08-04
feat: Add a new disk-based WAL implementation for standalone deployment

dracoooooo:
## Rationale

#1279

## Detailed Changes

1. Added a struct `Segment` responsible for reading and writing segment files, and it records the offset of each record.
2. Add a struct SegmentManager responsible for managing all segments, including:
	1.	Reading all segments from the folder upon creation.
	2.	Writing only to the segment with the largest ID.
	3.	Maintaining a cache where segments not in the cache are closed, while segments in the cache have their files open and are memory-mapped using mmap.
3. Implement the `WalManager` trait.

## Test Plan

Unit tests.

## Todos

- [ ] Implement a disk-based WAL that can pass the existing unit tests.
  - [x] write
  - [x] read
  - [x] scan
  - [ ] delete
  - [ ] multiple segments
- [x] Remove `unwarp` and handle errors.
- [x] Add unit tests for the new code.
- [ ] Test on large-scale data.
- [ ] Compare with the existing RocksDB WAL implementation and optimize performance.

### DIFF
--- a/src/wal/src/local_storage_impl/record_encoding.rs
+++ b/src/wal/src/local_storage_impl/record_encoding.rs
@@ -1,0 +1,259 @@
+use bytes_ext::{Buf, BufMut, SafeBuf, SafeBufMut};
+use codec::Encoder;
+use generic_error::GenericError;
+use macros::define_result;
+use snafu::{ensure, Backtrace, ResultExt, Snafu};
+
+pub const RECORD_ENCODING_V0: u8 = 0;
+pub const NEWEST_RECORD_ENCODING_VERSION: u8 = RECORD_ENCODING_V0;
+
+pub const VERSION_SIZE: usize = 1;
+pub const CRC_SIZE: usize = 4;
+pub const RECORD_LENGTH_SIZE: usize = 4;
+pub const KEY_LENGTH_SIZE: usize = 4;
+pub const VALUE_LENGTH_SIZE: usize = 4;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Version mismatch, expect:{}, given:{}", expected, given))]
+    Version { expected: u8, given: u8 },
+
+    #[snafu(display("Failed to encode, err:{}", source))]
+    Encoding { source: bytes_ext::Error },
+
+    #[snafu(display("Failed to decode, err:{}", source))]
+    Decoding { source: bytes_ext::Error },
+
+    #[snafu(display("Invalid record: {}, backtrace:\n{}", source, backtrace))]
+    InvalidRecord {
+        source: GenericError,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Length mismatch: expected {} but found {}", expected, actual))]
+    LengthMismatch { expected: usize, actual: usize },
+
+    #[snafu(display("Checksum mismatch: expected {}, but got {}", expected, actual))]
+    ChecksumMismatch { expected: u32, actual: u32 },
+}
+
+define_result!(Error);
+
+/// Record format:
+///
+/// ```text
+/// +---------+--------+--------+------------+-----+--------------+-------+
+/// | version |  crc   | length | key length | key | value length | value |
+/// |  (u8)   | (u32)  | (u32)  |   (u32)    |     |     (u32)    |       |
+/// +---------+--------+--------+------------+-----+--------------+-------+
+/// ```
+pub struct Record<'a> {
+    /// The version number of the record.
+    pub version: u8,
+
+    /// The CRC checksum of the record.
+    pub crc: u32,
+
+    /// The length of the record (excluding version, crc and length).
+    pub length: u32,
+
+    /// The length of the key in bytes.
+    pub key_length: u32,
+
+    /// Common log key.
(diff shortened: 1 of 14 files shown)


src/wal/src/local_storage_impl/segment.rs
```diff
@@ -0,0 +1,741 @@
...
+                .trim_end_matches(".wal")
+                .parse::<u64>()
+                .ok()
+            {
+                Some(id) => id,
+                None => continue,
+            };
+
+            let segment = Segment::new(path.to_string_lossy().to_string(), segment_id)?;
+            let segment_arc = Arc::new(Mutex::new(segment));
```
jiacai2050, line 371: Just name it `segment`.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -0,0 +1,741 @@
...
+                }
+            }
+        }
+        cache.push_back(segment_id);
+
+        Ok(segment.clone())
+    }
+
+    pub fn write(&self, _ctx: &WriteContext, batch: &LogWriteBatch) -> Result<SequenceNumber> {
+        let segment = self.get_segment(self.latest_segment_idx.load(Ordering::Relaxed))?;
```
jiacai2050, line 446: I found `latest_segment_idx` is only used here, do you forget to update it?

Also use a atomic integer may not a good choice, since there may exists race condition, I suggest use `current: Mutex<Segment>` to keep the latest segment for write.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -0,0 +1,741 @@
...
+            record_position.push(Position {
+                start: data.len() as u64,
+                end: (data.len() + record_content.len() + CRC_SIZE + RECORD_LENGTH_SIZE) as u64,
+            });
+
+            // Calculate and encode the CRC
+            let mut hasher = Hasher::new();
+            hasher.update(&record_content);
+            let crc = hasher.finalize();
+            data.put_u32_le(crc);
```
jiacai2050, line 486: Better to add an `RecordEncode` struct to do this, then we can easily test it.

jiacai2050 (changes requested):

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -0,0 +1,741 @@
...
+        self.is_open = false;
+        self.file.take();
+        self.mmap.take();
+        self.record_position.take();
+        Ok(())
+    }
+
+    pub fn append(&mut self, data: &[u8]) -> Result<()> {
+        ensure!(self.is_open, SegmentNotOpen { id: self.id });
+        ensure!(self.size + data.len() as u64 <= MAX_FILE_SIZE, SegmentFull);
```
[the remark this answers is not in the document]
dracoooooo: Currently, reading and writing have only been implemented on a single segment file. Reading and writing across multiple segments have not yet been implemented. I’ll leave a TODO here.

src/wal/src/local_storage_impl/segment.rs
dracoooooo: Currently, reading and writing have only been implemented on a single segment file, so the `latest_segment_idx` will not be updated.

Since all segments are stored in all_segments, I think storing just a segment ID here should be sufficient, so I plan to use `current: Mutex<u64>`.

src/wal/src/local_storage_impl/record_encoding.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub version: u8,
+
+    /// The CRC checksum of the record.
+    pub crc: u32,
+
+    /// The length of the record (excluding version, crc and length).
+    pub length: u32,
+
+    /// The length of the key in bytes.
+    pub key_length: u32,
```
jiacai2050, line 78: Our key is fixed sized now(`table_id+seq_num`), so I think we don't need key length field.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -0,0 +1,790 @@
...
+#[derive(Debug, Clone)]
+pub struct Position {
+    start: u64,
+    end: u64,
+}
+
+/// Segment file format:
+///
+/// ```text
+/// +----------+--------+--------+---+--------+
```
jiacai2050, line 158: Segment file also need to version field.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -0,0 +1,790 @@
...
+    #[snafu(display("Length mismatch: expected {} but found {}", expected, actual))]
+    LengthMismatch { expected: usize, actual: usize },
+
+    #[snafu(display("Checksum mismatch: expected {}, but got {}", expected, actual))]
+    ChecksumMismatch { expected: u32, actual: u32 },
+}
+
+define_result!(Error);
+
+const HEADER: &str = "HoraeDB WAL";
```
jiacai2050, line 112: Usually we define it like this
```suggestion
const SEGMENT_HEADER: &[u8] = b"HoraeDBWAL";
```

jiacai2050 (approved):
LGTM

PR #1557, opened 2024-08-24
feat: use opendal to access  underlying storage

baojinri:
## Rationale
Use opendal to access the object store, thus unifying the access method of the underlying storage.

## Detailed Changes
- use opendal to access s3/oss/local file

## Test Plan
- Existed tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -877,6 +877,18 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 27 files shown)


jiacai2050 (changes requested):

jiacai2050 (approved):
LGTM

PR #1561, opened 2024-08-31
chore: update link

caicancai:
## Rationale
I noticed that the previous repository has been archived, maybe it would be better to update the new link


caicancai:
cc @jiacai2050

jiacai2050 (approved):
LGTM

Issue #1545, opened 2024-07-17
Tracking issue for compaction offloading


LeslieKid:
### Describe This Problem

We found in production that the speed of sst compaction is unable to keep up with the speed of sst generation, leading to poor query performance. However we are unable give more resource to compaction to solve the problem because query/write is more important than compaction in the same node.

It is really hard to do a trade-off about resource allocation among query, write and compaction in lsm model. We want to compact the generated small ssts as fast as possible, but we can't tolerate its influence to query/write. And finally I think offload the compaction to the seperated nodes may be the key for it.

### Proposal

**The following is the architecture for compaction offloading.**

![Architecture.png](https://s2.loli.net/2024/07/17/ivH4E2MJq8pZbnQ.png)

To support compaction offloading, we need:

- **Compaction server supporting remote compaction service (Execute Plane)**
  - [ ] Introduce runnable compaction server and expose the api  for horaedb node to ask for remote compaction service
- **Horaemeta supports managing the compaction servers (Control Plane)**
  - [ ] Impl the ability to manage the compaction servers and collect their status (Monitor)
  - [ ] Impl scheduling algorithm for load balance and expose the api for horaedb node to get the proper remote compaction node (Scheduler)
- **Horaedb node supports submitting the compaction task to remote**
  - [ ] Impl remote mode compactor based on the provided compaction service and horaemeta service above

### Additional Context

This issue replaces issue #1480. Please close issue #1480 as it is outdated.

[incubator-horaedb-proto#133](https://github.com/apache/incubator-horaedb-proto/issues/133) is highly related to this issue.

Rachelint:
looks great!

LeslieKid:
> I think in the first stage, we can treat the `compaction cluster` as same as the `horaedb cluster`. And we just reuse the ability about monitoring and scheduling for `horaedb cluster` in `horaemeta`.
> 
> And we can start to desigin the specific `schedule` strategy for `compaction cluster` after it can work.

Thanks for your suggestion! I think it's a good idea to first implement a simple working version.

Rachelint:
I think `Horaedb node supports submitting the compaction task to remote` may be similar as `Compaction node supporting remote compaction service `.

I guess the third step can be integration test, test is very important actually.

Rachelint:
Maybe you will be insterested when writing test
https://github.com/apache/horaedb/tree/main/integration_tests/dist_query

PR #1563, opened 2024-09-03
feat: support horaedb submit compaction task to remote

LeslieKid:
## Rationale
The subtask to support compaction offloading. See #1545 

## Detailed Changes
**Compaction node support remote compaction service**

- Define `CompactionServiceImpl` to support compaction rpc service.
- Introduce `NodeType` to distinguish compaction node and horaedb node. Enable the deployment of compaction node.

- Impl `compaction_client` for horaedb node to access remote compaction node.

**Horaedb node support compaction offload**

- Introduce `compaction_mode` in analytic engine's `Config` to determine whether exec compaction offload or not.
- Define `CompactionNodePicker` trait, supporting get remote compaction node info.
- Impl `RemoteCompactionRunner`, supporting pick remote node and pass compaction task to the node.
- Add docs (e.g. `example-cluster-n.toml`) to explain how to deploy a cluster supporting compaction offload.


src/cluster/src/cluster_impl.rs
```diff
@@ -341,10 +346,58 @@ impl Inner {
...
+            .get_compaction_node()
+            .await
+            .expect("fail to get remote compaction node");
+
+        build_compaction_client(config)
+            .await
+            .expect("fail to build compaction client")
+    }
+
+    async fn compact(
```
Rachelint, line 379: Maybe it is not suitable to add `compact` function to `Cluster`.
I have some thoughts:
- We place all logic into `RemoteCompactionRunner`(pick node, pass task to node)
- Abstract `pick node` logic to a trait `RemoteCompactionNodePicker`, it have extrat benefits for easy testing, too.

src/analytic_engine/src/compaction/runner/remote_runner.rs
```diff
@@ -0,0 +1,48 @@
...
+    compaction::runner::{CompactionRunner, CompactionRunnerResult, CompactionRunnerTask},
+    instance::flush_compaction::{ConvertCompactionTaskResponse, RemoteCompact, Result},
+};
+
+pub struct RemoteCompactionRunner {
+    pub cluster: ClusterRef,
+}
+
+#[async_trait]
+impl CompactionRunner for RemoteCompactionRunner {
```
Rachelint, line 33: It may be good to place all remote compaction logic into this as described above.

src/analytic_engine/src/setup.rs
```diff
@@ -97,6 +99,8 @@ pub struct EngineBuilder<'a> {
     pub config: &'a Config,
     pub engine_runtimes: Arc<EngineRuntimes>,
     pub opened_wals: OpenedWals,
+    pub cluster: Option<ClusterRef>,
```
Rachelint, line 102: Maybe standalone engine should not know the cluster concept?

src/analytic_engine/src/compaction/runner/node_picker.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use macros::define_result;
+use meta_client::{types::FetchCompactionNodeRequest, MetaClientRef};
+use snafu::{Snafu, ResultExt};
+
+#[async_trait]
+pub trait RemoteCompactionNodePicker: Send + Sync {
```
Rachelint, line 28: We can impl a simple local picker initialized from conifg for easy testing.

Rachelint:
It looks good to me.
I just helped to sort out codes, mainly move the compaction client to runner crate.

I am checking it in my local, let's merge it after the double check!

LeslieKid:
> I am checking it in my local, let's merge it after the double check!

I check your commits in my local.
It's indeed better to move compaction client to runner crate.

I think we can merge it.

src/server/src/grpc/mod.rs
```diff
@@ -294,26 +310,54 @@ impl Builder {
...
+                        instance: instance.clone(),
+                        runtime: runtimes.meta_runtime.clone(),
+                        opened_wals,
+                    };
+                    meta_rpc_server = Some(MetaEventServiceServer::new(builder.build()));
+
+                    // Support remote compaction rpc service.
+                    let compaction_runner =
+                        self.compaction_runner.context(MissingCompactionRunner)?;
+                    let builder = compaction_service::Builder {
```
Rachelint, line 348: Maybe the compaction service is not something so related to the `Cluster`.
`Cluster` is used by the physical node to operate the `virtual node`(aka `shard`).

Currently, I think we can jsut always open the compaction service in all horaedb node?

src/server/src/grpc/compaction_service/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+            instance,
+            runtime,
+            compaction_runner,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct CompactionServiceImpl {
+    pub cluster: ClusterRef,
```
Rachelint, line 66: I guess we can compaction serivce may be unnecessary to hold the `cluster`.

src/server/src/grpc/compaction_service/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub instance: InstanceRef,
```
Rachelint, line 67: Seems `instance` is unused currently.

src/analytic_engine/src/compaction/runner/remote_runner.rs
```diff
@@ -0,0 +1,92 @@
...
+        let client = self
+            .get_compaction_client()
+            .await
+            .box_err()
+            .context(GetCompactionClientFailed);
+
+        let pb_resp = match client {
+            Ok(client) => match client.execute_compaction_task(task.clone().into()).await {
+                Ok(resp) => resp,
+                Err(_) => {
```
Rachelint, line 76: I think maybe we need a flag like `fallback_local_when_failed` here for testing.
If it is `true`, we fallback when failed, otherwise we just return the error.

And may be better to add an error log here, it is necessary for us to be notified when remote compaction encounters error.

src/analytic_engine/src/compaction/runner/remote_runner.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub node_picker: RemoteCompactionNodePickerRef,
+    /// Responsible for executing compaction task locally if fail to remote
+    /// compact, used for better fault tolerance.
+    pub local_compaction_runner: LocalCompactionRunner,
+}
+
+impl RemoteCompactionRunner {
+    async fn get_compaction_client(&self) -> Result<CompactionClientRef> {
+        let mut config = CompactionClientConfig::default();
+        let node_addr = self
```
Rachelint, line 44: A `http://` prefix is necessary for the node_addr.
We can refer to the `make_formatted_endpoint` function in remote engine client.

src/server/src/grpc/mod.rs
LeslieKid: Right, the `Cluster` is unnecessary, i should remove it. 

And the current implementation already always open the compaction service in all horaedb nodes.

Rachelint (approved):
LGTM

Rachelint:
Thanks @LeslieKid

PR #1566, opened 2024-09-08
feat: Implement delete operation for WAL based on local storage

dracoooooo:
## Rationale

Currently the WAL based on the local disk does not support the delete function. This PR implements that functionality.

This is a follow-up task of #1552 and #1556.

## Detailed Changes

1. For each `Segment`, add a hashmap to record the minimum and maximum sequence numbers of all tables within that segment. During `delete` and `write` operations, this hashmap will be updated. During read operations, logs will be filtered based on this hashmap.

2. During the `delete` operation, based on the aforementioned hashmap, if all logs of all tables in a read-only segment (a segment that is not currently being written to) are marked as deleted, the segment file will be physically deleted from the disk.

## Test Plan

Unit test, TSBS and running a script locally that repeatedly inserts data, forcibly kills, and restarts the database process to test persistence.

### DIFF
--- a/src/wal/src/local_storage_impl/segment.rs
+++ b/src/wal/src/local_storage_impl/segment.rs
@@ -416,10 +416,6 @@ impl Segment {
         Ok(())
     }
 
-    fn contains_table(&self, table_id: TableId) -> bool {
-        self.table_ranges.contains_key(&table_id)
-    }
-
     fn mark_deleted(&mut self, table_id: TableId, sequence_num: SequenceNumber) {
         if let Some(range) = self.table_ranges.get_mut(&table_id) {
             // If sequence number is MAX, remove the range directly to prevent overflow
@@ -525,17 +521,12 @@ impl SegmentManager {
         location: WalLocation,
         sequence_num: SequenceNumber,
     ) -> Result<()> {
+        let current_segment_id = self.current_segment.lock().unwrap().lock().unwrap().id;
         let mut segments_to_remove: Vec<u64> = Vec::new();
         let mut all_segments = self.all_segments.lock().unwrap();
-        let current_segment_id = self.current_segment.lock().unwrap().lock().unwrap().id;
 
         for (_, segment) in all_segments.iter() {
             let mut segment = segment.lock().unwrap();
-
-            // Skip segments that are not relevant
-            if segment.min_seq > sequence_num || !segment.contains_table(location.table_id) {
-                continue;
-            }
 
             segment.mark_deleted(location.table_id, sequence_num);
 
--- a/src/wal/src/local_storage_impl/segment.rs
+++ b/src/wal/src/local_storage_impl/segment.rs
@@ -19,7 +19,7 @@ use std::{
     cmp::{max, min},
     collections::{HashMap, VecDeque},
     fmt::Debug,
-    fs::{self, remove_file, File, OpenOptions},
+    fs::{self, File, OpenOptions},
     io::{self, Write},
     path::Path,
     sync::{
@@ -116,6 +116,9 @@ pub enum Error {
         source: GenericError,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("{}", source))]
+    Internal { source: anyhow::Error },
 }
 
 define_result!(Error);
@@ -296,6 +299,7 @@ impl Segment {
             }
         }
 
+        self.segment_size = file_size as usize;
         self.mmap = Some(mmap);
         self.record_position = record_position;
         self.current_size = pos;
@@ -418,6 +422,11 @@ impl Segment {
 
     fn mark_deleted(&mut self, table_id: TableId, sequence_num: SequenceNumber) {
         if let Some(range) = self.table_ranges.get_mut(&table_id) {
+            // If sequence number is MAX, remove the range directly to prevent overflow
+            if sequence_num == MAX_SEQUENCE_NUMBER {
+                self.table_ranges.remove(&table_id);
+                return;
+            }
             range.0 = max(range.0, sequence_num + 1);
             if range.0 > range.1 {
                 self.table_ranges.remove(&table_id);
@@ -431,7 +440,9 @@ impl Segment {
 
     fn delete(&mut self) -> Result<()> {
         self.close()?;
-        remove_file(&self.path).context(SegmentAppend)?;
+        fs::remove_file(&self.path)
+            .map_err(anyhow::Error::new)
+            .context(Internal)?;
         Ok(())
     }
 }
@@ -833,8 +844,7 @@ impl Region {
         sequence_num: SequenceNumber,
     ) -> Result<()> {
         self.segment_manager
-            .mark_delete_entries_up_to(location, sequence_num)?;
-        Ok(())
+            .mark_delete_entries_up_to(location, sequence_num)
     }
 
     pub fn close(&self) -> Result<()> {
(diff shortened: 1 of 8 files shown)


src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +515,76 @@ impl SegmentManager {
...
+            if segment.is_empty() && segment.id != current_segment_id {
+                let mut cache = self.cache.lock().unwrap();
+
+                // Check if segment is already in cache
+                if let Some(index) = cache.iter().position(|id| *id == segment.id) {
+                    cache.remove(index);
+                }
+
+                segments_to_remove.push(segment.id);
+                segment.delete()?;
```
jiacai2050, line 543: This is a heavy task, I think we can do this outside all_segment's lock.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +515,76 @@ impl SegmentManager {
...
+        sequence_num: SequenceNumber,
+    ) -> Result<()> {
+        let current_segment_id = self.current_segment.lock().unwrap().lock().unwrap().id;
+        let mut segments_to_remove: Vec<u64> = Vec::new();
+        let mut all_segments = self.all_segments.lock().unwrap();
+
+        for (_, segment) in all_segments.iter() {
+            let mut segment = segment.lock().unwrap();
+
+            segment.mark_deleted(location.table_id, sequence_num);
```
jiacai2050, line 531: Since the seq in once segment is increasing among all tables, so we can just compare the `sequence_num` with its min/max segment of one segment?

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +515,76 @@ impl SegmentManager {
...
         Ok(())
     }
+
+    pub fn mark_delete_entries_up_to(
+        &self,
+        location: WalLocation,
+        sequence_num: SequenceNumber,
+    ) -> Result<()> {
+        let current_segment_id = self.current_segment.lock().unwrap().lock().unwrap().id;
+        let mut segments_to_remove: Vec<u64> = Vec::new();
```
jiacai2050, line 525: Can we avoid this vec with `HashMap::retain`?

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +515,76 @@ impl SegmentManager {
...
+        table_id: Option<TableId>,
+        start: SequenceNumber,
+        end: SequenceNumber,
+    ) -> Result<Vec<u64>> {
+        // Find all segments that contain the requested sequence numbers
+        let mut relevant_segments = Vec::new();
+
+        let all_segments = self.all_segments.lock().unwrap();
+
+        for (_, segment) in all_segments.iter() {
```
jiacai2050, line 565:
```suggestion
        for  segment in all_segments.values() {
```

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -370,9 +401,44 @@ impl Segment {
...
+                seq_range.1 = max(seq_range.1, next_sequence_num - 1);
+            })
+            .or_insert((prev_sequence_num, next_sequence_num - 1));
+
+        Ok(())
+    }
+
+    fn mark_deleted(&mut self, table_id: TableId, sequence_num: SequenceNumber) {
+        if let Some(range) = self.table_ranges.get_mut(&table_id) {
+            // If sequence number is MAX, remove the range directly to prevent overflow
```
jiacai2050, line 421: 👍

src/wal/src/local_storage_impl/segment.rs
dracoooooo: I think no if we move `segment.delete()` outside all_segment's lock,

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -1300,4 +1464,63 @@ mod tests {
...
+        // Mark delete entries up to sequence - 1, so only the last segment should
+        // remain
+        let mark_delete_sequence = sequence - 1;
+        region
+            .mark_delete_entries_up_to(location, mark_delete_sequence)
+            .unwrap();
+
+        {
+            let all_segments = region.segment_manager.all_segments.lock().unwrap();
+            assert_eq!(all_segments.len(), 1);
```
jiacai2050, line 1521: We also need to check file inside wal directory is 1.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -1300,4 +1464,63 @@ mod tests {
...
+
+        for _i in 0..10 {
+            let log_entries = 0..100;
+            let log_batch_encoder = LogBatchEncoder::create(location);
+            let log_batch = log_batch_encoder
+                .encode_batch(log_entries.clone().map(|v| MemoryPayload { val: v }))
+                .expect("should succeed to encode payloads");
+
+            let write_ctx = WriteContext::default();
+            region.write(&write_ctx, &log_batch).unwrap();
```
jiacai2050, line 1498: We need to check the returned seq is expected.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +516,89 @@ impl SegmentManager {
...
+                if let Some(index) = cache.iter().position(|(id, _)| *id == guard.id) {
+                    cache.remove(index);
+                }
+
+                segments_to_remove.push((guard.id, segment.clone()));
+            }
+        }
+
+        // Delete segments in all_segments
+        for (segment_id, _) in segments_to_remove.iter() {
```
jiacai2050, line 548: keys()

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +516,89 @@ impl SegmentManager {
...
+            all_segments.remove(segment_id);
+        }
+
+        drop(all_segments);
+
+        // Delete segments on disk
+        for (_, segment) in segments_to_remove.iter() {
```
jiacai2050, line 555: values()

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -454,6 +516,89 @@ impl SegmentManager {
...
+        let mut all_segments = self.all_segments.lock().unwrap();
+        let mut segments_to_remove = Vec::new();
+
+        for (_, segment) in all_segments.iter() {
+            let mut guard = segment.lock().unwrap();
+
+            guard.mark_deleted(location.table_id, sequence_num);
+
+            // Delete this segment if it is empty
+            if guard.is_empty() && guard.id != current_segment_id {
```
jiacai2050, line 535: Since current segment may changed, the check here should be `guard.id < current_segment_id`.

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -383,10 +465,13 @@ pub struct SegmentManager {
     all_segments: Mutex<HashMap<u64, Arc<Mutex<Segment>>>>,
 
     /// Cache for opened segments
-    cache: Mutex<VecDeque<u64>>,
+    cache: Mutex<VecDeque<(u64, Arc<Mutex<Segment>>)>>,
```
jiacai2050, line 468: Why add seq here?

src/wal/src/local_storage_impl/segment.rs
dracoooooo: `segments_to_remove` is `Vec<(u64, Arc<...>)>`, so there is no `keys()` for this type. Do I need to change it to a hashmap?

src/wal/src/local_storage_impl/segment.rs
dracoooooo: The `u64` in this tuple represents the segment ID. The reason for storing it is to determine if a segment exists in the cache without needing to acquire the lock for that segment. The newly added `Arc<Mutex<Segment>>` is to remove the original `SegmentManager::get_segment` method, which is redundant and may cause deadlock.

jiacai2050 (approved):
LGTM

PR #1568, opened 2024-09-14
fix: support to compat the old layered memtable options

Rachelint:
## Rationale
We introduce the explicit flag to control should we enable layered memtable, but it has some compatibility problem when upgrading from old version.
This pr add an option to support compating the old layered memtable on/off control method.

## Detailed Changes
 Add an option to support compating the old layered memtable on/off control method.

## Test Plan
Manually.

### DIFF
--- a/src/analytic_engine/src/instance/open.rs
+++ b/src/analytic_engine/src/instance/open.rs
@@ -122,6 +122,7 @@ impl Instance {
             preflush_write_buffer_size_ratio: ctx.config.preflush_write_buffer_size_ratio,
             manifest_snapshot_every_n_updates: ctx.config.manifest.snapshot_every_n_updates,
             enable_primary_key_sampling: ctx.config.enable_primary_key_sampling,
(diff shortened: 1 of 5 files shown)


jiacai2050 (approved):
LGTM

jiacai2050:
```
2024-09-14 15:41:23.595 ERRO [src/components/panic_ext/src/lib.rs:57] thread 'main' panicked 'Failed to create catalog manager: BuildSysCatalog { sou
rce: OpenTable { source: OpenTableWithCause { msg: None, source: ReadMetaUpdate { table_id: TableId(1099511627777), source: Error(Other { source: fai
led to new table_data, space_id:1, table_name:sys_catalog                                                                                          
                                                                                            
Caused by:                                                                                                                        
    Found invalid table opts, msg:layered memtable is enabled but mutable_switch_threshold is 0.
    Backtrace:
```
When error above is thrown, try add following config:

```
[analytic]
try_compat_old_layered_memtable_opts = true
```

PR #1569, opened 2024-09-14
chore: record replay cost in log

jiacai2050:
## Detailed Changes
1. Add replay cost in log
2. Remove verbose http log
3. Recover default to shard based, which is faster in most wal implementation.


Rachelint (approved):
LGTM

PR #1570, opened 2024-09-19
fix: logs might be missed during RegionBased replay in the WAL based on local disk

dracoooooo:
## Rationale
In RegionBased replay, a batch of logs is first scanned from the WAL, and then replayed on various tables using multiple threads. This approach works fine for WALs based on tables, as the logs for each table are clustered together. However, in a WAL based on local disk, the logs for each table may be scattered across different positions within the batch. During multi-threaded replay, it is possible that for a given table, log2 is replayed before log1, resulting in missed logs.

## Detailed Changes
1. Modify `split_log_batch_by_table` function to aggregate all logs for a table together.
2. Modify `tableBatch` struct to change a single range into a `Vec<Range>`.

## Test Plan
Manual testing.

### DIFF
--- a/src/analytic_engine/src/instance/wal_replayer.rs
+++ b/src/analytic_engine/src/instance/wal_replayer.rs
@@ -395,6 +395,11 @@ impl RegionBasedReplay {
             if failed_tables.contains_key(&table_batch.table_id) {
                 continue;
             }
+            let log_entries: Vec<_> = table_batch
+                .ranges
+                .iter()
+                .flat_map(|range| log_batch.range(range.clone()))
+                .collect();
 
             let serial_exec_ctxs = serial_exec_ctxs.clone();
             replay_tasks.push(async move {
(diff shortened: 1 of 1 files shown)


src/analytic_engine/src/instance/wal_replayer.rs
```diff
@@ -439,13 +444,14 @@ impl RegionBasedReplay {
...
-        //  |1|1|2|2|2|3|3|3|3|
+        //  |1|1|2|2|2|3|3|3|3|1|1|
         //
         // output batches:
-        //  |1|1|, |2|2|2|, |3|3|3|3|
+        //  |1|1|1|1|, |2|2|2|, |3|3|3|3|
         let mut start_log_idx = 0usize;
         let mut curr_log_idx = 0usize;
         let mut start_table_id = log_batch.get(start_log_idx).unwrap().table_id;
+        let mut table_ranges = BTreeMap::new();
```
jiacai2050, line 454: Why we need to order by table id?

IMO, there is no need for it.

dracoooooo: Yse, fixed.

jiacai2050 (approved):
LGTM

PR #1572, opened 2024-09-24
feat: use multithreading to optimize WAL replayer

dracoooooo:
## Rationale
Currently, the WAL replayer uses coroutines to replay the WAL logs of multiple tables in parallel. However, coroutines utilize at most one CPU. By switching to a multithreaded approach, we can fully leverage multiple CPUs.

## Detailed Changes
1. Add dependency [async-scoped](https://docs.rs/async-scoped/latest/async_scoped/).
2. Modify both `TableBasedReplay` and `RegionBasedReplay` to use the `spawn task` approach for parallelism, with a maximum of 20 tasks running concurrently.

## Test Plan
Manual testing.

### DIFF
--- a/src/analytic_engine/src/instance/wal_replayer.rs
+++ b/src/analytic_engine/src/instance/wal_replayer.rs
@@ -30,14 +30,13 @@ use common_types::{
     schema::{IndexInWriterSchema, Schema},
     table::ShardId,
 };
-use futures::StreamExt;
 use generic_error::BoxError;
 use lazy_static::lazy_static;
 use logger::{debug, error, info, trace, warn};
 use prometheus::{exponential_buckets, register_histogram, Histogram};
 use snafu::ResultExt;
 use table_engine::table::TableId;
-use tokio::sync::{Mutex, MutexGuard};
+use tokio::sync::{Mutex, MutexGuard, Semaphore};
 use wal::{
     log_batch::LogEntry,
     manager::{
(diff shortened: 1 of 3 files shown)


src/wal/src/local_storage_impl/segment.rs
```diff
@@ -1199,25 +1210,87 @@ impl MultiSegmentLogIterator {
         Ok(iter)
     }
 
+    fn preload_next_segment(&mut self) {
```
jiacai2050, line 1213: assert!(next_iterator_future.is_none())

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -1150,6 +1152,9 @@ pub struct MultiSegmentLogIterator {
     /// Current segment iterator.
     current_iterator: Option<SegmentLogIterator>,
 
+    /// Future iterator for preloading the next segment.
+    next_iterator_future: Option<JoinHandle<Result<SegmentLogIterator>>>,
```
jiacai2050, line 1156: next_segment_iterator

src/analytic_engine/src/instance/wal_replayer.rs
```diff
@@ -381,49 +381,53 @@ impl RegionBasedReplay {
...
-                        &ctx.table_data,
-                        log_entries.into_iter(),
-                    )
-                    .await;
-                    (table_batch.table_id, Some(result))
-                } else {
-                    (table_batch.table_id, None)
+        let ((), results) = async_scoped::TokioScope::scope_and_block(|scope| {
+            // Run at most 20 tasks in parallel
+            let semaphore = Arc::new(Semaphore::new(20));
```
jiacai2050, line 393: Define 20 as a const value, so it can be reused.

jiacai2050 (approved):
LGTM

PR #1573, opened 2024-09-25
test: add integration test for compaction offload

### LINKED ISSUE #1571: Integration test for compaction offload
### Describe This Problem

The subproblem for #1545 . 
Currently, the compaction offload lacks integration tests. Impl these tests is crucial for ensuring the reliability and correctness of the compaction offload.
Secondly, adding support for a manual compact operation in SQL would provide more flexibility to users, allowing them to trigger table compaction on demand. 

### Proposal

The procedure to test the compaction module (in sql):
1. Create a table.
2. Insert data into table.
3. Flush in-memory data into sst.
4. Update data to simulate conditions requiring compaction.
5. Flush.
6. Trigger compaction. (manual compaction)
7. ~~Show compaction status.(optional)~~ (delete for easiness)
8. Verify the result.

The problem is that horaedb has not support manual compaction in sqlness yet. I want to impl `compact` as a pre-command, similar to `flush` command. The reason is: 
+ **Consistency with existing command.** `flush` forces in-memory data to be written to SST files. Similarly, `compact` could be triggered to merge and optimize these SST files, aligning the two operations under the same command paradigm.

### Additional Context

_No response_

LeslieKid:
## Rationale
Close #1571 

## Detailed Changes
- Impl `compact` as pre-command for sqlness, with http compaction service.
- update integration tests to test compaction offload.


Rachelint:
For simplificity, I think we can just setup a standalone horaedb in test which acts as both `remote compaction caller` and `remote compaction server`?

integration_tests/cases/common/compact/compact.sql
```diff
@@ -0,0 +1,52 @@
...
+    VALUES (1, 100, "d1"), (2, 200, "d2"), (3, 300, "d3");
+
+-- SQLNESS ARG pre_cmd=flush
+INSERT INTO `compact_table1` (`timestamp`, `value`, `dic`)
+    VALUES (1, 100, "update_d1"), (2, 200, "update_d2"), (3, 300, "update_d3");
+
+
+-- trigger manual compaction after flush memtable
+-- SQLNESS ARG pre_cmd=flush
+-- SQLNESS ARG pre_cmd=compact
```
Rachelint, line 43: I remember an min threshold exists for triggering compaction of table.
Maybe we should generate more ssts before calling to try triggering the compaction.
```
impl Default for SizeTieredCompactionOptions {
    fn default() -> Self {
        Self {
            bucket_low: 0.5,
            bucket_high: 1.5,
            min_sstable_size: ReadableSize::mb(50),
            min_threshold: 4,
            max_threshold: 16,
            max_input_sstable_size: get_max_input_sstable_size(),
        }
    }
}
```

integration_tests/config/horaedb-cluster-0.toml
```diff
@@ -40,6 +40,11 @@ data_dir = "/tmp/horaedb0"
 type = "RocksDB"
 data_dir = "/tmp/horaedb0"
 
+[analytic.compaction_mode]
```
Rachelint, line 43: I guess we should create an new config only for testing remote compaction.

src/server/src/http.rs
```diff
@@ -524,6 +525,54 @@ impl Service {
             })
     }
 
+    // POST /debug/compact_table
+    fn compact_table(
+        &self,
+    ) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+        warp::path!("debug" / "compact_table")
```
Rachelint, line 532: 👍

integration_tests/Makefile
```diff
@@ -80,15 +85,20 @@ run-horaedb-cluster: build-horaedb
 	nohup ${HORAEDB_BINARY_PATH} --config ${HORAEDB_CONFIG_FILE_1} > ${CLUSTER_HORAEDB_STDOUT_FILE_1} 2>&1 &
 	sleep 30
 
-run: prepare build-meta
-	$(HORAEDB_TEST_BINARY)
+run:
+	make run-local
+	make run-cluster
+	make run-compaction-offload
```
LeslieKid, line 91: In the previous implementation, it just run `prepare` to clear the old process in the very beginning, which may lead to `ConnectError` bug. (e.g. `run-local` and `run-cluster` use the same grpc port but the previous process is not stop yet, which makes the later process can not use the port.)

The current implementation avoids the bug describe above by running `prepare` before each (different) processes.

It needs further improvement in the future -- The log now looks a bit strange

Rachelint (approved):
LGTM

PR #1576, opened 2024-10-16
fix: remove redundancy length field in wal record

jiacai2050:
## Rationale
`length` field is not required in wal record, it's duplicated with value_length.

## Detailed Changes
- Remove length from wal record
- Remove rocksdb-wal from default features

## Test Plan
CI and manually do some benchmark with [avalanche](https://github.com/prometheus-community/avalanche)

### DIFF
--- a/src/wal/src/local_storage_impl/record_encoding.rs
+++ b/src/wal/src/local_storage_impl/record_encoding.rs
@@ -17,6 +17,7 @@
 
 use bytes_ext::{Buf, BufMut, SafeBuf, SafeBufMut};
 use codec::Encoder;
+use crc32fast::Hasher;
 use generic_error::GenericError;
 use macros::define_result;
 use snafu::{ensure, Backtrace, ResultExt, Snafu};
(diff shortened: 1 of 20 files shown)


src/wal/src/local_storage_impl/record_encoding.rs
```diff
@@ -172,38 +172,26 @@ impl RecordEncoding {
...
-        );
+        let table_id = buf.try_get_u64().context(Decoding)?;
+        let sequence_num = buf.try_get_u64().context(Decoding)?;
+        let value_length = buf.try_get_u32().context(Decoding)?;
+        let value = buf[0..value_length as usize].to_vec();
 
         // Verify CRC
-        let data = &buf[0..length as usize];
-        let computed_crc = crc32fast::hash(data);
+        let mut h = Hasher::new();
```
chunshao90, line 189: Change the calculation logic? Can you comment out the calculation logic?

jiacai2050: fixed

baojinri (approved):
LGTM

Issue #1567, opened 2024-09-14
Cli tools to decode wal segment


jiacai2050:
### Describe This Problem

I'm excited to share that we've successfully implemented a local Write-Ahead Log (WAL) following the completion of #1566. While we haven't conducted extensive performance benchmarking yet, our initial tests suggest that our implementation is performing comparably to the one based on RocksDB. This marks a significant milestone in our project's development, and we're looking forward to further optimizing and testing this new feature.

### Proposal

In order to trouble shoot WAL issues, we need a cli tool to decode wal segment, to include at least following infos:
- segment version
- min/max seq of each table

zealchen:
Nice. Let me handle this one.

jiacai2050:
Thanks, you can add another project here https://github.com/apache/horaedb/tree/main/src/tools/src/bin

PR #1584, opened 2024-10-30
feat: add a cli tool to read wal meta information

zealchen:
## Rationale
#1567 

## Detailed Changes
A cli tool to decode wal segment.

## Test Plan
Manual test

image

image


src/tools/src/bin/wal-reader.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    if !std::path::Path::new(&args.data_dir).is_dir() {
+        pretty_error_then_exit(
+            format!("Error: Data directory '{}' does not exist", &args.data_dir).as_str(),
+        );
+    }
+
+    let runtime = Arc::new(Builder::default().build().unwrap());
+    let region_manager =
+        RegionManager::new(args.data_dir.clone(), 32, args.segment_size, runtime).unwrap();
```
jiacai2050, line 140: segment_size is only used when create new segment, so I guess we can pass a fixed one here.

src/tools/src/bin/wal-reader.rs
```diff
@@ -0,0 +1,165 @@
...
+    eprintln!("\x1b[31m{}\x1b[0m", err_msg);
+    std::process::exit(1);
+}
+
+fn main() {
+    let args = Args::parse();
+    println!("WAL type: {}", args.r#type);
+    println!("Data directory: {}", args.data_dir);
+
+    if args.r#type.to_lowercase() != "local" {
```
jiacai2050, line 128: I think we can remove `type` arg, we could infer wal type based on file name.

Users of this tool don't care about wal type.

src/tools/src/bin/wal-reader.rs
```diff
@@ -0,0 +1,165 @@
...
+    #[clap(short, long, default_value = "/tmp/horaedb/wal")]
+    data_dir: String,
+
+    /// Region id
+    #[clap(short = 'r', long, default_value = None)]
+    region_id: Option<u64>,
+
+    /// Segment id
+    #[clap(short = 's', long, default_value = None)]
+    segment_id: Option<u64>,
```
jiacai2050, line 47: This arg is not used?

src/wal/src/local_storage_impl/segment.rs
```diff
@@ -615,6 +615,32 @@ impl SegmentManager {
     }
 }
 
+pub struct SegmentTableMeta {
```
jiacai2050, line 618: I think we should define this in `tools`, not here.

zealchen: To minimize exposing the internal objects, I chose not to return all the segments directly. Furthermore, since all_segments contain Mutexs, exposing it through the interface could compromise both safety and maintainability.

jiacai2050: > To minimize exposing the internal objects, I chose not to return all the segments directly.

Sounds reasonable for me. I suggest we rename this struct to `SegmentView` to represent it's a static version of segment.

zealchen: Done.

jiacai2050 (approved):
LGTM

PR #1587, opened 2024-11-05
chore: fix LICENSE and NOTICE

jiacai2050:
## Rationale
See https://lists.apache.org/thread/29nyod5d73tn5nov7f5pm0tg42gq57vw

Adhere to https://www.apache.org/legal/src-headers.html

## Test Plan
No need.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -200,3 +200,43 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+================================================================
+
+This product includes a number of Dependencies with separate copyright notices
+and license terms. Your use of these submodules is subject to the terms and
+conditions of the following licenses.
+
+================================================================
+
+================================================================
+Apache-2.0 licenses
+================================================================
+The following components are provided under the Apache-2.0 License.See project link for details.
+The text of each license is the standard Apache 2.0 license.
+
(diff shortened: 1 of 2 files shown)


tisonkun (approved):

LICENSE
```diff
@@ -199,4 +199,44 @@
...
+File src/components/future_ext/src/cancel.rs is modified from influxdb.
+Files src/components/tracing_util/src/lib.rs, src/components/tracing_util/src/logging.rs, src/components/tracing_util/Cargo.toml are modified from databend.
+
+================================================================
+MIT licenses
+================================================================
+The following components are provided under the MIT License. See project link for details.
+The text of each license is also included in licenses/LICENSE-[project].txt
+
+* consistent(https://github.com/buraksezer/consistent)
```
pjfanning, line 239: You should include the copyright header from https://github.com/buraksezer/consistent/blob/master/LICENSE

jiacai2050: It's ready here: https://github.com/apache/horaedb/blob/main/licenses/LICENSE-consistent.txt

Should I copy that to this file?

pjfanning: that's a reasonable point - it's fine as is, then

@jiacai2050 can you do a new RC? The LICENSE files matter and it would be good for the new release to contain this fix.

jiacai2050: Yes, I will start a new VOTE threads after cherry pick this patch.

PR #1590, opened 2024-11-12
fix: skip build latest docker tag for RC version.

### LINKED ISSUE #1589: Skip build latest docker tag for RC version
ASF require this

> The latest tag should not point to an artifact containing unapproved code e.g. to master or dev branches or to a RC or snapshot.

https://incubator.apache.org/guides/distribution.html#docker

Related file:
https://github.com/apache/horaedb/blob/0d8d43d4f434e098353ee0940382a8f00e144e8b/.github/workflows/publish-image.yml#L45

misselvexu:
## Rationale
Close #1589 

## Detailed Changes
Upgrade `publish-image.yml` for skip build latest docker tag for RC version.

## Test Plan
pass

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -44,7 +44,9 @@ jobs:
         with:
           context: horaemeta
           push: true
-          tags: apache/horaemeta-server:${{ inputs.version }}
+          tags: |
+            apache/horaemeta-server:${{ inputs.version }}
+            ${{ contains(inputs.version.toLowerCase(), 'rc') == false && 'apache/horaemeta-server:latest' || '' }}
 
   horaedb:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-image.yml
```diff
@@ -42,13 +44,15 @@ jobs:
         with:
           context: horaemeta
           push: true
-          tags: apache/horaemeta-server:latest,apache/horaemeta-server:${{ inputs.version }}
```
jiacai2050, line 45: This will skip `latest` in all tags, what I want is to skip latest only when tags contains `rc`, for tags like `v.2.1.0`, we should include latest tag.

jiacai2050 (approved):
LGTM

PR #1593, opened 2024-11-12
fix:  avoid splicing bucket multiple times in oss url

baojinri:
## Rationale
 Avoid splicing bucket multiple times in oss url

## Test Plan
CI


jiacai2050 (approved):
LGTM

PR #1596, opened 2024-11-15
feat: support persist manifest efficientlly

### LINKED ISSUE #1592: Support more efficient way to persist manifest of TimeMergeStorage
### Describe This Problem

Now, when meta information of TimeMergeStorage changes, the efficiency of persisting manifest sst files is very low. Each time, the new sst file will be added to the manifest and then persisted it whole.

### Proposal

Each time, the new sst file will be persisted instantly, then add the new sst into snapshot background.
The manifest file is organized as follows:
```
/root/manifest/snapshot
/root/manifest/new_sst_0
/root/manifest/new_sst_1
/root/manifest/new_sst_2
...
```

### Additional Context
Process manifest files：
- When initializing the manifest, we will list all the sst files in the manifest directory and then add them into the snapshot
- Every time a new file is added, we will count the new files. When the number of new files reaches 50, we will list all the sst files in the manifest directory and then add them into the snapshot

baojinri:
## Rationale
close #1592 

## Detailed Changes
- support merge manifest ssts background

## Test Plan
CI

### DIFF
--- a/horaedb/metric_engine/src/manifest.rs
+++ b/horaedb/metric_engine/src/manifest.rs
@@ -108,8 +108,6 @@ impl Manifest {
             merge_options,
         )
         .await?;
-        // Merge all delta files when startup
-        merger.do_merge().await?;
 
         {
             let merger = merger.clone();
@@ -142,7 +140,7 @@ impl Manifest {
             .context("failed to encode manifest file")?;
         let put_payload = PutPayload::from_bytes(Bytes::from(buf));
 
-        // 1. Persist the manifest file
+        // 1. Persist the delta manifest
         self.store
             .put(&new_sst_path, put_payload)
             .await
@@ -153,6 +151,9 @@ impl Manifest {
             let mut payload = self.payload.write().await;
             payload.files.push(new_sst);
         }
+
+        // 3. Update delta files num
+        self.merger.add_delta_num();
 
         Ok(())
     }
@@ -213,7 +214,7 @@ impl ManifestMerger {
         merge_options: ManifestMergeOptions,
     ) -> Result<Arc<Self>> {
         let (tx, rx) = mpsc::channel(merge_options.channel_size);
-        let manifest_merge = Self {
+        let merger = Self {
             snapshot_path,
             delta_dir,
             store,
@@ -222,8 +223,10 @@ impl ManifestMerger {
             deltas_num: AtomicUsize::new(0),
             merge_options,
         };
+        // Merge all delta files when startup
+        merger.do_merge().await?;
 
-        Ok(Arc::new(manifest_merge))
+        Ok(Arc::new(merger))
     }
 
     async fn run(&self) {
@@ -237,11 +240,12 @@ impl ManifestMerger {
                             error!("Failed to merge delta, err:{err}");
                         }
                     }
-
                 }
                 _merge_type = receiver.recv() => {
-                    if let Err(err) = self.do_merge().await {
-                        error!("Failed to merge delta, err:{err}");
+                    if self.deltas_num.load(Ordering::Relaxed) > self.merge_options.min_merge_threshold {
+                        if let Err(err) = self.do_merge().await {
+                            error!("Failed to merge delta, err:{err}");
+                        }
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


horaedb/metric_engine/src/manifest.rs
```diff
@@ -29,13 +39,16 @@ use crate::{
 
 pub const PREFIX_PATH: &str = "manifest";
 pub const SNAPSHOT_FILENAME: &str = "snapshot";
+pub const SST_PREFIX: &str = "sst";
```
jiacai2050, line 42: Those files should not be named sst, sst is only used in LSM, and represent sorted string table, here those file are delta files based on snapshot.

horaedb/metric_engine/src/manifest.rs
```diff
@@ -140,4 +173,189 @@ impl Manifest {
...
+        let mut paths = Vec::with_capacity(meta_infos.len());
+        for meta_info in meta_infos {
+            let path = meta_info
+                .context("failed to get path of manifest sst")?
+                .location;
+            paths.push(path);
+        }
+
+        let mut sst_files = Vec::with_capacity(paths.len());
+        for path in &paths {
```
jiacai2050, line 294: We could fetch those files concurrently, one after one is too slow.

horaedb/metric_engine/src/manifest.rs
```diff
@@ -140,4 +173,189 @@ impl Manifest {
...
+            let bytes = self
+                .store
+                .get(path)
+                .await
+                .context("failed to read sst file")?
+                .bytes()
+                .await
+                .context("failed to read sst file")?;
+            let pb_sst = pb_types::SstFile::decode(bytes).context("failed to decode sst file")?;
+            sst_files.push(SstFile::try_from(pb_sst)?);
```
jiacai2050, line 304: We need to dedup files before merge, since there may exists old delta files, which are failed to delete in last merge.

horaedb/metric_engine/src/manifest.rs
```diff
@@ -140,4 +173,189 @@ impl Manifest {
...
+        let put_payload = PutPayload::from_bytes(Bytes::from(buf));
+
+        // 1. Persist the snapshot
+        self.store
+            .put(&self.snapshot_path, put_payload)
+            .await
+            .context("Failed to update manifest")?;
+
+        // 2. Delete the old sst files
+        for path in &paths {
```
jiacai2050, line 352: Delete concurrently

horaedb/metric_engine/src/manifest.rs
```diff
@@ -140,4 +173,189 @@ impl Manifest {
...
+            self.store
+                .delete(path)
+                .await
+                .context("failed to delete sst file")?;
```
jiacai2050, line 356: add path name to log

horaedb/metric_engine/src/manifest.rs
```diff
@@ -69,8 +82,29 @@ impl From<Payload> for pb_types::Manifest {
...
+            rx,
+            merge_options,
+        )
+        .await?;
+
+        // Start merge ssts task
+        tokio::spawn(async move {
+            ssts_merge.run().await;
+        });
+
```
jiacai2050, line 107: When start up, we also need to read delta files, if we only read snapshot, then part of sst won't be queried later?

horaedb/metric_engine/src/manifest.rs
```diff
@@ -140,4 +173,189 @@ impl Manifest {
...
+    max_interval_seconds: usize,
+    merge_threshold: usize,
+}
+
+impl Default for SstsMergeOptions {
+    fn default() -> Self {
+        Self {
+            channel_size: 100,
+            max_interval_seconds: 5,
+            merge_threshold: 50,
```
jiacai2050, line 200: 10-20 is a reasonable choice.

horaedb/metric_engine/src/manifest.rs
```diff
@@ -90,43 +124,42 @@ impl Manifest {
...
     }
 
-    // TODO: Now this functions is poorly implemented, we concat new_sst to
-    // snapshot, and upload it back in a whole.
-    // In more efficient way, we can create a new diff file, and do compaction in
-    // background to merge them to `snapshot`.
     pub async fn add_file(&self, id: FileId, meta: FileMeta) -> Result<()> {
-        let mut payload = self.payload.write().await;
-        let mut tmp_ssts = payload.files.clone();
+        let new_sst_path = Path::from(format!("{}/{id}", self.sst_path));
```
jiacai2050, line 140: If there are too many pending delta files, we should block the write here.

jiacai2050 (approved):
LGTM

PR #1603, opened 2024-12-06
feat: support pick ssts in compact

baojinri:
## Rationale
Implement pick_candidate for compaction

## Test Plan
UT

### DIFF
--- a/horaedb/metric_engine/src/compaction/picker.rs
+++ b/horaedb/metric_engine/src/compaction/picker.rs
@@ -22,19 +22,20 @@ use std::{
 
 use tracing::debug;
 
-use crate::{
-    compaction::Task,
-    sst::{FileId, SstFile},
-    types::Timestamp,
-};
+use super::SchedulerConfig;
+use crate::{compaction::Task, sst::SstFile, types::Timestamp};
 
 pub struct TimeWindowCompactionStrategy {
     segment_duration: Duration,
+    config: SchedulerConfig,
 }
 
 impl TimeWindowCompactionStrategy {
-    pub fn new(segment_duration: Duration) -> Self {
-        Self { segment_duration }
+    pub fn new(segment_duration: Duration, config: SchedulerConfig) -> Self {
+        Self {
+            segment_duration,
+            config,
+        }
     }
 
     pub fn pick_candidate(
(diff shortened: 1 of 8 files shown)


horaedb/metric_engine/src/compaction/picker.rs
```diff
@@ -28,7 +37,105 @@ impl TimeWindowCompactionStrategy {
...
+    }
+
+    fn get_compact_files(&self, buckets: HashMap<i64, Vec<FileId>>) -> Vec<FileId> {
+        let all_keys: BTreeSet<_> = buckets.keys().collect();
+        let mut compact_files = Vec::new();
+
+        for key in all_keys.into_iter().rev() {
+            if let Some(bucket) = buckets.get(key) {
+                if bucket.len() >= 2 {
+                    compact_files = bucket.clone();
```
jiacai2050, line 109: We need to limit input compact files, also we should compact smaller files first.

jiacai2050: We a SST is picked, we should mark it in compacting, so it won't got compacted twice.

horaedb/metric_engine/src/sst.rs
```diff
@@ -71,6 +74,18 @@ impl SstFile {
     pub fn mark_compaction(&self) {
         self.inner.in_compaction.store(true, Ordering::Relaxed);
     }
+
+    pub fn is_uncompacted(&self) -> bool {
+        !self.inner.in_compaction.load(Ordering::Relaxed)
+    }
```
jiacai2050, line 80:
```suggestion
    pub fn is_compaction(&self) -> bool {
        self.inner.in_compaction.load(Ordering::Relaxed)
    }
```

Negate this bool inside this function is not very readable.

horaedb/metric_engine/src/compaction/picker.rs
```diff
@@ -28,7 +37,105 @@ impl TimeWindowCompactionStrategy {
...
+
+    ///  Group files of similar timestamp into buckets.
+    fn get_buckets(&self, files: &[SstFile]) -> HashMap<i64, Vec<FileId>> {
+        let mut buckets: HashMap<i64, Vec<FileId>> = HashMap::new();
+        for f in files {
+            let (left, _) = self.get_window_bounds(f.meta().time_range.end);
+
+            let bucket_files = buckets.entry(left).or_default();
+
+            bucket_files.push(f.id());
```
jiacai2050, line 81: FileId is not suitable here, we need to mutate sst inner state, so a Task should hold `SstFile`

horaedb/metric_engine/src/compaction/picker.rs
```diff
@@ -28,7 +37,105 @@ impl TimeWindowCompactionStrategy {
...
+        }
+
+        debug!(
+            "Group files of similar timestamp into buckets: {:?}",
+            buckets
+        );
+        buckets
+    }
+
+    fn get_window_bounds(&self, ts: Timestamp) -> (i64, i64) {
```
jiacai2050, line 91: Add UT for this function.

jiacai2050 (changes requested):

PR #1616, opened 2024-12-20
fix: manifest delta num may overflow

jiacai2050:
## Rationale
Fix bugs found in local write bench.

## Detailed Changes
- When manifest starts up, the delta num may overflow since it's initialized to 0.
- When manifest merge_update, since the deltas is unsorted, so we may first delete a non-existing file.

## Test Plan
CI

### DIFF
--- a/src/metric_engine/src/manifest/encoding.rs
+++ b/src/metric_engine/src/manifest/encoding.rs
@@ -99,12 +99,12 @@ impl SnapshotHeader {
     pub const LENGTH: usize = 4 /*magic*/ + 1 /*version*/ + 1 /*flag*/ + 8 /*length*/;
     pub const MAGIC: u32 = 0xCAFE_1234;
 
-    pub fn new(length: u64) -> Self {
+    pub fn new() -> Self {
         Self {
             magic: SnapshotHeader::MAGIC,
             version: SnapshotRecord::VERSION,
             flag: 0,
-            length,
+            length: 0,
         }
     }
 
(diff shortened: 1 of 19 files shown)


src/metric_engine/src/manifest/encoding.rs
```diff
@@ -303,18 +303,30 @@ impl Snapshot {
...
+
+        self.records
+            .retain(|record| !to_deletes.contains(&record.id));
         self.header.length = (self.records.len() * SnapshotRecord::LENGTH) as u64;
-        Ok(())
     }
 
     pub fn into_bytes(self) -> Result<Bytes> {
-        let buf = Vec::with_capacity(self.header.length as usize + SnapshotHeader::LENGTH);
+        let buf = Vec::with_capacity(self.header.length as usize * SnapshotHeader::LENGTH);
```
zealchen, line 329: self.header.length as usize + SnapshotHeader::LENGTH is right ?

self.header.length is the total length of the payload of Snapshot in bytes.

zealchen:
"When manifest merge_update, since the deltas is unsorted, so we may first delete a non-existing file."

Why is the failure to delete related to the sorting of the deltas? Intuitively they are mutually irrelevent with each other.

src/server/src/main.rs
```diff
@@ -52,6 +62,16 @@ async fn hello() -> impl Responder {
     HttpResponse::Ok().body("Hello world!")
 }
 
+#[get("/toggle")]
+async fn toggle(data: web::Data<AppState>) -> impl Responder {
```
zealchen, line 66: Any reasons for choosing Actix over Axum?

src/metric_engine/src/manifest/encoding.rs
jiacai2050: Good catch!

jiacai2050:
> Why is the failure to delete related to the sorting of the deltas? Intuitively they are mutually irrelevent with each other.

This is a tricky bug. For example, we have tow delta files to merge:
1. to_add: 100
2. to_add:  101, to_delete: 100

If the delta files are unsorted, if we first merge `1`,  100 is not in snapshot yet, so it will delete nothing.

src/server/src/main.rs
jiacai2050: No specific reason. GPT says actix is more feature-rich.

baojinri (approved):
LGTM

PR #1618, opened 2024-12-24
feat: skip add sequence column when compact 

jiacai2050:
## Detailed Changes
- Refactor package metadata
- ParquetReader add `keep_sequence` args, and set it to `true`for compaction, `false` for query.

## Test Plan

CI
```
 cargo publish --dry-run --registry crates-io
```
This commands run successfully.


baojinri (approved):
LGTM

Issue #1623, opened 2025-01-15
Implement MetricManager


jiacai2050:
The main task of MetricManager is to manage metric name of a sample, it will:
1. Calculate the id for metric name, using seahash
2. Persist new name to storage. For existing metric name, we should avoid this step, since it will involve IO operation.
3. In order to support 2, we need to add a cache, and when server start up, we load all existing metric name into cache.

PR #1624, opened 2025-01-25
Feat metrics engine cache

zealchen:
## Rationale

#1623 

## Detailed Changes

1. Sequence diagram
image

2. Implement three caches: MetricsCache, SeriesCache, and TagIndexCache, each with an asynchronous serialization function.
3. Implement async write to storage in batch mode.

TODO: 
1. Some parameters need to be made configurable.

## Test Plan

UT
```sh
running 1 test
test index::cache::tests::test_cache_manager_updates ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.05s

     Running unittests src/lib.rs (target/debug/deps/pb_types-2fb383c00286addd)
```

### DIFF
--- a/src/metric_engine/src/index/cache.rs
+++ b/src/metric_engine/src/index/cache.rs
@@ -1,3 +1,4 @@
+use core::num;
 // Licensed to the Apache Software Foundation (ASF) under one
 // or more contributor license agreements.  See the NOTICE file
 // distributed with this work for additional information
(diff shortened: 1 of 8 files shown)


src/metric_engine/src/index/cache.rs
```diff
@@ -0,0 +1,1293 @@
...
+}
+
+#[derive(PartialEq, Eq, Hash, Debug)]
+struct SegmentSeries {
+    segment: SegmentDuration,
+    series_id: SeriesId,
+}
+
+struct TagIndexCache {
+    cache: DashMap<SegmentDuration, ConcurrentTagKVMap>,
```
jiacai2050, line 94: Two thoughts here:
1. We should prefer std map over dash map, those third party may introduce unexpected bugs.
2. For current segment cache, we can use an independent field to represent, so we can save one hashmap lookup.

Other questions we need to consider: 
1. How will you evict segment?

src/metric_engine/src/index/cache.rs
```diff
@@ -0,0 +1,1293 @@
...
+            metric_name,
+            field_name,
+            field_type,
+            filed_id,
+            metric_id,
+            duration,
+        ))
+    }
+
+    async fn load_from_storage(&mut self) -> Result<()> {
```
jiacai2050, line 172: The cache module shouldn't known how the persistence layer is implemented, we can move this method to `IndexManager`.

src/metric_engine/src/index/cache.rs
```diff
@@ -0,0 +1,1293 @@
...
+                    field_name,
+                    field_type,
+                )
+                .await?;
+            }
+        }
+        Ok(())
+    }
+
+    fn schema() -> Arc<Schema> {
```
jiacai2050, line 198: Cache module has no schema, this field belong to IndexManager.

jiacai2050 (changes requested):

PR #1627, opened 2025-05-14
feat: add pprof api

BaldDemian:
## Rationale
Close https://github.com/apache/horaedb/issues/1577

## Detailed Changes
- In `src/server/src/http.rs`, register the pprof api, similar to the existing `/debug/profile/heap/{seconds}`.
- In `src/components/profile/src/lib.rs`, implement `dump_heap_pprof` function, similar to the existing `dump_heap_prof`.

## Test Plan
Manual test. More details are given below.


BaldDemian:
how to test

1.   In shell 1, build and run the HoraeDB server with heap profiling enabled:

     ```shell
     cargo build
     export MALLOC_CONF="prof:true,prof_active:false,lg_prof_sample:19"
     ./target/debug/horaedb-server --config docs/minimal.toml
     ```

2.   In shell 2, send the pprof request by running the following command:

     ```shell
     /bin/python3 ./integration_tests/pprof/pprof.py
     ```

     This script sends a GET `/debug/pprof/heap/{seconds}` HTTP request to the server, enabling heap profiling for the specified duration (default is **20 seconds**). After the profiling completes, a file named `heap.pb.gz` will be generated.

3.   In shell 3, **immediately** after step 2, run the following command to stress the system:

     ```shell
     /bin/python3 ./integration_tests/pprof/stress.py
     ```

     This script creates several random tables and executes memory-intensive operations such as aggregations and joins. On my machine, this stress test completes in about 16 seconds.

4.   Install `pprof` and `Graphviz`. You can generate report files(txt, svg...) using the `pprof` command and the `heap.pb.gz` file.  Generally, running `pprof` commands in machines other than the original HoraeDB machine may see **symbolization** faiure like this:

     
     


     It says there are two missing directories:

     -   `./target/debug/horaedb-server`, the executable binary Rust code
     -   `/lib/x86_64-linux-gnu/libc.so`, the Linux specific libc code

     The reason is that the original `heap.pb.gz` file only contains the addresses of the called functios instead of their names.  `pprof` relies on the above two missing directories to map the function addresses to their names. 

     It is worth mentioning that  the `symbolize` feature is enabled in the `jemalloc_pprof` dependency, so the output `heap.pb.gz` file already contains all Rust functions names. 

     In theory, maybe we can run `pprof` in other machines of the **exact the same** (otherwise the libc code may be different ) OS version of the HoraeDB machine, without seeing above symbolization faiure. 

     But on other OSes, the failure still exists since they don't contain the correct `/lib/x86_64-linux-gnu/libc.so` file.  For example, on my arm64 macOS machine, it shows:

     
     


     But you can still see that the Rust funtion names are shown, due to the enabled `symbolize` feature in the `jemalloc_pprof` dependency. This feature may slightly hurt the performance, but also brings some convenience when using the `pprof` command, so I enable it in this commit.

     Above discussion is similar to this [[comment](https://github.com/apache/horaedb/issues/1577#issuecomment-2464937796)]( https://github.com/apache/horaedb/issues/1577#issuecomment-2464937796).

     An alternative is to add an API that returns the flamegraph directly, like this example:
     https://github.com/polarsignals/rust-jemalloc-pprof/blob/196f26d3ac6d85a8519abc42d2dc0061b64a6069/example/src/main.rs#L42-L57

BaldDemian:
I doubt the necessity of adding these two scripts in integration tests🤔. A simple `curl` command may be enough to test the api.

BaldDemian:
Saw the failing CI checks. Could be related to the changed `cargo.toml` or `cargo.lock`. I will try fix them.

